PR #195, opened 2018-06-12
Update opcodes

Kais-DkM:
Updated script execution related code to match with updated spec.


sgkim126:
@Kais-DkM Would you check rustfmt?

Kais-DkM:
@sgkim126 Applied rustfmt.

PR #208, opened 2018-06-15
Apply conditional build flag(in C compiler)

gurrpi:
Fix: https://github.com/CodeChain-io/codechain/issues/142

Cause: Since ```-Wnonnull-compare``` build option is only available for gcc(https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#Warning-Options),
clang(used by Mac) could not resolve the flag. (https://clang.llvm.org/docs/DiagnosticsReference.html)


kseo:
@gurrpi Would you split the PR into two PRs?

1. Use clogs
1. Conditional build flag

gurrpi:
@kseo OK, I submitted new PR.

PR #209, opened 2018-06-16
Apply conditional build flag(in C compiler)

gurrpi:
Fix: #142
Cause: Since ```-Wnonnull-compare``` build option is only available for gcc(https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#Warning-Options),
clang(used by Mac) could not resolve the flag. (https://clang.llvm.org/docs/DiagnosticsReference.html)


sgkim126:
Thank you @gurrpi

PR #212, opened 2018-06-18
Create snapshot periodically

Kais-DkM:
Create snapshot periodically to predefined folder.
Root directory for snapshots can be specified with command line option `--snapshot-path`.

Under root directory, each snapshot is saved in separated folder.
Name for snapshot's directory is same as key of state root.
Contents of snapshot directory is as follows:
* Head, which is named as `head`.
* Chunks, which is named by its root key


codechain/main.rs
```diff
@@ -225,6 +225,11 @@ fn run_node(matches: ArgMatches) -> Result<(), String> {
         }
     };
 
+    // FIXME: Get snapshot root directory from config
+    // FIXME: Get snapshot period from genesis block
+    let snapshot_service = SnapshotService::new(client.client(), "snapshot".to_string(), 1 << 14);
```
sgkim126: What does `1 << 14` mean?

codechain/main.rs
```diff
@@ -225,6 +225,11 @@ fn run_node(matches: ArgMatches) -> Result<(), String> {
...
+    client.client().add_notify(snapshot_service.clone());
```
sgkim126: I cannot find the reason to clone.

sync/src/snapshot/service.rs
```diff
@@ -55,16 +60,88 @@ impl ChainNotify for Service {
...
-                    // unimplemented!()
+                let number = (best_number / period - 1) * period;
+                let header = client.block_header(BlockId::Number(number)).expect("Snapshot target must exist");
+                let path: PathBuf = [root_dir, format!("{:x}", header.hash())].iter().collect();
+                match create_dir_all(&path) {
+                    Ok(_) => {}
+                    // FIXME: Handle file system errors
+                    Err(_) => return,
+                }
+                let db = client.database();
```
sgkim126: As we had talked F2F, creating the snapshot right after the block generation can be dangerous because of reorganization.

sync/src/snapshot/service.rs
```diff
@@ -55,16 +60,88 @@ impl ChainNotify for Service {
...
+
+                // FIXME: Handle non-existant nodes
+                let root = if let Some(node) = get_node(&db, &header.state_root()) {
```
sgkim126:
```rust
let state_root = header.state_root();
let node = if let Some(node) = get_node(&db, state_root) {
    node
} else {
```

sync/src/snapshot/service.rs
```diff
@@ -55,16 +60,88 @@ impl ChainNotify for Service {
...
+                    (header.state_root(), node)
+                } else {
+                    return
+                };
+                let children = if let Some(nodes) = children_of(&db, &root.1) {
```
sgkim126:
```rust
let children = if let Some(nodes) = children_of(&db, &node) {
```

sync/src/snapshot/service.rs
```diff
@@ -55,16 +60,88 @@ impl ChainNotify for Service {
...
+                        grandchildren.extend(nodes);
+                    } else {
+                        return
+                    }
+                }
+                match File::create(path.join("head")) {
+                    Ok(mut file) => {
+                        let mut stream = RlpStream::new();
+                        stream.begin_unbounded_list();
+                        for (key, value) in vec![root].iter().chain(&grandchildren).chain(&children) {
```
sgkim126:
```rust
for (key, value) in vec![(state_root, node)].iter().chain(&grandchildren).chain(&children) {
```

sync/src/snapshot/service.rs
```diff
@@ -55,16 +60,88 @@ impl ChainNotify for Service {
...
+                };
+                let mut grandchildren = Vec::new();
+                for (_, value) in &children {
+                    if let Some(nodes) = children_of(&db, &value) {
+                        grandchildren.extend(nodes);
+                    } else {
+                        return
+                    }
+                }
+                match File::create(path.join("head")) {
```
sgkim126: Is there the recovery if the server crashed while updating the file?

sync/src/snapshot/service.rs
```diff
@@ -55,16 +60,88 @@ impl ChainNotify for Service {
...
+                            stream.append(key);
+                            stream.append(value);
+                        }
+                        stream.complete_unbounded_list();
+                        match file.write(&stream.drain()) {
+                            Ok(_) => {}
+                            Err(_) => return,
+                        };
+                    }
+                    Err(_) => return,
```
sgkim126: I saw many early returns in this function.
How about making it return ```Result```

codechain/main.rs
Kais-DkM: It's hardcoded constant, and it needs to be read from genesis.
That work will be done in next PR.

sync/src/snapshot/service.rs
Kais-DkM: Current implementation doesn't take snapshot on best block.
Snapshot target must be something like `floor(best_number - snapshot_period, snapshot_period)`.

sync/src/snapshot/service.rs
Kais-DkM: Why do you think that's better?

Kais-DkM: Oh, I see what's the difference. I'll fix this.

sync/src/snapshot/service.rs
Kais-DkM: I'm not sure if `spawn` allows returning `Result` type. I'll check it again.

sync/src/snapshot/service.rs
Kais-DkM: I'm planning on periodically checking and removing malformed snapshot files.
Do you have any better idea on this?

sync/src/snapshot/service.rs
sgkim126: https://doc.rust-lang.org/std/thread/fn.spawn.html `spawn` can return the `Result` if it the `Result` is `Send`.

codechain/main.rs
kseo: Clarify the intention by introducing a named constant.

sync/src/snapshot/service.rs
```diff
@@ -0,0 +1,70 @@
...
+use std::sync::Arc;
+use std::thread::spawn;
+
+use ccore::{BlockChainClient, BlockId, ChainNotify};
+use ctypes::H256;
+
+pub struct Service {
+    client: Arc<BlockChainClient>,
+    root_dir: String,
+    period: u64,
```
kseo: What's the unit of `period`? Please add a comment.

sync/src/snapshot/service.rs
```diff
@@ -0,0 +1,70 @@
...
+            .iter()
+            .map(|hash| self.client.block_number(BlockId::Hash(*hash)).expect("Enacted block must exist"))
+            .any(|number| number % self.period == 0);
+        if is_checkpoint && best_number > self.period {
+            let root_dir = self.root_dir.clone();
+            let period = self.period;
+            spawn(move || {
+                let target = (best_number / period - 1) * period;
+                let path: PathBuf = [root_dir, target.to_string()].iter().collect();
+                if let Ok(_) = create_dir_all(path) {
```
kseo: What if this fails?

sync/src/snapshot/service.rs
```diff
@@ -55,16 +60,83 @@ impl ChainNotify for Service {
...
             let root_dir = self.root_dir.clone();
             let period = self.period;
             spawn(move || {
-                let target = (best_number / period - 1) * period;
-                let path: PathBuf = [root_dir, target.to_string()].iter().collect();
-                if let Ok(_) = create_dir_all(path) {
-                    // FIXME: implement this
-                    // unimplemented!()
+                let number = (best_number / period - 1) * period;
+                let header = client.block_header(BlockId::Number(number)).expect("Snapshot target must exist");
```
kseo: Can you come up with a better name? What about `check_point_number`?

PR #230, opened 2018-06-21
Add IPC support to JSON RPC

### LINKED ISSUE #226: Add IPC support to JSON RPC

redongjun:
Fixes #226 

Request Example
```
echo -n '{"jsonrpc": "2.0", "method": "chain_getBlockNumber", "params": [], "id": null}' | nc -U /tmp/jsonrpc.ipc
```

Response Example
```
{"jsonrpc":"2.0","result":0,"id":null}
```

### DIFF
--- a/codechain/config/mod.rs
+++ b/codechain/config/mod.rs
@@ -22,7 +22,7 @@ use cdiscovery::{KademliaConfig, UnstructuredConfig};
 use clap;
 use cnetwork::{NetworkConfig, SocketAddr};
 use ctypes::{Address, Secret};
-use rpc::HttpConfiguration as RpcHttpConfig;
+use rpc::{HttpConfiguration as RpcHttpConfig, IpcConfiguration as RpcIpcConfig};
 use toml;
 
 #[derive(Debug, PartialEq, Deserialize)]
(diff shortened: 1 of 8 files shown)


codechain/rpc.rs
```diff
@@ -64,6 +66,37 @@ pub fn setup_http_rpc_server(
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub struct IpcConfiguration {
+    pub socket_addr: String,
+}
+
+impl Default for IpcConfiguration {
```
joojis: Please remove it unless it's necessary. The hard-coded path for default value seems not great.

codechain/config/mod.rs
```diff
@@ -221,3 +221,15 @@ pub fn parse_rpc_config(matches: &clap::ArgMatches) -> Result<Option<RpcHttpConf
 
     Ok(Some(config))
 }
+
+pub fn parse_rpc_ipcconfig(matches: &clap::ArgMatches) -> Result<Option<RpcIpcConfig>, String> {
```
joojis: I think "parse_rpc_ipc_config" is better for consistency.

codechain/rpc.rs
```diff
@@ -64,6 +66,29 @@ pub fn setup_http_rpc_server(
...
 }
 
+#[derive(Debug, PartialEq)]
+pub struct IpcConfiguration {
+    pub socket_addr: String,
+}
+
+pub fn new_ipc(cfg: IpcConfiguration, deps: Arc<rpc_apis::ApiDependencies>) -> Result<IpcServer, String> {
+    let path = PathBuf::from(&cfg.socket_addr);
+    let server = setup_ipc_rpc_server(path.to_str().unwrap(), deps)?;
```
joojis: How about passing cfg.socket_addr.as_ref() instead of the PathBuf?

I'm not sure PathBuf is necessary.

redongjun: You're right, it's not necessary. 

Lock files were automatically created and not intentionally included. I didn't know it was necessary.

I will apply your requests, and commit it again. 
Please review it again.

PR #238, opened 2018-06-21
Burn Assets

Kais-DkM:


kseo:
@Kais-DkM Please add burn tests.

core/src/parcel.rs
```diff
@@ -404,12 +404,16 @@ pub struct AssetOutPoint {
...
+pub const UNLOCK_SUCCESS: u8 = 0x01;
+pub const UNLOCK_BURN: u8 = 0x02;
+
 #[derive(Debug, Clone, Eq, PartialEq, RlpDecodable, RlpEncodable, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct AssetTransferInput {
     pub prev_out: AssetOutPoint,
     pub lock_script: Bytes,
     pub unlock_script: Bytes,
+    pub unlock_type: u8,
```
sgkim126: I don't agree with that making this flag.
The most use cases will be just the transfer not burning, and this flag seems verbose.
How about splitting transaction input list?

Kais-DkM: Agreed. I'll fix this.

core/src/state/mod.rs
```diff
@@ -633,6 +637,7 @@ impl<B: Backend> State<B> {
                 ..
             } => Ok(self.mint_asset(transaction.hash(), metadata, lock_script_hash, parameters, amount, registrar)?),
             Transaction::AssetTransfer {
+                ref burns,
```
sgkim126: You can skip `ref` because it's reference binding.

sgkim126: I don't think this must be fixed in this patch. It's just for your information.

PR #251, opened 2018-06-22
Update document

ScarletBlue:
Add information about blockchain configuration and tendermint configuration. Also add information about how to check if they are configured correctly.


kseo:
@ScarletBlue Please merge two commits into one.

PR #269, opened 2018-06-25
Split State

sgkim126:
This patch prepares sharding.


sgkim126:
@kseo @joojis I'll merge it without review. The most part of this code is copy and paste.

PR #297, opened 2018-06-29
Add Sanity check for unlock script

### LINKED ISSUE #280: Sanity check for unlock script
Unlock script is expected to be sane only if it's composed of PUSH-related opcodes, and any other types of unlock script must be rejected.
I think allowing only PUSH, PUSHB would be enough for initial implementation.

redongjun:
Fix #280 

- check valid instructions for unlock
- fix test cases (simple_burn and underflow)

### DIFF
--- a/vm/src/executor.rs
+++ b/vm/src/executor.rs
@@ -18,7 +18,7 @@ use ccrypto::{blake256, keccak256, ripemd160, sha256};
 use ckeys::{verify_ecdsa, ECDSASignature};
 use ctypes::{H256, H520, Public};
 
-use instruction::Instruction;
+use instruction::{is_valid_unlock_script, Instruction};
 
 const DEFAULT_MAX_MEMORY: usize = 1024;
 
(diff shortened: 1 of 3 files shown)


vm/src/executor.rs
```diff
@@ -147,6 +147,11 @@ pub fn execute(
     config: Config,
 ) -> Result<ScriptResult, RuntimeError> {
     // FIXME: don't merge scripts
+
+    if is_valid_unlock_script(unlock) == false {
```
Kais-DkM: Let's use `!` instead of comparing with `false`.

vm/src/instruction.rs
```diff
@@ -36,3 +36,14 @@ pub enum Instruction {
...
+
+pub fn is_valid_unlock_script(instrs: &[Instruction]) -> bool {
+    for instr in instrs {
+        match instr {
+            Instruction::Push(_) => {}
+            Instruction::PushB(_) => {}
+            _ => return false,
+        }
+    }
+    true
```
Kais-DkM: I think we can use `any` or `all` in `Iterator`.
e.g. `instrs.iter().all(|inst| ...)`

redongjun: I tried it, It's not possible. 
would you advise me?

redongjun: Thank you, but I'm not sure what's better than the old one.

Kais-DkM: I initially thought it would become simpler, but it seems not for this case.
I'd prefer functional style, but I guess it's up to you.
I'll accept either `iter` or `for` syntax for here.

redongjun:
@Kais-DkM Thank you for review

PR #333, opened 2018-07-06
Add a default shard

sgkim126:


sgkim126:
@joojis @kseo @Kais-DkM This patch adds a default shard to genesis state.
It will make the existing test scenarios work well, but it must be changed when we make the test net or main net.

sgkim126:
I'll merge it. It doesn't change the current behavior.

PR #347, opened 2018-07-09
Rename Payment's value field to amount

joojis:


kseo:
@joojis Do we need to change the SDK too?

PR #364, opened 2018-07-11
Add network id to AssetMint transaction

majecty:


core/src/parcel.rs
```diff
@@ -39,6 +39,8 @@ pub enum ParcelError {
     TooCheapToReplace,
     /// Invalid chain ID given.
     InvalidNetworkId,
+    /// Transaction's chain ID is different with parcel's
+    InvalidTransactionNetworkId,
```
kseo: Why don't we reuse `InvalidNetworkId`?

core/src/parcel.rs
```diff
@@ -78,7 +78,7 @@ pub fn parcel_error_message(error: &ParcelError) -> String {
         AlreadyImported => "Already imported".into(),
         Old => "No longer valid".into(),
         TooCheapToReplace => "Gas price too low to replace".into(),
-        InvalidNetworkId => "Parcel of this network ID is not allowed on this chain.".into(),
+        InvalidNetworkId => "Network ID is not allowed on this chain.".into(),
```
kseo: This network ID is not allowed on this chain

majecty: @kseo Fixed it

PR #366, opened 2018-07-11
Edit genesis account's initial balance

GNUp:
Change balance value 100 to 1000000


kseo:
@GNUp Please change the commit log to

Increase initial account balance

PR #373, opened 2018-07-11
Add net_getPeerCount, net_getPort, net_getNetworkId to RPC spec

majecty:


majecty:
@joojis I add the "(not implemented)"

kseo:
@joojis @majecty The naming is not consistent. Some APIs have get prefix, but the suggested APIs do not have get prefix. What's our convention?

majecty:
@kseo Like you said, It is better to use “get” as prefix.
@joojis I’ll change the api names.

Issue #220, opened 2018-06-20
Compress snapshot files


Kais-DkM:
Snapshot sync protocol requires its content to be compressed, but our snapshot files are saved in uncompressed format.
Since compressing is quite computationally expensive operation, it would be better to save them in compressed format rather than compressing it every time before sending snapshot message.

gurrpi:
I figured out how to fix this! I'll solve this issue.

PR #377, opened 2018-07-11
Compress snapshot files

gurrpi:
Fix: #220


kseo:
@gurrpi Please fix build.

PR #382, opened 2018-07-12
Throw error if spec file's common params are different from genesis block

majecty:


core/src/client/client.rs
```diff
@@ -98,9 +98,9 @@ impl Client {
 
         let journal_db = journaldb::new(db.clone(), journaldb::Algorithm::Archive, ::db::COL_STATE);
         let mut state_db = StateDB::new(journal_db, config.state_cache_size);
+        state_db = spec.ensure_genesis_state(state_db, &trie_factory)?;
         if state_db.journal_db().is_empty() {
```
sgkim126: Is it possible that `state_db` is empty after the `ensure_genesis_state` is called?

sgkim126: The `journal_db` is empty until the `journal_under` is called.

PR #388, opened 2018-07-13
Make UntrustedRlp::at not to read over prefixed length

majecty:
RlpStream's append modify internal state about list length.
So if you want to just one entity, you should use append_internal

### DIFF
--- a/util/rlp/src/untrusted_rlp.rs
+++ b/util/rlp/src/untrusted_rlp.rs
@@ -201,6 +201,9 @@ where
             return Err(DecoderError::RlpExpectedToBeList)
         }
 
+        let payload_info = self.payload_info()?;
+        let offset_max = payload_info.header_len + payload_info.value_len - 1;
+
         // move to cached position if its index is less or equal to
         // current search index, otherwise move to beginning of list
         let c = self.offset_cache.get();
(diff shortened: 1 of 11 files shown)


util/rlp/src/untrusted_rlp.rs
```diff
@@ -424,4 +435,29 @@ mod tests {
         let res: Result<u8, DecoderError> = rlp.as_val();
         assert_eq!(Err(DecoderError::RlpInvalidLength), res);
     }
+
+    #[test]
+    fn at_overflow() {
+        // [1,2,3]
+        let bs = [0xc3, 0x01, 0x02, 0x03];
+        let rlp = UntrustedRlp::new(&bs);
+        let first_element: Result<u8, _> = rlp.at(0).and_then(|elem| elem.as_val());
```
sgkim126: `and_then(|elem| elem.as_val());` -> `and_then(UntrustedRlp::as_val);`

util/rlp/src/untrusted_rlp.rs
```diff
@@ -424,4 +435,29 @@ mod tests {
         let res: Result<u8, DecoderError> = rlp.as_val();
         assert_eq!(Err(DecoderError::RlpInvalidLength), res);
     }
+
+    #[test]
+    fn at_overflow() {
+        // [1,2,3]
+        let bs = [0xc3, 0x01, 0x02, 0x03];
```
sgkim126: How about
```rust
let bs = vec![0x01, 0x02, 0x03].rlp_bytes();
```

majecty: 👍

PR #406, opened 2018-07-16
Fix wrong hardcoded directory path in run_account_command

joojis:


kseo:
@joojis I don't think this is the right approach. `account` subcommand should not share command line options with the main program.

joojis:
@kseo Ok, I will change it not to share, but just add the dedicated CLI option.

kseo:
@joojis Squash last two commits into one.

PR #417, opened 2018-07-17
Fixed #386 that adding optional block number param to getAsset() RPC method

### LINKED ISSUE #386: Add optional block number param to getAsset() RPC method
getAsset() gets the asset from the latest state only.
Since asset is a part of the blockchain state, it should be gettable from the specific state.

See getBalance and getNonce which already have the optional block number param.

itsukis:
Add a third parameter, block number, to the getAsset () RPC method.

### DIFF
--- a/core/src/client/client.rs
+++ b/core/src/client/client.rs
@@ -265,8 +265,8 @@ impl AssetClient for Client {
         }
     }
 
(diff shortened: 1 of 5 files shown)


joojis:
Looks good to me. Thank you for your contribution.

PR #436, opened 2018-07-19
Extract StateError from CoreError

sgkim126:


sgkim126:
@Kais-DkM Now we have StateError. It's a sum of TrieError, ParcelError and TransactionError.

Issue #235, opened 2018-06-21
Simplify Merkle Patricia Trie


kseo:
Simplify MPT by removing Extension node and adding path to Branch node.

sgkim126:
https://github.com/CodeChain-io/codechain/wiki/Merkle-Trie

sgkim126:
#508 fixes it

PR #437, opened 2018-07-19
Add Simplified Merkle Patricia Trie

GNUp:
issue #235

### DIFF
--- a/util/merkle/Cargo.toml
+++ b/util/merkle/Cargo.toml
@@ -4,7 +4,13 @@ version = "0.1.0"
 authors = ["CodeChain Team <codechain@kodebox.io>"]
 
 [dependencies]
-codechain-crypto = { path="../../crypto" }
+elastic-array = "0.9"
+rand = "0.5.3"
+hashdb = {path = "../hashdb" }
+codechain-crypto = { path = "../../crypto" }
+primitives = { path = "../primitives" }
+rlp = {path = "../rlp" }
 
 [dev-dependencies]
-primitives = { path = "../primitives" }
+trie-standardmap = { path = "../trie-standardmap" }
+memorydb = {path = "../memorydb" }
--- a/util/merkle/src/lib.rs
+++ b/util/merkle/src/lib.rs
@@ -15,10 +15,74 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 extern crate codechain_crypto as ccrypto;
+extern crate elastic_array;
+extern crate hashdb;
+#[cfg(test)]
+extern crate memorydb;
+extern crate primitives;
+extern crate rlp;
 
 #[cfg(test)]
-extern crate primitives;
+extern crate trie_standardmap as standardmap;
 
+use std::fmt;
+
+use hashdb::DBValue;
+use primitives::H256;
+
+mod nibbleslice;
+pub mod node;
 mod skewed;
+pub mod triedbmut;
 
 pub use skewed::skewed_merkle_root;
+
+/// Trie Errors.
+///
+/// These borrow the data within them to avoid excessive copying on every
+/// trie operation.
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum TrieError {
+    /// Attempted to create a trie with a state root not in the DB.
+    InvalidStateRoot(H256),
+    /// Trie item not found in the database,
+    IncompleteDatabase(H256),
+}
+
+impl fmt::Display for TrieError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            TrieError::InvalidStateRoot(root) => write!(f, "Invalid state root: {}", root),
+            TrieError::IncompleteDatabase(missing) => write!(f, "Database missing expected key: {}", missing),
+        }
+    }
+}
+
+/// Trie result type. Boxed to avoid copying around extra space for `H256`s on successful queries.
+pub type Result<T> = ::std::result::Result<T, Box<TrieError>>;
+
+
+/// A key-value datastore implemented as a database-backed modified Merkle tree.
+pub trait TrieMut {
+    /// Return the root of the trie.
+    fn root(&self) -> &H256;
+
+    /// Is the trie empty?
+    fn is_empty(&self) -> bool;
+
+    /// Does the trie contain a given key?
+    fn contains(&self, key: &[u8]) -> Result<bool> {
+        self.get(key).map(|x| x.is_some())
+    }
+
+    /// What is the value of the given key in this trie?
+    fn get(&self, key: &[u8]) -> Result<Option<DBValue>>;
+
+    /// Insert a `key`/`value` pair into the trie. An empty value is equivalent to removing
+    /// `key` from the trie. Returns the old value associated with this key, if it existed.
+    fn insert(&mut self, key: &[u8], value: &[u8]) -> Result<Option<DBValue>>;
+
+    /// Remove a `key` from the trie. Equivalent to making it equal to the empty
+    /// value. Returns the old value associated with this key, if it existed.
+    fn remove(&mut self, key: &[u8]) -> Result<Option<DBValue>>;
+}
--- a/util/merkle/src/nibbleslice.rs
+++ b/util/merkle/src/nibbleslice.rs
@@ -1,0 +1,274 @@
+// Copyright 2018 Kodebox, Inc.
+// This file is part of CodeChain.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+use std::cmp::*;
+use std::fmt;
+
+use elastic_array::ElasticArray36;
+
+
+#[derive(Eq, Ord)]
+pub struct NibbleSlice<'a> {
+    pub data: &'a [u8],
+    pub offset: usize,
+}
+
+impl<'a, 'view> NibbleSlice<'a>
+where
+    'a: 'view,
+{
+    /// Create a new nibble slice with the given byte-slice.
+    pub fn new(data: &'a [u8]) -> Self {
+        NibbleSlice::new_offset(data, 0)
+    }
+
+    /// Create a new nibble slice with the given byte-slice with a nibble offset.
+    pub fn new_offset(data: &'a [u8], offset: usize) -> Self {
+        NibbleSlice {
+            data,
+            offset,
+        }
+    }
+
+    /// Create a new nibble slice from the given HPE encoded data (e.g. output of `encoded()`).
+    pub fn from_encoded(data: &'a [u8]) -> NibbleSlice {
+        let offset = if data[0] & 16 == 16 {
+            1
+        } else {
+            2
+        };
+        Self::new_offset(data, offset)
+    }
+
+    /// Is this an empty slice?
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Get the length (in nibbles, naturally) of this slice.
+    pub fn len(&self) -> usize {
+        self.data.len() * 2 - self.offset
+    }
+
+    /// Get the nibble at position `i`.
+    pub fn at(&self, i: usize) -> u8 {
+        if (self.offset + i) & 1 == 1 {
+            self.data[(self.offset + i) / 2] & 15u8
+        } else {
+            self.data[(self.offset + i) / 2] >> 4
+        }
+    }
+
+    /// Return object which represents a view on to this slice (further) offset by `i` nibbles.
+    pub fn mid(&'view self, i: usize) -> NibbleSlice<'a> {
+        NibbleSlice {
+            data: self.data,
+            offset: self.offset + i,
+        }
+    }
+
+    /// Do we start with the same nibbles as the whole of `them`?
+    pub fn starts_with(&self, them: &Self) -> bool {
+        self.common_prefix(them) == them.len()
+    }
+
+    /// How many of the same nibbles at the beginning do we match with `them`?
+    pub fn common_prefix(&self, them: &Self) -> usize {
+        let s = min(self.len(), them.len());
+        let mut i = 0usize;
+        while i < s {
+            if self.at(i) != them.at(i) {
+                break
+            }
+            i += 1;
+        }
+        i
+    }
+
+    /// Encode while nibble slice in prefixed hex notation, noting whether it `is_leaf`.
+    pub fn encoded(&self) -> ElasticArray36<u8> {
+        let l = self.len();
+        let mut r = ElasticArray36::new();
+        let mut i = l % 2;
+        r.push(if i == 1 {
+            0x10 + self.at(0)
+        } else {
+            0
+        });
+        while i < l {
+            r.push(self.at(i) * 16 + self.at(i + 1));
+            i += 2;
+        }
+        r
+    }
+
+
+    /// Encode only the leftmost `n` bytes of the nibble slice in prefixed hex notation,
+    /// noting whether it `is_leaf`.
+    pub fn encoded_leftmost(&self, n: usize) -> ElasticArray36<u8> {
+        let l = min(self.len(), n);
+        let mut r = ElasticArray36::new();
+        let mut i = l % 2;
+        r.push(if i == 1 {
+            0x10 + self.at(0)
+        } else {
+            0
+        });
+        while i < l {
+            r.push(self.at(i) * 16 + self.at(i + 1));
+            i += 2;
+        }
+        r
+    }
+
+    pub fn to_vec(&self) -> Vec<u8> {
+        let mut vec: Vec<u8> = Vec::new();
+        for i in 0..self.len() {
+            vec.push(self.at(i));
+        }
+        vec
+    }
+
+    pub fn from_vec(v: Vec<u8>) -> (ElasticArray36<u8>, usize) {
+        let mut r = ElasticArray36::new();
+        let l = v.len();
+        let mut i = l % 2;
+        r.push(if i == 1 {
+            0x10 + v[0] & 15u8
+        } else {
+            0
+        });
+        while i < l {
+            r.push(((v[i] & 15u8) << 4) + (v[i + 1] & 15u8));
+            i += 2;
+        }
+        (r, 2 - (l % 2))
+    }
+}
+
+impl<'a> PartialEq for NibbleSlice<'a> {
+    fn eq(&self, them: &Self) -> bool {
+        self.len() == them.len() && self.starts_with(them)
+    }
+}
+
+impl<'a> PartialOrd for NibbleSlice<'a> {
+    fn partial_cmp(&self, them: &Self) -> Option<Ordering> {
+        let s = min(self.len(), them.len());
+        let mut i = 0usize;
+        while i < s {
+            match self.at(i).partial_cmp(&them.at(i)).unwrap() {
+                Ordering::Less => return Some(Ordering::Less),
+                Ordering::Greater => return Some(Ordering::Greater),
+                _ => i += 1,
+            }
+        }
+        self.len().partial_cmp(&them.len())
+    }
+}
+
+impl<'a> fmt::Debug for NibbleSlice<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        for i in 0..self.len() {
+            match i {
+                0 => write!(f, "{:01x}", self.at(i))?,
+                _ => write!(f, "'{:01x}", self.at(i))?,
+            }
+        }
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::NibbleSlice;
+    use elastic_array::ElasticArray36;
+
+    static D: &'static [u8; 3] = &[0x01u8, 0x23, 0x45];
+
+    #[test]
+    fn basics() {
+        let n = NibbleSlice::new(D);
+        assert_eq!(n.len(), 6);
+        assert!(!n.is_empty());
+
+        let n = NibbleSlice::new_offset(D, 6);
+        assert!(n.is_empty());
+
+        let n = NibbleSlice::new_offset(D, 3);
+        assert_eq!(n.len(), 3);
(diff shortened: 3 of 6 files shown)


util/merkle/src/lib.rs
```diff
@@ -14,11 +14,86 @@
...
+        match *self {
+            TrieError::InvalidStateRoot(ref root) => write!(f, "Invalid state root: {}", root),
+            TrieError::IncompleteDatabase(ref missing) => write!(f, "Database missing expected key: {}", missing),
+        }
+    }
+}
+
+impl error::Error for TrieError {
+    fn description(&self) -> &str {
+        match *self {
```
Kais-DkM: Dereference is not required here.

util/merkle/src/lib.rs
```diff
@@ -14,11 +14,86 @@
...
+    InvalidStateRoot(H256),
+    /// Trie item not found in the database,
+    IncompleteDatabase(H256),
+}
+
+impl fmt::Display for TrieError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            TrieError::InvalidStateRoot(ref root) => write!(f, "Invalid state root: {}", root),
+            TrieError::IncompleteDatabase(ref missing) => write!(f, "Database missing expected key: {}", missing),
```
Kais-DkM: You can use if as
```
match self {
    TrieError::InvalidStateRoot(root) => ...,
    ...
}

File: util/merkle/src/lib.rs [hunk 3]
```diff
@@ -14,11 +14,86 @@
...
+        }
+    }
+}
+
+impl error::Error for TrieError {
+    fn description(&self) -> &str {
+        match *self {
+            TrieError::InvalidStateRoot(_) => "Invalid state root",
+            TrieError::IncompleteDatabase(_) => "Incomplete database",
```
Kais-DkM: Why are arguments ignored here?

util/merkle/Cargo.toml
```diff
@@ -4,7 +4,13 @@ version = "0.1.0"
...
 
 [dependencies]
-codechain-crypto = { path="../../crypto" }
+elastic-array = "0.9"
+rand = "0.5.3"
+hashdb = { version = "0.1.1", path = "../hashdb" }
+codechain-crypto = { path = "../../crypto" }
+primitives = { path = "../primitives" }
+rlp = { version = "0.2.1", path = "../rlp" }
+memorydb = { version = "0.1.0", path = "../memorydb" }
```
Kais-DkM: `memorydb` crate is dev dependency.

util/merkle/src/lib.rs
```diff
@@ -14,11 +14,86 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-extern crate codechain_crypto as ccrypto;
+extern crate codechain_crypto;
+extern crate elastic_array;
+extern crate hashdb;
+extern crate memorydb;
```
Kais-DkM: Add `#[cfg(test)]` here.

util/merkle/src/nibbleslice.rs
```diff
@@ -0,0 +1,300 @@
...
+            } else {
+                self.data_suffix[(self.offset_suffix + i) / 2] >> 4
+            }
+        }
+    }
+
+    /// Return object which represents a view on to this slice (further) offset by `i` nibbles.
+    pub fn mid(&'view self, i: usize) -> NibbleSlice<'a> {
+        let l = self.data.len();
+
```
sgkim126:
```rust
let (offset, offset_suffix) = if l > i {
    (self.offset + i, self.offset_suffix)
} else {
    (self.offset + l, self.offset_suffix + i - l)
};
NibbleSlice {
    data: self.data,
    offset,
    idx: self.idx,
    data_suffix: self.data_suffix,
    offset_suffix,
}
```

util/merkle/src/node.rs
```diff
@@ -0,0 +1,111 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use elastic_array::ElasticArray36;
```
sgkim126: https://github.com/CodeChain-io/codechain/wiki/Coding-Style#imports-order

util/merkle/src/nibbleslice.rs
```diff
@@ -0,0 +1,300 @@
...
+        NibbleSlice::new_offset(data, 0)
+    }
+
+    /// Create a new nibble slice with the given byte-slice with a nibble offset.
+    pub fn new_offset(data: &'a [u8], offset: usize) -> Self {
+        NibbleSlice {
+            data,
+            offset,
+            idx: 16,
+            data_suffix: &b""[..],
```
Kais-DkM: What is this?

util/merkle/Cargo.toml
```diff
@@ -4,7 +4,13 @@ version = "0.1.0"
...
 
 [dev-dependencies]
-primitives = { path = "../primitives" }
+trie-standardmap = { path = "../trie-standardmap" }
```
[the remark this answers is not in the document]
GNUp: It is used in test code

util/merkle/src/nibbleslice.rs
```diff
@@ -0,0 +1,300 @@
...
+        assert_eq!(n.mid(1).encoded(), ElasticArray36::from_slice(&[0x11, 0x23, 0x45]));
+    }
+
+    #[test]
+    fn from_encoded() {
+        let n = NibbleSlice::new(D);
+        assert_eq!(n, NibbleSlice::from_encoded(&[0x00, 0x01, 0x23, 0x45]));
+        assert_eq!(n, NibbleSlice::from_encoded(&[0x20, 0x01, 0x23, 0x45]));
+        assert_eq!(n.mid(1), NibbleSlice::from_encoded(&[0x11, 0x23, 0x45]));
+        assert_eq!(n.mid(1), NibbleSlice::from_encoded(&[0x31, 0x23, 0x45]));
```
Kais-DkM: Are both `31 23 45` and `11 23 45` valid encoding of same value?
I think there should be only one valid encoding for consistent hash value.

util/merkle/src/node.rs
```diff
@@ -0,0 +1,113 @@
...
+                return Some(Node::Leaf(slice, DBValue::from_slice(&r.at(1).data())))
+            }
+            // branch node - first is nibbles (or empty), the rest 16 are nodes.
+            Prototype::List(17) => {
+                let mut nodes = [None; 16];
+                for i in 0..16 {
+                    nodes[i] = if r.at(i + 1).is_empty() {
+                        None
+                    } else {
+                        Some(r.at(i + 1).as_val::<H256>())
```
Kais-DkM: Use `val_at` instead of `at(..).as_val<..>()`

util/merkle/src/nibbleslice.rs
```diff
@@ -0,0 +1,300 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use elastic_array::ElasticArray36;
+use std::cmp::*;
+use std::fmt;
+
+#[derive(Eq, Ord)]
```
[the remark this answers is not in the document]
GNUp: There are partialEq and partialOrd. That was enought

util/merkle/src/nibbleslice.rs
GNUp: That was wrong test case. I change it correctly

util/merkle/src/node.rs
```diff
@@ -0,0 +1,110 @@
...
+#[derive(Eq, PartialEq, Debug)]
+pub enum Node<'a> {
+    Leaf(NibbleSlice<'a>, DBValue),
+
+    Branch(NibbleSlice<'a>, [Option<H256>; 16]),
+}
+
+impl<'a> Node<'a> {
+    /// Decode the `node_rlp` and return the Node.
+    pub fn decoded(node_rlp: &'a [u8]) -> Option<Self> {
```
Kais-DkM: Implement `Decodable` and `Encodable` instead of implementing customized functions.

util/merkle/src/node.rs
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    /// Encode the node into RLP.
+    pub fn encoded(node: Option<Self>, size: Option<usize>) -> ElasticArray1024<u8> {
+        match node {
+            Some(Node::Leaf(slice, value)) => {
+                let mut stream = RlpStream::new_list(2);
+                stream.append(&&*slice.encoded());
+                stream.append(&&*value);
```
Kais-DkM: Could you explain these `&&*`s?

util/merkle/src/node.rs
```diff
@@ -0,0 +1,110 @@
...
+                            if let Some(hash) = child {
+                                stream.append(hash);
+                            } else {
+                                stream.append_empty_data();
+                            }
+                        }
+                        stream.drain()
+                    }
+                    Some(s) => {
+                        stream.append(&&*slice.encoded_leftmost(s));
```
Kais-DkM:
```
let path = match size {
    None => &&*slice.encoded(),
    Some(s) => &&*slice.encoded_leftmost(s),
};
stream.append(path);
...
```

util/merkle/src/triedbmut.rs
```diff
@@ -0,0 +1,539 @@
...
+                            Ok(Some(v))
+                        } else {
+                            Ok(None)
+                        }
+                    }
+                    Some(RlpNode::Branch(partial, key)) => {
+                        if path.starts_with(&partial) {
+                            self.get_aux(path.mid(partial.len() + 1), key[path.mid(partial.len()).at(0) as usize])
+                        } else {
+                            Ok(None)
```
Kais-DkM: If this occurs, it means database is corrupted.
I think we should return some kind of error in this case.

util/merkle/src/triedbmut.rs
```diff
@@ -0,0 +1,539 @@
...
+                match RlpNode::decoded(&node_rlp[..]) {
+                    Some(RlpNode::Leaf(partial, v)) => {
+                        if partial == path {
+                            Ok(Some(v))
+                        } else {
+                            Ok(None)
+                        }
+                    }
+                    Some(RlpNode::Branch(partial, key)) => {
+                        if path.starts_with(&partial) {
```
Kais-DkM: Is this check necessary?
It seems quite inefficient to run this check on every tree traversal....

util/merkle/src/triedbmut.rs
```diff
@@ -0,0 +1,539 @@
...
+                    Some(RlpNode::Leaf(partial, v)) => {
+                        // Renew the Leaf
+                        if partial == path {
+                            let node = RlpNode::Leaf(path, v_);
+                            let node_rlp = RlpNode::encoded(Some(node), None);
+                            let hash = self.db.insert(&node_rlp[..]);
+
+                            *old_val = Some(v);
+
+                            self.db.remove(&cur_node_hash.expect("Can not change Leaf"));
```
Kais-DkM: Node removal is handled outside of `TrieDB`. You must not remove any nodes while insertion.

util/merkle/src/node.rs
GNUp: stream.append get argument as trait, encoding and encoding implement on &[u8]. It is trick to take the [u8] array out of ElasticArray. If we use only one ampersand we'll get ref of ElasticArray

util/merkle/src/node.rs
GNUp: you mean trait encoding and decoding in util/rlp/src/trait.rs?

Kais-DkM: Yes. That's more idiomatic way for handling rlp encoding in our code.

util/merkle/src/node.rs
Kais-DkM: `ElasticArray<T>` implements `AsRef<[T]>`.
Can you check again if single ampersand really doesn't work?

util/merkle/src/lib.rs
```diff
@@ -15,10 +15,77 @@
...
 #[cfg(test)]
+extern crate memorydb;
 extern crate primitives;
+extern crate rlp;
+
+#[cfg(test)]
+extern crate trie_standardmap as standardmap;
+
 
+use hashdb::DBValue;
```
sgkim126:
```rust
use std::fmt;

use hashdb::DBValue;
use primitives::H256;
```

util/merkle/src/node.rs
sgkim126: No. It can't. Unlike `Deref` `AsRef` doesn't work implicitly.
You should call it `&&*s` or `&s.as_ref()`

util/merkle/src/triedbmut.rs
```diff
@@ -0,0 +1,539 @@
...
+        let mut root = H256::new();
+        let mut t = TrieDBMut::new(&mut db, &mut root);
+        for &(ref key, ref value) in &x {
+            assert!(t.insert(key, value).unwrap().is_none());
+            assert_eq!(t.insert(key, value).unwrap(), Some(DBValue::from_slice(value)));
+        }
+
+        for (key, value) in x {
+            assert_eq!(t.remove(&key).unwrap(), Some(DBValue::from_slice(&value)));
+            assert!(t.remove(&key).unwrap().is_none());
```
sgkim126:
```rust
assert_eq!(Ok(None), t.remove(&key));
```

util/merkle/src/triedbmut.rs
```diff
@@ -0,0 +1,539 @@
...
+        let mut db = MemoryDB::new();
+        let mut root = H256::new();
+        let mut t = TrieDBMut::new(&mut db, &mut root);
+        for &(ref key, ref value) in &x {
+            assert!(t.insert(key, value).unwrap().is_none());
+            assert_eq!(t.insert(key, value).unwrap(), Some(DBValue::from_slice(value)));
+        }
+
+        for (key, value) in x {
+            assert_eq!(t.remove(&key).unwrap(), Some(DBValue::from_slice(&value)));
```
sgkim126:
```rust
assert_eq!(Ok(Some(DBValue::from_slice(&value))), t.remove(&key));
```

sgkim126: Two code works same when the test passes, but it gives more information when the test fails.

util/merkle/src/triedbmut.rs
```diff
@@ -0,0 +1,539 @@
...
+            journal_key: 0,
+            value_mode: ValueMode::Index,
+            count: 4,
+        }.make_with(&mut seed);
+
+        let mut db = MemoryDB::new();
+        let mut root = H256::new();
+        let mut t = TrieDBMut::new(&mut db, &mut root);
+        for &(ref key, ref value) in &x {
+            assert!(t.insert(key, value).unwrap().is_none());
```
sgkim126:
```rust
assert_eq!(Ok(None), t.insert(key, value));
```

util/merkle/src/triedbmut.rs
```diff
@@ -0,0 +1,539 @@
...
+            assert_eq!(t.insert(key, value).unwrap(), Some(DBValue::from_slice(value)));
```
sgkim126:
```rust
assert_eq!(Ok(Some(DBValue::from_slice(value))), t.insert(key, value));
```

util/merkle/src/node.rs
sgkim126: I'm not sure about that since `encoded` function takes additional parameter, `size`.

util/merkle/src/triedbmut.rs
```diff
@@ -0,0 +1,539 @@
...
+        old_val: &mut Option<DBValue>,
+    ) -> super::Result<Option<H256>> {
+        match cur_node_hash {
+            Some(hash) => {
+                let node_rlp = self.db.get(&hash).ok_or_else(|| Box::new(TrieError::IncompleteDatabase(hash)))?;
+
+                match RlpNode::decoded(&node_rlp[..]) {
+                    Some(RlpNode::Leaf(partial, v)) => {
+                        if path == partial {
+                            self.db.remove(&cur_node_hash.expect("Can not remove leaf"));
```
sgkim126: You MUST NOT remove any item in `db`.
The `remove` function just change the shape of trie.

util/merkle/src/triedbmut.rs
```diff
@@ -0,0 +1,539 @@
...
+                    }
+                    None => Ok(cur_node_hash),
+                }
+            }
+            None => Ok(cur_node_hash),
+        }
+    }
+
+    /// Print Trie in format
+    pub fn fmt(&self, cur_node_hash: H256) {
```
sgkim126:
```rust
pub fn fmt(&self, cur_node_hash: &H256) {
```

util/merkle/src/triedbmut.rs
```diff
@@ -0,0 +1,539 @@
...
+        let node_rlp = self.db.get(&cur_node_hash).expect("Broken Database");
+
+        let node = RlpNode::decoded(&node_rlp[..]);
+        println!("{:?}", node);
+        match node {
+            Some(RlpNode::Branch(_partial, key)) => {
+                for i in 0..16 {
```
sgkim126:
```rust
            Some(RlpNode::Branch(_partial, children)) => {
                for child in children.iter() {
                    match child {
                        None => continue,
                        Some(h) => self.fmt(h),
                    }
                }
            }
```

util/merkle/src/triedbmut.rs
```diff
@@ -0,0 +1,539 @@
...
+            None => Ok(cur_node_hash),
+        }
+    }
+
+    /// Print Trie in format
+    pub fn fmt(&self, cur_node_hash: H256) {
+        let node_rlp = self.db.get(&cur_node_hash).expect("Broken Database");
+
+        let node = RlpNode::decoded(&node_rlp[..]);
+        println!("{:?}", node);
```
sgkim126: Use [Formatter](https://doc.rust-lang.org/std/fmt/struct.Formatter.html) instead of `println`.

util/merkle/src/triedbmut.rs
```diff
@@ -0,0 +1,539 @@
...
+use hashdb::DBValue;
+use hashdb::HashDB;
+use primitives::H256;
+
+use super::node::Node as RlpNode;
+use super::{TrieError, TrieMut};
+use nibbleslice::NibbleSlice;
+
+
+fn empty_child() -> [Option<H256>; 16] {
```
sgkim126: `empty_children`

util/merkle/src/node.rs
Kais-DkM: What a shame :(

util/merkle/src/node.rs
Kais-DkM: In that case, we can have two functions, one without size argument and the other accepting additional parameter.
Former one would be implementation of `Encodable`.

util/merkle/src/triedbmut.rs
```diff
@@ -0,0 +1,539 @@
...
+            db,
+            root,
+        })
+    }
+
+    /// Insert auxiliary
+    fn insert_aux(
+        &mut self,
+        path: NibbleSlice,
+        v_: DBValue,
```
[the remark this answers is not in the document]
GNUp: I'll change all variable name which is ambiguous

util/merkle/src/node.rs
GNUp: it gives me this error
error[E0597]: borrowed value does not live long enough
  --> util/merkle/src/node.rs:78:33
   |
78 |                     Some(s) => &&*slice.encoded_leftmost(s),
   |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^-
   |                                 |                         |
   |                                 |                         temporary value dropped here while still borrowed
   |                                 temporary value does not live long enough
...
90 |             }
   |             - temporary value needs to live until here

util/merkle/src/triedbmut.rs
GNUp: when I test with original one, i got this error
```
failures:

---- triedbmut::tests::return_old_values stdout ----
thread 'triedbmut::tests::return_old_values' panicked at 'assertion failed: `(left == right)`
  left: `None`,
 right: `Some([128])`', util/merkle/src/triedbmut.rs:537:13
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```
and using code you gave me, I got similar one
```
failures:

---- triedbmut::tests::return_old_values stdout ----
thread 'triedbmut::tests::return_old_values' panicked at 'assertion failed: `(left == right)`
  left: `Ok(Some([128]))`,
 right: `Ok(None)`', util/merkle/src/triedbmut.rs:537:13
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```
Could you explain why you recommend ```assert_eq!(Ok(Some(DBValue::from_slice(&value))), t.remove(&key));``` or you can give me a keywords to search

sgkim126: When `remove` returns `Err(_)`, the assetion is in `Result` so it doesn't print `util/merkle/src/triedbmut.rs:537:13` but prints `some path to result.rs:...:..`.

util/merkle/src/triedbmut.rs
GNUp: I can not understand what you mean. So I just make 
```

impl<'a> fmt::Display for RlpNode<'a> {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        match self {
            RlpNode::Leaf(partial, value) => writeln!(f, "Leaf - key({:?}), value({:?})", partial, value),
            RlpNode::Branch(partial, children) => {
                writeln!(f, "Branch - path({:?})", partial)?;

                for i in 0..16 {
                    writeln!(f, "child {} - hash({:?})", i, children[i])?;
                }
                Ok(())
            }
        }
    }
}
```
and print node with `println!`

util/merkle/src/node.rs
GNUp: This might cause large change of design so I think it's better to keep this. Additionally, I split `encoded` into two separate function

util/merkle/src/triedbmut.rs
GNUp: I remove all `db.remove()` however I wonder why there is `db.remove()` in util/patricia_trie/triedbmut?

util/merkle/src/lib.rs
```diff
@@ -15,10 +15,76 @@
...
+
+    /// Does the trie contain a given key?
+    fn contains(&self, key: &[u8]) -> Result<bool> {
+        self.get(key).map(|x| x.is_some())
+    }
+
+    /// What is the value of the given key in this trie?
+    fn get<'a, 'key>(&'a self, key: &'key [u8]) -> Result<Option<DBValue>>
+    where
+        'a: 'key;
```
sgkim126: It seems that this lifetime dependency is not necessary.

PR #447, opened 2018-07-20
Prioritize parcels according to fee per byte

majecty:


core/src/miner/mem_pool.rs
```diff
@@ -86,6 +86,8 @@ struct ParcelOrder {
     nonce_height: U256,
     /// Fee of the parcel.
     fee: U256,
+    /// Size of parcel's rlp serialized data
+    serialized_size: usize,
```
sgkim126: How about add `fee_per_bytes` field instaed of `serialized_size`, if it's only used to calculate `fee_per_byte`

majecty: 👍

majecty:
@sgkim126 Fixed

core/src/miner/mem_pool.rs
```diff
@@ -1239,4 +1248,65 @@ pub mod test {
 
         assert_eq!(fee + amount, item.cost());
     }
+
+    #[test]
+    fn test_fee_per_byte_order_simple() {
+        let order1 = create_parcel_order(U256::from(1000_000_000), 100);
+        let order2 = create_parcel_order(U256::from(1500_000_000), 200);
+        println!("{} {}", order1.fee_per_byte, order2.fee_per_byte);
```
kseo: Remove println here.

PR #464, opened 2018-07-24
Prevent the same transaction from executing twice

sgkim126:


core/src/client/mod.rs
```diff
@@ -72,6 +72,15 @@ pub trait ParcelInfo {
 
 pub trait TransactionInfo {
     fn transaction_parcel(&self, id: TransactionId) -> Option<ParcelAddress>;
+
+    fn is_duplicated<'a>(&self, transactions: &'a mut Iterator<Item = H256>) -> bool {
```
kseo: `is_duplicated` -> `is_already_included`

types/src/parcel/error.rs
```diff
@@ -74,6 +76,7 @@ impl Display for Error {
     fn fmt(&self, f: &mut Formatter) -> FormatResult {
         let msg: String = match self {
             Error::AlreadyImported => "Already imported".into(),
+            Error::AlreadyIncluded => "The transaction is already included".into(),
```
kseo: `AlreadyIncluded` -> `ParcelAlreadyIncluded`

types/src/parcel/error.rs
```diff
@@ -74,6 +76,7 @@ impl Display for Error {
     fn fmt(&self, f: &mut Formatter) -> FormatResult {
         let msg: String = match self {
             Error::AlreadyImported => "Already imported".into(),
```
kseo: `AlreadyImported` -> `ParcelAlreadyImported`

types/src/parcel/error.rs
kseo: The transaction is already included -> Transaction already imported

core/src/client/mod.rs
sgkim126: How about `is_any_transaction_included`?

core/src/client/mod.rs
```diff
@@ -69,6 +70,19 @@ pub trait ParcelInfo {
     fn parcel_block(&self, id: ParcelId) -> Option<H256>;
 }
 
+pub trait TransactionInfo {
+    fn transaction_parcel(&self, id: TransactionId) -> Option<ParcelAddress>;
+
+    fn is_any_transaction_included<'a>(&self, transactions: &'a mut Iterator<Item = H256>) -> bool {
```
sgkim126: @kseo How about `is_any_transaction_included`?

kseo: @sgkim126 Sounds good.

PR #469, opened 2018-07-24
Remove type parameter from TopLevelState

Kais-DkM:


core/src/spec/spec.rs
```diff
@@ -377,7 +375,8 @@ fn load_from(s: cjson::spec::Spec) -> Result<Spec, Error> {
     match g.state_root {
         Some(root) => *s.state_root_memo.get_mut() = root,
         None => {
-            let db = BasicBackend(MemoryDB::new());
+            let memorydb = Arc::new(::kvdb_memorydb::create(0));
+            let db = StateDB::new(journaldb::new(memorydb, Algorithm::Archive, None), 0);
```
sgkim126: This pattern is repeated in `core::tests::helpers::get_temp_state_db` and `state::tests::helpers::get_temp_state_db`. How about extract it as a function? I think `StateDB::new_memory(size_of_cache)` looks fine.

Kais-DkM: Agreed. I'll add it.

PR #483, opened 2018-07-25
Add support for custom action

Kais-DkM:


state/src/item/cache.rs
```diff
@@ -30,6 +31,13 @@ pub trait CacheableItem: Clone + fmt::Debug + Decodable + Encodable {
     fn is_null(&self) -> bool;
 }
 
+impl CacheableItem for Bytes {
+    type Address = H256;
+    fn is_null(&self) -> bool {
+        self.len() == 0
```
sgkim126:
```rust
self.is_empty()
```

codechain/config/chain_type.rs
```diff
@@ -66,17 +68,17 @@ impl fmt::Display for ChainType {
 }
 
 impl ChainType {
-    pub fn spec<'a>(&self) -> Result<Spec, String> {
+    pub fn spec<'a>(&self, handlers: &[Arc<ActionHandler>]) -> Result<Spec, String> {
```
sgkim126: Why do you use slice?
How about using `Vec<Arc<ActionHandler>>`?

core/src/spec/spec.rs
```diff
@@ -350,7 +373,7 @@ impl Spec {
 }
 
 /// Load from JSON object.
-fn load_from(s: cjson::spec::Spec) -> Result<Spec, Error> {
+fn load_from(s: cjson::spec::Spec, handlers: &[Arc<ActionHandler>]) -> Result<Spec, Error> {
```
sgkim126: IMO `handlers` must be Vec. Passing it as slice and converting it to `Vec` is redundant.

state/src/action_handler/hit.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl ActionHandler for HitHandler {
+    fn init(&self, state: &mut TrieMut) -> StateResult<()> {
+        let r = state.insert(&self.address(), &0u32.rlp_bytes());
+        debug_assert_eq!(Ok(None), r);
+        r?;
+        Ok(())
+    }
+
+    fn is_target(&self, bytes: &Bytes) -> bool {
```
sgkim126: How about remove this method

state/src/action_handler/hit.rs
```diff
@@ -0,0 +1,84 @@
...
+        HitAction::decode(&UntrustedRlp::new(bytes)).is_ok()
+    }
+
+    /// `bytes` must be valid encoding of HitAction
+    fn execute(&self, bytes: &Bytes, state: &mut TopLevelState) -> StateResult<Outcome> {
```
sgkim126: and make this function return `Option<StateResult<Outcome>>`?

state/src/impls/top_level.rs
```diff
@@ -410,6 +410,15 @@ impl TopLevelState {
                     error: None,
                 })
             }
+            Action::Custom(bytes) => {
+                let handlers = self.db.custom_handlers().to_vec();
```
sgkim126: Why do you convert it do Vec?

core/src/client/client.rs
```diff
@@ -449,6 +452,10 @@ impl BlockChainClient for Client {
             })
         })
     }
+
+    fn custom_handlers(&self) -> Vec<Arc<ActionHandler>> {
```
sgkim126: Why noy `fn custom_handlers(&self) -> &[Arc<ActionHandler>>] {`

state/src/impls/top_level.rs
Kais-DkM: `custom_handlers` returns reference, so `handlers` borrows `self`.
`self` is used as mutable reference in later code, so lifetime conflicts there.
So I wanted to clone `handlers`, and I thought `to_vec` would be a good solution.

core/src/client/client.rs
Kais-DkM: If we return reference, then lock over `state_db` won't be released while holding that reference.
I thought it would be simpler to return a fresh copy of the vector and release the lock earlier instead of holding the lock.

state/src/action_handler/hit.rs
Kais-DkM: We need some sort of function that only checks if provided byte array is valid one that this handler can decode, without execution. It's used for blocking invalid parcels in sync and rpc.
Since `execute` requires `TopLevelState` as argument, I think it's not a good idea to merge these.

core/src/client/client.rs
sgkim126: Okay. I see.

PR #489, opened 2018-07-26
Add error codes to JSON-RPC spec

joojis:
https://github.com/joojis/codechain/blob/rpc-error/spec/JSON-RPC.md

### DIFF
--- a/spec/JSON-RPC.md
+++ b/spec/JSON-RPC.md
@@ -94,6 +94,29 @@ A base32 string that starts with "ccc" or "tcc". See [the specification](https:/
 ## Signature
 `H520` for ECDSA signature | `H512` for Schnorr signature
 
+# Error codes
+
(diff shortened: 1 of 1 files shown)


joojis:
@majecty Could you review the error spec of the net_* methods?

joojis:
@kseo Fixed.

spec/JSON-RPC.md
```diff
@@ -246,6 +246,12 @@ Params:
 
 Return Type: `null` | `H256`
 
+Errors:
+
+| Code | Message | Description |
+|---|---|---|
+| -32602 | Invalid Params | At least one of the parameters is invalid |
```
kseo: Why don't you add error codes section above and mention only the error codes here instead of repeating the same descriptions over and over again?

PR #495, opened 2018-07-26
Use FullAddress instead of Address for account RPC APIs

joojis:


key/src/address.rs
```diff
@@ -70,17 +70,19 @@ fn rearrange_bits(data: &[u8], from: usize, into: usize) -> Vec<u8> {
...
         data.extend(&self.address.to_vec());
-        let encode_result = Bech32 {
+        let mut encoded = Bech32 {
             hrp,
             data: rearrange_bits(&data, 8, 5),
-        }.to_string();
-        write!(f, "{}", encode_result.unwrap())
+        }.to_string()
+            .unwrap();
+        encoded.remove(3);
```
kseo: What does this line do?

key/src/address.rs
```diff
@@ -33,6 +35,25 @@ pub struct FullAddress {
     pub address: Address,
 }
 
+impl FullAddress {
```
kseo: `fmt::Display` implementation for `FullAddress` looks almost the same. What's the difference here?

key/src/address.rs
```diff
@@ -69,20 +90,7 @@ fn rearrange_bits(data: &[u8], from: usize, into: usize) -> Vec<u8> {
 
 impl fmt::Display for FullAddress {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let hrp = match self.network {
```
kseo: Oh, now I see. Ignore the comment above.

key/src/address.rs
```diff
@@ -90,10 +92,15 @@ impl FromStr for FullAddress {
     fn from_str(s: &str) -> Result<Self, Error>
     where
         Self: Sized, {
-        let decoded = Bech32::from_string(s.to_string())?;
+        if s.len() < 7 {
+            return Err(Error::Bech32InvalidLength)
+        }
+        let mut encoded = s.to_string();
+        encoded.insert(3, '1');
```
kseo: What does this line do?

key/src/address.rs
```diff
@@ -36,6 +36,19 @@ pub struct FullAddress {
 }
 
 impl FullAddress {
+    pub fn create_version0(network_id: u64, address: Address) -> Result<Self, Error> {
```
kseo: Why don't you pass `0` as an argument?

key/src/address.rs
joojis: This is the separator of Bech32. I had to decide whether to change Bech32 implementation or not, and I decided not to change because it's more simple.

key/src/address.rs
joojis: The arguments can be different for the different version, in my expectation.

PR #498, opened 2018-07-27
Make ShardId 16 bits

sgkim126:
This PR makes ShardId 16 bits and changes the mechanism to create asset/asset scheme address.
Currently, CodeChain generates 256 bits hash and replaces some bits with prefix and shard id and world id.
This patch changes CodeChain generates 208 bits hash instead of 256 bits hash since 48 bits will be removed.


sgkim126:
This patch must go with https://github.com/CodeChain-io/codechain-sdk-js/pull/93

PR #499, opened 2018-07-27
Add chain_getTransaction RPC spec

joojis:


joojis:
I will merge this after https://github.com/CodeChain-io/codechain/pull/489

PR #500, opened 2018-07-27
Change the account command so that it uses FullAddress instead of Add…

joojis:
…ress

### DIFF
--- a/codechain/account_command.rs
+++ b/codechain/account_command.rs
@@ -20,13 +20,14 @@ use std::str::FromStr;
 use rpassword;
 
 use ccore::AccountProvider;
-use ckey::{Address, Private};
+use ckey::{FullAddress, Private};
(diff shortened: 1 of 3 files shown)


codechain/account_command.rs
```diff
@@ -37,6 +37,8 @@ pub fn run_account_command(matches: ArgMatches) -> Result<(), String> {
     clogger::init(&LoggerConfig::new(0)).expect("Logger must be successfully initialized");
 
     let keys_path = matches.value_of("keys-path").unwrap_or(DEFAULT_KEYS_PATH);
+    // FIXME: Do not hardcode
+    let network_id = 0x11;
```
kseo: Move `network_id` to `constants` module.

joojis:
@kseo I updated the patch.

I think it's necessary to pass something about network.
One of the alternatives is using Network enum instead of network_id which is just a u64 constant.

PR #502, opened 2018-07-27
Add get_with() function and make TrieFactory and TrieKinds

GNUp:


util/merkle/src/lib.rs
```diff
@@ -66,6 +66,26 @@ impl fmt::Display for TrieError {
 /// Trie result type. Boxed to avoid copying around extra space for `H256`s on successful queries.
 pub type Result<T> = ::std::result::Result<T, Box<TrieError>>;
 
+/// Description of what kind of query will be made to the trie.
+pub trait Query {
```
Kais-DkM: Why do you need this trait?
Isn't `Decodable` doing the job for this kind of property?

util/merkle/src/lib.rs
```diff
@@ -109,3 +135,87 @@ pub trait TrieMut {
...
+pub enum TrieKinds<'db> {
+    /// A generic trie db.
+    Generic(TrieDB<'db>),
+}
+
+// wrapper macro for making the match easier to deal with.
+macro_rules! wrapper {
+	($me: ident, $f_name: ident, $($param: ident),*) => {
+		match *$me {
+			TrieKinds::Generic(ref t) => t.$f_name($($param),*),
```
Kais-DkM: I think following code works here.
```
match $me {
    TrieKinds::Generic(t) => ...
}
```

util/merkle/src/lib.rs
```diff
@@ -109,3 +135,87 @@ pub trait TrieMut {
     /// value. Returns the old value associated with this key, if it existed.
     fn remove(&mut self, key: &[u8]) -> Result<Option<DBValue>>;
 }
+
+
+/// Trie types
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum TrieSpec {
+    /// Generic trie.
+    Generic,
```
Kais-DkM: Why did you leave `Generic` instead of `Secure`?

util/merkle/src/lib.rs
sgkim126: It's used in `get_with` method. A `get_with` is more flexible than just returning the decoded value.
It receives the bytes and returning any type that can be made by the bytes.

I think this flexibility is more than we want, and we can refactor it to return the specified type.
But it should be another patch, since replacing implementation and interface in one commit makes review hard.

sgkim126:
LGTM. Let's merge it after apply @Kais-DkM 's comments.

util/merkle/src/lib.rs
GNUp: Secure will be removed, so I think i doesn't matter

util/merkle/src/lib.rs
GNUp: I changed it

util/merkle/src/lib.rs
Kais-DkM: Confirmed.

PR #509, opened 2018-07-31
Change how to calculate an address of a state

sgkim126:


sgkim126:
Sorry. I misunderstood blake2b. I thought the blake2b runs less round if the output length is short. But it's not. Blake2b executes the same round regardless of the output length.

PR #510, opened 2018-07-31
Remove UNSIGNED_SENDER in SignedParcel

majecty:
The UNSIGNED_SENDER is introduced in EIP 86.
The purpose of UNSIGNED_SENDER is paying a transaction fee
from a contract. In codechain, we do not support smart contract.
So remove UNSIGNED_SENDER.


majecty:
@kseo I fixed your request

PR #513, opened 2018-07-31
Add chain_isAssetSpent RPC API

joojis:
SDK patch: https://github.com/CodeChain-io/codechain-sdk-js/pull/99

### DIFF
--- a/core/src/client/client.rs
+++ b/core/src/client/client.rs
@@ -277,6 +277,55 @@ impl AssetClient for Client {
             Ok(None)
         }
     }
+
+    /// Checks whether an asset is spent or not.
+    ///
+    /// It returns None if such an asset never existed in the shard at the given block.
+    fn is_asset_spent(
+        &self,
+        transaction_hash: H256,
+        index: usize,
+        shard_id: ShardId,
+        block_id: BlockId,
+    ) -> TrieResult<Option<bool>> {
+        match self.transaction_address(transaction_hash.into()) {
+            Some(ref transaction_address)
+                if self.block_number(block_id)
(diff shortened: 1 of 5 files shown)


core/src/client/client.rs
```diff
@@ -277,6 +277,42 @@ impl AssetClient for Client {
...
+
+                if index >= output_length {
+                    return Ok(None)
+                }
+
+                match Client::state_at(&self, id) {
+                    Some(state) => {
+                        let shard_id = 0;
+                        let address = AssetAddress::new(transaction_hash, index, shard_id);
+                        match state.asset(shard_id, &address)? {
```
```rust
Ok(Some(state.asset(shard_id, &address)?.is_none()))
```

joojis: That sounds better. I will update.

joojis:
I'm going to change this patch not to hardcode a shard id. Please be patient.

joojis:
I updated the patch, and, I updated the SDK patch also. 
https://github.com/CodeChain-io/codechain-sdk-js/pull/99

core/src/client/client.rs
```diff
@@ -277,6 +277,52 @@ impl AssetClient for Client {
...
         }
     }
+
+    fn is_asset_spent(
+        &self,
+        transaction_hash: H256,
+        index: usize,
+        shard_id: ShardId,
+        block_id: BlockId,
+    ) -> TrieResult<Option<bool>> {
```
sgkim126: Would you add a comment about the meaning of output?

joojis: I added the comment explaining the case of None.

PR #516, opened 2018-08-01
Verify sum of inputs/outputs of AssetTransferTransaction when queuing a parcel

majecty:
I'll check the burns after https://github.com/CodeChain-io/codechain-sdk-js/issues/103 is solved

### DIFF
--- a/core/src/parcel.rs
+++ b/core/src/parcel.rs
@@ -141,6 +141,7 @@ impl UnverifiedParcel {
                 ..
             } => {
                 for t in transactions {
+                    t.verify()?;
                     match &t {
                         Transaction::AssetMint {
                             network_id,
--- a/state/src/impls/shard_level.rs
+++ b/state/src/impls/shard_level.rs
@@ -15,7 +15,6 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 use std::cell::RefMut;
-use std::collections::HashMap;
 use std::fmt;
 
 use ccrypto::{Blake, BLAKE_NULL_RLP};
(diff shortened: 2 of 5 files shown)


core/src/parcel.rs
```diff
@@ -156,11 +156,17 @@ impl UnverifiedParcel {
...
                         Transaction::AssetTransfer {
                             network_id,
+                            inputs,
+                            outputs,
                             ..
                         } => {
                             if network_id != &self.network_id {
                                 return Err(ParcelError::InvalidNetworkId)
                             }
+                            // TODO: check burns
```
kseo: TODO -> FIXME

types/src/parcel/error.rs
```diff
@@ -73,6 +73,8 @@ pub enum Error {
     RegularKeyAlreadyInUse,
     RegularKeyAlreadyInUseAsMaster,
     InvalidTransferDestination,
+    /// Returned when the sum of the transaction's input is different from the sum of outputs.
```
kseo: input -> inputs

state/src/impls/shard_level.rs
```diff
@@ -106,6 +105,7 @@ impl<B: Backend + ShardBackend> ShardLevelState<B> {
     }
 
     fn apply_internal(&mut self, transaction: &Transaction) -> StateResult<()> {
+        assert!(transaction.verify().is_ok());
```
```rust
assert_eq!(Ok((), transaction.verify());
```

types/src/parcel/error.rs
```diff
@@ -73,6 +74,8 @@ pub enum Error {
     RegularKeyAlreadyInUse,
     RegularKeyAlreadyInUseAsMaster,
     InvalidTransferDestination,
+    /// Transaction error
+    InvalidTransaction(String),
```
sgkim126: How about `InvalidTransaction(TransactionError)`?

types/src/transaction/transaction.rs
```diff
@@ -163,6 +165,52 @@ impl Transaction {
...
     }
+
+    pub fn verify(&self) -> Result<(), Error> {
+        match self {
+            Transaction::AssetTransfer {
+                inputs,
+                outputs,
+                ..
+            } => {
+                // FIXME: check burns
```
joojis: @majecty I'm just curious. Which verifications should we do here?

majecty: Check whether inputs and burns have common input or not.

PR #525, opened 2018-08-02
Asset compose

joojis:


sgkim126:
LGTM

state/src/item/asset_scheme.rs
```diff
@@ -21,19 +21,59 @@ use rlp::{Decodable, DecoderError, Encodable, RlpStream, UntrustedRlp};
...
+        })
+    }
+}
+
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 pub struct AssetScheme {
     metadata: String,
     amount: u64,
     registrar: Option<Address>,
+    components: Vec<AssetComponent>,
```
kseo: Let's call this `pool`.

state/src/item/asset_scheme.rs
```diff
@@ -21,19 +21,59 @@ use rlp::{Decodable, DecoderError, Encodable, RlpStream, UntrustedRlp};
 
 use super::cache::CacheableItem;
 
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct AssetComponent {
```
kseo: Why don't we reuse `Asset` instead of introducing a new type `AssetComponent`?

joojis:
I rebased this patch.

PR #526, opened 2018-08-03
Add error codes to each stratum API

### LINKED ISSUE #492: Add error codes to each stratum API

redongjun:
Fix #492

The error description has been changed to match the error used in the codechain-core crate and the unused error was deleted.

### DIFF
--- a/stratum/src/lib.rs
+++ b/stratum/src/lib.rs
@@ -181,22 +181,24 @@ impl StratumImpl {
 
     /// rpc method `mining.submit`
     fn submit(&self, params: Params, meta: SocketMetadata) -> RpcResult {
-        params
-            .parse::<(H256, Vec<String>)>()
-            .map(|(pow_hash, seal)| {
-                let seal = seal.iter().cloned().map(Into::into).collect();
-                match self.dispatcher.submit((pow_hash, seal)) {
-                    Ok(()) => {
-                        self.update_peers(&meta.tcp_dispatcher.expect("tcp_dispatcher is always initialized"));
-                        to_value(true)
-                    }
-                    Err(submit_err) => {
-                        cwarn!(STRATUM, "Error while submitting share: {:?}", submit_err);
-                        to_value(false)
-                    }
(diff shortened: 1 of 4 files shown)


stratum/src/lib.rs
```diff
@@ -181,22 +181,24 @@ impl StratumImpl {
...
+        if workers.contains_key(&meta.addr) == false {
+            return Err(Error::UnauthorizedWorker.into())
+        }
+
+        params.parse::<(H256, Vec<String>)>().and_then(|(pow_hash, seal)| {
+            let seal = seal.iter().cloned().map(Into::into).collect();
+            match self.dispatcher.submit((pow_hash, seal)) {
+                Ok(()) => {
+                    self.update_peers(&meta.tcp_dispatcher.expect("tcp_dispatcher is always initialized"));
+                    Ok(jsonrpc_core::Value::Bool(true))
```
kseo: It always returns `true`. Let's change the return type to `void`.

stratum/src/lib.rs
```diff
@@ -526,4 +528,34 @@ mod tests {
             .expect("Response should be utf-8");
         assert_eq!("{\"jsonrpc\":\"2.0\",\"result\":true,\"id\":2}\n", response);
     }
+
+    #[test]
+    fn can_occur_error_for_unauthorized_worker() {
```
kseo: returns_error_when_unauthorized_worker_submits

stratum/src/traits.rs
```diff
@@ -17,11 +17,16 @@
 use std;
 use std::error::Error as StdError;
 
+use jsonrpc_core::{Error as JsonError, ErrorCode as JsonErrorCode};
 use jsonrpc_tcp_server::PushMessageError;
 use primitives::{Bytes, H256};
 
 #[derive(Debug, Clone)]
 pub enum Error {
+    InternalError,
```
kseo: Remove `Dispatch` error message which is no longer used.

Issue #222, opened 2018-06-20
Introduce sharding - 4th wave


sgkim126:
Static shard validators
* [x] Implement World related features
* [x] Implement parcel handling with validators
* [x] Implement extension for validators

PR #529, opened 2018-08-03
Introduce World

sgkim126:


sgkim126:
It's part of #222

sgkim126:
@kseo @xpdlf1004 This patch breaks backward compatibility because 511a0c3461dafe0686c7cd50d03e8f44718a9dd3 changes the initial state. You cannot use the previously generated DB after this patch.

Issue #348, opened 2018-07-09
CodeChain crashes on SIGINT


sgkim126:
```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/codechain`
#469691130 2018-07-09 17:19:56  main INFO codechain  Starting client
#469691130 2018-07-09 17:19:57  main INFO codechain  Handshake Listening on 3485
#469691130 2018-07-09 17:19:57  main INFO discovery  Node runs with unstructured discovery
#469691130 2018-07-09 17:19:57  main INFO codechain  RPC Listening on 8080
#469691130 2018-07-09 17:19:57  IO Worker #1 INFO sync  Sync extension initialized
#469691130 2018-07-09 17:19:57  main INFO codechain  IPC Listening on /tmp/jsonrpc.ipc
#469691130 2018-07-09 17:19:57  main INFO test_script  Initialization complete
^Cpure virtual method called
terminate called without an active exception
Aborted (core dumped)
```

I bisected and it seems 9447945867d907589c9ef8313a3308e3467b4e37 caused this issue.

kseo:
@sgkim126 @majecty Any hint on this issue?

majecty:
If we turn on the stratum, the crash is not shown.
However, if we turn off the stratum, it occurred.
I'm struggling with this problem.

majecty:
I found a hint.
If the RPC struct has weak pointers of dependencies.
The crash disappeared.

I may send the pull request today.

majecty:
☕️

sgkim126:
I removed the Testnet milestone since there is the hotfix(#593). But I'll not close this issue until we find the reason and solve it.

sgkim126:
FYI, the hotfix makes CodeChain wait a second before shutdown. You can use the environment variable named `WAIT_BEFORE_SHUTDOWN` to change the waiting time.

sgkim126:
I found the reason.
It's because the main thread exits without waiting for the child threads.
It would be fixed if the main thread waits until the child threads join.

sgkim126:
Closed by #1498

sgkim126:
Reoccurred on the master branch.

sgkim126:
I'm suspecting that it's because of the rocksdb.
It can be fixed if we upgrade rocksdb.
Please try https://github.com/CodeChain-io/codechain/issues/1441 first.

foriequal0:
[...]

PR #537, opened 2018-08-04
Use weak reference in RPC structs

majecty:
Codechain crashes on SIGINT if stratum is off.
The cause of the crash may be the ownership.
So use weak reference in RPC structs.

After this patch, we cannot see the segfault on SIGINT.

This fixes https://github.com/CodeChain-io/codechain/issues/348


sgkim126:
@majecty I'm not sure about it.
I thought this issue is because of the cyclic reference, but I cannot find the cycle.

PR #545, opened 2018-08-04
Add net_getPeerCount RPC API

### LINKED ISSUE #368: Add JSON RPC API for querying number of peers currently connected to the client

conr2d:
Fixed #368

### DIFF
--- a/network/src/p2p/handler.rs
+++ b/network/src/p2p/handler.rs
@@ -339,6 +339,15 @@ impl Handler {
             max_peers,
         })
     }
+
+    pub fn get_port(&self) -> u16 {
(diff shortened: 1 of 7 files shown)


network/src/p2p/handler.rs
```diff
@@ -339,6 +339,15 @@ impl Handler {
             max_peers,
         })
     }
+
+    pub fn get_port(&self) -> IoHandlerResult<u16> {
```
sgkim126: Please make `get_port` return `u16` and `get_peer_count` return `usize`.
There is no reason to return `Result`.

PR #551, opened 2018-08-07
Pass the sender to ShardLevelState.

sgkim126:
These patches prepare the permission check.


sgkim126:
@joojis The first commit changes RPC interface, but it doesn't break the compatibility of SDK since SDK doesn't use it.

sgkim126:
I'll merge it when Travis passes.

sgkim126:
@majecty This patch changes the interface. You can implement the registrar checking on this patches.

Issue #474, opened 2018-07-24
Limit number of expensive operations in script


Kais-DkM:
Every opcodes in virtual machine has different computational complexity, but current code for sort/validation of transaction doesn't consider such aspects.
This can be a problem since transactions containing many expensive opcodes can make machine slower, providing opportunity for other attacks.
Simplest solution would be limiting number of complex operations allowed in single script.(e.g. allowing up to 6 `CHKSIG`s in one script)

sgkim126:
Fixed by #563

PR #563, opened 2018-08-08
Add logic to limit expensive operations in script

### LINKED ISSUE #473: Support partial signature for transactions
In Bitcoin, [OP_CHKSIG](https://en.bitcoin.it/wiki/OP_CHECKSIG) supports various kinds of ways to provide signature. In particular, it is possible to provide signature for only one transaction output, allowing other input/outputs to be added to transaction in future. We need similar feature for more flexible scripting.

junbeomlee:
fixed #473


junbeomlee:
Sorry, I think I wrote the issue number incorrectly. it was #474 not #473

PR #566, opened 2018-08-09
Check the permission before minting

sgkim126:


sgkim126:
@joojis This PR makes the minting fail if the sender is not one of the world owners.
I checked the integration test of SDK passes since the all account that test uses are world owners.

PR #578, opened 2018-08-13
Make RPC use PlatformAddress

sgkim126:

PR #579, opened 2018-08-13
Use platform addresses in the scheme file

joojis:


key/src/platform_address.rs
```diff
@@ -141,6 +141,12 @@ impl From<&'static str> for PlatformAddress {
     }
 }
 
+impl Into<Address> for PlatformAddress {
```
sgkim126: FYI, it will conflict with https://github.com/CodeChain-io/codechain/pull/578/commits/86328a85584a0b2af399b6818d7b49241e75f069

core/res/null.json
```diff
@@ -19,7 +19,7 @@
       "generic": "0x0"
     },
     "score": "0x20000",
-    "author": "0x0000000000000000000000000000000000000000",
+    "author": "cccqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq6n0r0d",
```
kseo: Why do you use `ccc` prefix here? It should start with `ccc` to be consistent with other files.

PR #580, opened 2018-08-14
Add a specification about a whitelist/blacklist

sgkim126:


sgkim126:
@kseo I'll implement these RPC API if you have no comment.

PR #585, opened 2018-08-14
Implement the errors of account RPC methods

joojis:


keystore/src/keystore.rs
```diff
@@ -64,7 +64,11 @@ impl KeyStore {
 
 impl SimpleSecretStore for KeyStore {
     fn insert_account(&self, secret: Secret, password: &Password) -> Result<Address, Error> {
-        self.store.insert_account(secret, password)
```
kseo: What does `store.insert_account` return when the account already exists?

joojis: I guess it's a legacy from Parity, the store is KeyMultiStore. KeyMultiStore can store multiple accounts with different passwords for one address. In that case, the store allows the address to use if one of the passwords is given.

PR #586, opened 2018-08-14
Don't hardcode the network id for the account CLI

joojis:


codechain/subcommand/account_command.rs
```diff
@@ -163,3 +165,22 @@ fn read_password_and_confirm() -> Option<Password> {
         None
     }
 }
+
+fn read_network_id(chain: &str) -> Result<NetworkId, String> {
+    let mut contents = String::new();
+    let reader: &[u8] = match chain {
```
kseo: Code duplication. Use `ChainType::scheme` method.

joojis: Thank you. I changed it.

PR #593, opened 2018-08-16
Wait before shutdown to avoid the crash

sgkim126:
This patch is the hotfix for #348.
I can't find the exact reason for the SIGINT yet, but this patch fixes
the crash.

### DIFF
--- a/codechain/main.rs
+++ b/codechain/main.rs
@@ -30,6 +30,7 @@ extern crate serde_json;
 extern crate app_dirs;
 extern crate codechain_core as ccore;
 extern crate codechain_discovery as cdiscovery;
+extern crate codechain_finally as cfinally;
 extern crate codechain_key as ckey;
 extern crate codechain_keystore as ckeystore;
 #[macro_use]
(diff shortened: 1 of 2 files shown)


codechain/codechain.yml
```diff
@@ -46,6 +46,11 @@ args:
         long: chain
         help: Set the blockchain type out of solo, solo_authority, tendermint, cuckoo, blake_pow, husky or a path to chain scheme file.
         takes_value: true
+    - wait-before-shutdown:
```
kseo: I don't think we need a command line option here because this PR is just a hotfix for the crash. Just use a hardcoded constant which is large enough to avoid the crash.

sgkim126: It's to help test.
I'm not sure waiting 1 second is enough to avoid the crash. It was the experimental result on my machine.

sgkim126: I think we need internal command line options to help test and debug.

kseo: It is not a good idea to expose a command line option that is used only for test & debug purposes. What about adding an environment variable instead? Users don't know what environment variables exist, but we can use them for test & debug purposes.

sgkim126: Okay. I agree with you. I'll change it.

sgkim126:
@kseo PR updated.

PR #596, opened 2018-08-16
Add snapshot subcommand

gurrpi:
This subcommand works only when a node isn't running.

Below error occurs when calling snapshot subcommand while a node is running. (As I know is that `ClientSerive` couldn't be running twice at once.)
```
hyundp:codechain daehyun$ ./target/debug/codechain snapshot generate
Error: "Client service error: Database error: IO error: While lock file: db/LOCK: Resource temporarily unavailable"
```

### DIFF
--- a/codechain/codechain.yml
+++ b/codechain/codechain.yml
@@ -247,4 +247,8 @@ subcommands:
                         help: Address of desired password change
                         required: true
                         index: 1
-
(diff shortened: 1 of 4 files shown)


kseo:
This PR is outdated and has conflicts. Please create a new PR.

PR #600, opened 2018-08-19
Introduce bytes type to deserialize vector of bytes

### LINKED ISSUE #599: Need to deserialize vector of bytes for stratum
The seal field of the submit method should be a vector of bytes.

redongjun:
Fix #599

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -473,6 +473,7 @@ name = "codechain-stratum"
 version = "1.11.0"
 dependencies = [
  "codechain-crypto 0.1.0",
+ "codechain-json 0.1.0",
  "codechain-logger 0.1.0",
  "env_logger 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


redongjun:
@keo I know it, but I added it because of the following problem:

```
error: cyclic package dependency: package `codechain-core v0.1.0 (file:///Users/redongjun/workspace/codechain/core)` depends on itself. Cycle:
package `codechain-core v0.1.0 (file:///Users/redongjun/workspace/codechain/core)`
    ... which is depended on by `codechain-rpc v0.1.0 (file:///Users/redongjun/workspace/codechain/rpc)`
    ... which is depended on by `codechain v0.1.0 (file:///Users/redongjun/workspace/codechain)`
```

Could you advise me?

kseo:
@redongjun Did you add `json` package to the dependencies of `stratum`?

redongjun:
@kseo I was mistaken for a different path (rpc/v1/type/bytes.rs). so the error (cyclic package dependency) occurred.

PR #601, opened 2018-08-19
Implement partial signing

GNUp:


kseo:
@GNUp The Travis CI build failed. Please fix it first.

types/src/transaction/transaction.rs
```diff
@@ -264,6 +268,77 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+    mut result: Vec<AssetTransferOutput>,
+) -> Vec<AssetTransferOutput> {
+    match bitmask {
+        0 => return result,
+        _ => {
+            if ((bitmask & 0x1) == 1) && !outputs.is_empty() {
+                result.push(outputs.remove(0));
+                apply_bitmask(bitmask >> 1, outputs, result)
+            } else {
+                apply_bitmask(bitmask >> 1, outputs, result)
```
sgkim126: I think you must remove output here.

types/src/transaction/transaction.rs
```diff
@@ -264,6 +268,77 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
 
+fn apply_bitmask(
+    bitmask: u128,
+    mut outputs: Vec<AssetTransferOutput>,
+    mut result: Vec<AssetTransferOutput>,
+) -> Vec<AssetTransferOutput> {
+    match bitmask {
+        0 => return result,
+        _ => {
+            if ((bitmask & 0x1) == 1) && !outputs.is_empty() {
```
sgkim126: I think it must be the error if the bit is set but the corresponding output does not exist.

vm/src/executor.rs
```diff
@@ -139,12 +140,24 @@ impl Stack {
     }
 }
 
-pub fn execute(
+fn make_bitmask(mut item: Vec<u8>, mut result: u128) -> u128 {
+    if item.is_empty() {
+        return result
+    } else {
```
sgkim126: Do not use `else` when the above statements always return.

types/src/transaction/transaction.rs
```diff
@@ -264,6 +268,77 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+                result.push(outputs.remove(0));
+                apply_bitmask(bitmask >> 1, outputs, result)
+            } else {
+                apply_bitmask(bitmask >> 1, outputs, result)
+            }
+        }
+    }
+}
+
+impl<'a> PartialHashing for &'a Transaction {
```
sgkim126: Why do you implement `PartialHashing` for the reference of `Transaction` instead of `Transaction`?

types/src/transaction/transaction.rs
```diff
@@ -264,6 +268,77 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+
+        match self {
+            Transaction::AssetTransfer {
+                network_id,
+                burns,
+                inputs,
+                outputs,
+                nonce,
+            } => {
+                let new_burns: Vec<_> = if input_scheme {
```
sgkim126: I think rust compiler can infer that this variable is a `Vec`.

types/src/transaction/transaction.rs
GNUp: Cause when `execute` is called, there is only a borrowed transaction we can use in `fn transfer_asset`.

vm/src/executor.rs
```diff
@@ -16,10 +16,11 @@
 
 use ccrypto::{blake256, keccak256, ripemd160, sha256};
 use ckey::{verify, Public, Signature, SIGNATURE_LENGTH};
-use primitives::H256;
+use ctypes::transaction::{AssetOutPoint, HashingError, PartialHashing};
 
 use instruction::{has_expensive_opcodes, is_valid_unlock_script, Instruction};
 
+
```
joojis: Don't include unnecessary change.

GNUp: Okay, I will pay more attention next time.

types/src/transaction/transaction.rs
```diff
@@ -264,6 +273,81 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+    mut outputs: Vec<AssetTransferOutput>,
+    mut result: Vec<AssetTransferOutput>,
+) -> Result<Vec<AssetTransferOutput>, HashingError> {
+    match bitmask {
+        0 => return Ok(result),
+        _ => {
+            if outputs.is_empty() {
+                return Err(HashingError::InvalidFilter)
+            }
+            if (bitmask & 0x1) == 1 {
```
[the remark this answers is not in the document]
sgkim126: FYI, it's bad for performance to remove the first item of a vector.

sgkim126:
Please add some test cases about partial signing.

types/src/transaction/transaction.rs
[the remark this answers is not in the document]
GNUp: Okay, Thank you

vm/src/executor.rs
```diff
@@ -47,6 +47,13 @@ pub enum RuntimeError {
     IndexOutOfBound,
     StackUnderflow,
     TypeMismatch,
+    InvalidFilter,
+}
+
+impl From<HashingError> for RuntimeError {
+    fn from(_error: HashingError) -> Self {
+        RuntimeError::InvalidFilter
```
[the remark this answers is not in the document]
GNUp: Okay, I didn't think about it.

types/src/transaction/transaction.rs
```diff
@@ -264,6 +232,148 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+    for i in 0.. {
+        if bitmask == 0 {
+            return Ok(result)
+        }
+
+        if outputs.is_empty() {
+            return Err(HashingError::InvalidFilter)
+        }
+
+        let mut output = outputs[i].clone();
```
sgkim126: 1. No reason to make `output` mutable.
2. No reason to clone `outputs[i]` if it's not pushed in the result.

```rust
if (bitmask & 0x1) == 1 {
    result.push(outputs[i].clone());
}
```

types/src/transaction/transaction.rs
```diff
@@ -264,6 +232,148 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+        match self {
+            Transaction::AssetTransfer {
+                network_id,
+                burns,
+                inputs,
+                outputs,
+                nonce,
+            } => {
+                let new_burns = if input_scheme {
+                    let mut tmp: Vec<_> = burns
```
sgkim126: Please use proper name instead of `tmp`

types/src/transaction/transaction.rs
GNUp: I resolved it

types/src/transaction/transaction.rs
GNUp: I resolved it

types/src/transaction/transaction.rs
```diff
@@ -754,4 +862,22 @@ mod tests {
...
+            outputs: Vec::new(),
+            nonce: 0,
+        };
+        let outpoint = AssetOutPoint {
+            transaction_hash: H256::default(),
+            index: 0,
+            asset_type: H256::default(),
+            amount: 0,
+        };
+        assert_eq!(transaction.hash_partially(5 as u128, &outpoint, false), Err(HashingError::InvalidFilter));
```
sgkim126: `0b101` instead of 5.
Using binary format is better since it's a bitmask.

GNUp: I resolved it

types/src/transaction/transaction.rs
```diff
@@ -264,6 +232,146 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+            result.push(outputs[i].clone());
+        }
+        bitmask = bitmask >> 1
+    }
+    Err(HashingError::InvalidFilter)
+}
+
+impl PartialHashing for Transaction {
+    fn hash_partially(&self, bitmask: u128, cur: &AssetOutPoint, burn: bool) -> Result<H256, HashingError> {
+        let input_scheme = (bitmask & 0x1) == 1;
```
sgkim126: `let sign_all_input = ...`

types/src/transaction/transaction.rs
```diff
@@ -264,6 +232,146 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+        let output_scheme = (bitmask >> 1 & 0x1) == 1;
```
sgkim126: `let sign_all_output = ...`

types/src/transaction/transaction.rs
```diff
@@ -264,6 +232,146 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+
+        match self {
+            Transaction::AssetTransfer {
+                network_id,
+                burns,
+                inputs,
+                outputs,
+                nonce,
+            } => {
+                let new_burns = if input_scheme {
```
sgkim126: It seems that 271~303 and 305~337 do the same thing.
Please extract it a function.

types/src/transaction/transaction.rs
```diff
@@ -264,6 +232,146 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+                    let mut burns_without_script: Vec<_> = burns
+                        .iter()
+                        .map(|input| AssetTransferInput {
+                            prev_out: input.prev_out.clone(),
+                            lock_script: Vec::new(),
+                            unlock_script: Vec::new(),
+                        })
+                        .collect();
+                    burns_without_script.push(AssetTransferInput {
```
sgkim126: IMO, appending one item for the signing all scheme seems too ad-hoc.
How about passing a bitmask to the blake function?

types/src/transaction/transaction.rs
GNUp: I changed it

types/src/transaction/transaction.rs
GNUp: I changed it

types/src/transaction/transaction.rs
sgkim126: FYI, There is a function named [blake256_with_key](https://github.com/CodeChain-io/codechain/blob/049e64991339fc5ec85bb0efcc08cbcb16993d08/crypto/src/blake.rs#L26)

types/src/transaction/transaction.rs
GNUp: I resolved it

types/src/transaction/transaction.rs
GNUp: I resolved it

types/src/transaction/transaction.rs
```diff
@@ -264,6 +232,115 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+            result.push(outputs[i].clone());
+        }
+        bitmask = bitmask >> 1
+    }
+    Err(HashingError::InvalidFilter)
+}
+
+fn apply_sign_scheme_to_input(
+    inputs: &Vec<AssetTransferInput>,
+    scheme: bool,
```
sgkim126: Rename scheme to is_sign_all

types/src/transaction/transaction.rs
```diff
@@ -264,6 +232,115 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+            .iter()
+            .map(|input| AssetTransferInput {
+                prev_out: input.prev_out.clone(),
+                lock_script: Vec::new(),
+                unlock_script: Vec::new(),
+            })
+            .collect()
+    }
+
+    if is_burn {
```
sgkim126: Would you explain what it is?

types/src/transaction/transaction.rs
```diff
@@ -264,6 +232,115 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+                outputs,
+                nonce,
+            } => {
+                let new_burns = apply_sign_scheme_to_input(burns, sign_all_inputs, burn, cur);
+                let new_inputs = apply_sign_scheme_to_input(inputs, sign_all_inputs, burn, cur);
+
+                let new_outputs = if sign_all_outputs {
+                    outputs
+                        .iter()
+                        .map(|output| AssetTransferOutput {
```
sgkim126: Why not `.map(Clone::clone)`?

sgkim126: It seems that it can be `outputs.clone()` if it can be `.map(Clone::clone).

types/src/transaction/transaction.rs
```diff
@@ -264,6 +232,115 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
     sum.iter().all(|(_, sum)| sum.is_zero())
 }
 
+fn make_tag(mut item: Vec<u8>, mut result: u128) -> u128 {
```
sgkim126: IMO, passing a vector reference and an index instead of a mutable vector will be better.

sgkim126: For instance, you can refactor this function like the below.
```rust
fn make_tag(item: &[u8]) -> u128 {
    if item.is_empty() {
        return 0
    }
    fn make_tag_internal(item: &[u8], index: usize, mut result: u128) -> u128 {
        if item.len() == index {
            return result
        }
        // Assume that the mask is stored as little endian
        let head = item[index] as u128;
        result = (result << 8) | head as u128;
        make_tag_internal(item, index + 1, result)
    }
    make_tag_internal(item, 0, 0)
}
```

types/src/transaction/transaction.rs
```diff
@@ -264,6 +232,115 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+    if item.is_empty() {
+        return result
+    }
+    // Assume that the mask is stored as little endian
```
sgkim126: Usually, a bitmask uses a big-endian. Why do you use a little-endian?

types/src/transaction/transaction.rs
GNUp: I changed it

types/src/transaction/transaction.rs
GNUp: It is my mistake. I didn't recover that from the old scheme. In the previous version I implemented, there was an additional element for storing filter in output and input. So I had to clone other elements one by one like this way. However, there is no need to do this now.

types/src/transaction/transaction.rs
GNUp: It also seems more elegant because the function's arguments are more intuitive. I replaced it with the code you provided.

types/src/transaction/transaction.rs
GNUp: I just specified the way how u128, the bitmask, is stored in Vec<u8>. I did it cause it was a more familiar way. I change it to stored as big-endian.

types/src/transaction/transaction.rs
```diff
@@ -264,6 +232,113 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
     sum.iter().all(|(_, sum)| sum.is_zero())
 }
 
+fn make_tag(item: &[u8]) -> u128 {
+    if item.is_empty() {
+        return 0
+    }
+    debug_assert!(item.len() == 16, "The size of filter vec should be 16");
```
sgkim126:
```rust
debug_assert_eq!(16, item.len());
```

state/src/impls/shard_level.rs
```diff
@@ -159,7 +159,12 @@ impl<B: Backend + ShardBackend> ShardLevelState<B> {
                 inputs,
                 outputs,
                 ..
-            } => self.transfer_asset(&transaction, sender, burns, inputs, outputs),
+            } => {
+                if outputs.len() > 126 {
```
sgkim126: I think it should be an assertion and verify method should fail.

types/src/transaction/transaction.rs
GNUp: I resolved it

state/src/impls/shard_level.rs
GNUp: I moved it to verify method

types/src/transaction/transaction.rs
```diff
@@ -224,6 +192,9 @@ impl Transaction {
                 outputs,
                 ..
             } => {
+                if outputs.len() > 26 {
```
sgkim126: 126

kseo:
@GNUp Please resolve conflicts.

joojis:
@GNUp Please update the SDK version to `0.4.0-rc1` for the integration tests

types/src/transaction/transaction.rs
sgkim126: It's not changed yet.

types/src/transaction/transaction.rs
```diff
@@ -346,6 +317,109 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+
+    loop {
+        if !(index < filter_len) {
+            return Ok(result)
+        }
+
+        let mut filter = bitmask.pop().unwrap();
+
+        for i in 0..8 {
+            if index == output_len as usize {
```
sgkim126: Move it out of for loop.

types/src/transaction/transaction.rs
```diff
@@ -346,6 +317,109 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
     sum.iter().all(|(_, sum)| sum.is_zero())
 }
 
+fn apply_bitmask_to_output(
+    mut bitmask: Vec<u8>,
+    filter_len: usize,
```
sgkim126: Why does it need `filter_len`?
Isn't it the length of `bitmask`?

types/src/transaction/transaction.rs
```diff
@@ -346,6 +317,109 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+                let new_burns = is_sign_all(burns, sign_all_inputs, is_burn, cur);
+                let new_inputs = is_sign_all(inputs, sign_all_inputs, !is_burn, cur);
+
+                let new_outputs = if sign_all_outputs {
+                    outputs.clone()
+                } else {
+                    if bitvec.len() != filter_len as usize {
+                        return Err(HashingError::InvalidFilter)
+                    }
+                    apply_bitmask_to_output(bitvec.clone(), filter_len as usize, outputs.clone(), Vec::new())?
```
sgkim126: What happens if the length of `outputs` is not the same with `bitvec.len()`

types/src/transaction/transaction.rs
```diff
@@ -346,6 +317,109 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+                result.push(outputs[index + i].clone());
+            }
+
+            filter = filter >> 1;
+        }
+        index += 1;
+    }
+}
+
+fn is_sign_all(
```
sgkim126: IMO, the name of this function is not matched with this behavior.

types/src/transaction/transaction.rs
GNUp: It is actually (index + i). my mistake, I changed it.

types/src/transaction/transaction.rs
GNUp: I removed that argument.

types/src/transaction/transaction.rs
GNUp: I defined some rules.

1. filter_len must be same with the bitvec's byte length which is the tag except for one byte for output scheme, input scheme and filter_len.
2. filter_len can be greater than output's length. Only exist outputs will be signed. So an output can be added if a signer wants to do. He just adds zero behind the filter. If he doesn't want to extend the outputs, he can add 1 behind the filter.
3. filter_len can be shorter than output's length. Only relevant outputs will be signed.

I am now writing the specification for partial signing. If you give me your opinion, I will summarize that with the above rules.

types/src/transaction/transaction.rs
GNUp: I misunderstand your request. I changed it.

types/src/transaction/transaction.rs
GNUp: I changed it to `apply_input_scheme`

types/src/transaction/transaction.rs
```diff
@@ -346,6 +317,107 @@ fn is_input_and_output_consistent(inputs: &[AssetTransferInput], outputs: &[Asse
...
+fn apply_bitmask_to_output(
+    mut bitmask: Vec<u8>,
+    outputs: Vec<AssetTransferOutput>,
+    mut result: Vec<AssetTransferOutput>,
+) -> Result<Vec<AssetTransferOutput>, HashingError> {
+    let mut index = 0;
+    let output_len = outputs.len();
+
+    loop {
+        if !((index as usize) < bitmask.len()) {
```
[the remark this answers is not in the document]
GNUp: Oh, I fixed it.

joojis:
Let's merge these patches at here https://github.com/CodeChain-io/codechain/pull/789

GNUp:
@joojis Okay, I will close this PR. @sgkim126

PR #604, opened 2018-08-20
Add commitHash RPC

majecty:


majecty:
@kseo I added info_getCommitHash to the spec file.

PR #635, opened 2018-08-23
Push a parcel only when verification passes

sgkim126:


sgkim126:
I updated the PR because I didn't run rustfmt.

PR #636, opened 2018-08-23
Add integration tests to Travis CI

joojis:


.travis.yml
```diff
@@ -18,6 +18,7 @@ jobs:
         - cargo fetch --verbose
       script:
         - cargo +nightly-2018-07-17 fmt -- --check && RUST_BACKTRACE=1 cargo test --verbose --all
+        - cargo build && cd test && npm install && npm run start
```
kseo: Don't we use `yarn` everywhere?

joojis: In the current configuration, `yarn` is not pre-installed in Travis. I'm looking for a solution.

joojis:
@kseo I updated it. Would you review this again?

PR #657, opened 2018-08-30
Add sync test with multiple nodes

jhs7jhs:
Currently, it fails often.

The condition is:
Connect 4 solo-mod nodes in a row without discovering, and let's number them 0 to 3 from left to right.
The node 0 receives one parcel, and the node 3 receives one parcel. Those two parcels are different.

Then, the node 0 and the node 3 creates blocks. Call them block A and B each.
The block A will be propagated from the node 0 to the node 3, and block B from the node 3 to the node 0.
And, in somewhere in the middle, the problem occurs. Let's assume it's the node 1.
Currently the node 1 has no node, and it receives the block A from the left and also receives the block B from the right. Then it stops.

```
[ Node 0 ] - [ Node 1 ] - [ Node 2 ] - [ Node 3 ]
     A     ->    ?      <-     B     <-     B
```

### DIFF
--- a/test/src/integration/sync.test.ts
+++ b/test/src/integration/sync.test.ts
@@ -171,6 +171,51 @@ describe("sync", () => {
       });
     });
 
+    describe("Connected in a circle", () => {
+      const numHalf: number = Math.floor(numNodes / 2);
+
+      beforeEach(async () => {
+        for (let i = 0; i < numNodes; i++) {
+          nodes[i].connect(nodes[(i + 1) % numNodes]);
+        }
+      }, 5000 + 1500 * numNodes);
+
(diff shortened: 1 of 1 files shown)


sgkim126:
@jhs7jhs 
What is the meaning of stop? Crash? Deadlock?

jhs7jhs:
@sgkim126 The test code checks the sync status by getting the best block number and the hash of the best block. In the example, the best block number of the node 1 never changes: it's always zero. Never getting the block A, and also the block B.
I think the word 'stop' was not good choice for explanation. :( I'm checking trace logs now.

jhs7jhs:
Now #658 is resolved.

PR #660, opened 2018-08-31
Remove downloded bodies from the targets properly

GNUp:
Wrong targets were being removed

### DIFF
--- a/sync/src/block/downloader/body.rs
+++ b/sync/src/block/downloader/body.rs
@@ -85,12 +85,16 @@ impl BodyDownloader {
 
     pub fn drain(&mut self) -> Vec<(H256, Vec<UnverifiedParcel>)> {
         let mut result = Vec::new();
-        for (target, ..) in &self.targets {
+        let mut new_targets = Vec::new();
+
+        for (target, prev_root, parcels_root) in &self.targets {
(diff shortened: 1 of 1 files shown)


sync/src/block/extension.rs
```diff
@@ -479,6 +479,7 @@ impl Extension {
                 _ => {}
             }
         }
+        // I think it already had to be removed from drain() in line 468
```
sgkim126: No. `remove_target` removes the downloading list also. But `drain` removes the downloaded list only.

sgkim126: By the way, I think this change itself looks good.

sgkim126: Would you cleanup the PR and remove the WIP tag?

GNUp: Okay

Kais-DkM:
This patch is nice, but I didn't experience any problem related to this before.
Could you provide a situation that this patch solves?

sync/src/block/extension.rs
GNUp: `exists` were removed from downloading list in `import_bodies()`

GNUp:
@Kais-DkM I just read the code

GNUp:
@Kais-DkM I checked it with integration test

Issue #658, opened 2018-08-30
Fail to sync when a node receives blocks with the same difficulty


jhs7jhs:
It's an assumption now, and you can check the problem in #657.

Need to check for finding specific condition:
- Does it occur only when the receiving node has no blocks?

jhs7jhs:
1. `BodyDownloader` has the queue to check what targets are being downloaded now. If a target is already in the queue, it never sends a request about the target again. There's no timeout.
2. When `BodyDownloader` makes a request, it puts all the targets which is not in the downloading queue and the downloaded queue.
3. When sending requests to download targets' information, the sync agent iterates its peer list and sends the request which `BodyDownloader` made. But as 1. and 2., the agent only sends the request to the first peer in the peer list, unless some targets are added while iterating the peer list.
4. If there are headers with the same score, only one is selected to be downloaded and others are omitted.

1. ~ 3. can make a sync disorder, so it needs to be fixed, and with 4.,  this issue occurs.

jhs7jhs:
Fixed in #661

PR #661, opened 2018-09-04
Send body request in random order with the length limit

jhs7jhs:
This is the fix of #658. Also, it will stabilize sync.
In synchronization, a client will send body requests to its peers in random order to prevent asking specific hashes to one fixed peer. Also, a client limits the length of each body request, therefore it will send hashes evenly to its peers. Finally, the downloading queue of the body downloader will be flushed every timeout.

### DIFF
--- a/sync/src/block/downloader/body.rs
+++ b/sync/src/block/downloader/body.rs
@@ -37,10 +37,14 @@ impl BodyDownloader {
     }
 
     pub fn create_request(&mut self) -> Option<RequestMessage> {
+        const MAX_BODY_REQEUST_LENGTH: usize = 128;
         let mut hashes = Vec::new();
         for (hash, ..) in &self.targets {
             if !self.downloading.contains(hash) && !self.downloaded.contains_key(hash) {
                 hashes.push(*hash);
+            }
+            if hashes.len() >= MAX_BODY_REQEUST_LENGTH {
+                break
             }
         }
         if hashes.len() != 0 {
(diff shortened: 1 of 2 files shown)


jhs7jhs:
Not finished yet. Still fixing.

sync/src/block/extension.rs
```diff
@@ -140,39 +150,57 @@ impl NetworkExtension for Extension {
...
-                    })
-                } else {
-                    false
-                }
-            };
-            if !have_body_request && peer_score > total_score {
-                if let Some(request) = self.body_downloader.lock().create_request() {
-                    self.send_request(&id, request);
+            SYNC_EXPIRE_REQUEST_TOKEN => {
+                if let Some((id, request_id)) = self.body_requests_to_expire.write().pop_front() {
```
sgkim126:
```rust
let mut body_requests_to_expire = self.body_requests_to_expire.write().;
let requests = self.requests.read();
if let Some((id, request_id)) = body_requests_to_expire.pop_front() {
   if let Some(request_list) = requests.get(&id) {
```

sync/src/block/downloader/body.rs
```diff
@@ -83,6 +83,12 @@ impl BodyDownloader {
         }
     }
 
+    pub fn reset_downloading(&mut self, hashes: Vec<H256>) {
```
sgkim126: No need to move `hashes`.
How about `fn reset_downloading(&mut self, hashes: &[H256])`?

jhs7jhs: I agree, and also other functions in `BodyDownloader` can be changed in that way.

jhs7jhs: I'll fix `reset_downloading` and `remove_target`.

jhs7jhs: Fixed `reset_downloading`. Others may be fixed in the next.

PR #663, opened 2018-09-05
[WIP] Leave log files when sync tests fail, and make them able to download 

jhs7jhs:


joojis:
@jhs7jhs Could you create another PR for the first commit?

jhs7jhs:
Now leaving a log when a sync test fails.

joojis:
@jhs7jhs I can't find the URLs for uploaded files. https://travis-ci.org/CodeChain-io/codechain/jobs/425193878

jhs7jhs:
@joojis It seems that it's not able to catch exceptions in `BeforeEach()`. And, I think exceptions must be in `test()`. We need to change our convention.

jhs7jhs:
It is not possible to implement now, because it's not able with Jest to delete log files when test succeeds and print the name of log files when test fails.

PR #664, opened 2018-09-05
Send a sync parcel message only when there are unsent parcels

sgkim126:
@Kais-DkM 
Is there a reason that parcel-propagation extension should send a message when there are no parcels to sync?


Kais-DkM:
I think there's no reason for that. This patch seems to give more reasonable implementation.

PR #675, opened 2018-09-07
Create log files from tests

jhs7jhs:


test/src/helper/spawn.ts
```diff
@@ -96,11 +108,15 @@ export default class CodeChain {
...
       const readline = createReadline({ input: this.process!.stderr });
+      let flag = false;
       readline.on("line", (line: string) => {
         if (line.includes("Initialization complete")) {
-          readline.close();
+          flag = true;
           resolve();
         }
+        if (this.logFlag && flag) {
+          appendFileSync(this.logPath, line + "\n");
```
joojis: Why don't you use just `appendFile` here?

jhs7jhs: I thought synchronous appending would be safe in there at that time. I'll fix it.

PR #689, opened 2018-09-07
Add tests on the memory pool

jhs7jhs:


jhs7jhs:
There's a bug in the memory pool implementation. Fixing.

jhs7jhs:
Fixed.

PR #700, opened 2018-09-11
Add tests on sending invalid parcels

jhs7jhs:


test/src/integration/basic.test.ts
```diff
@@ -63,6 +105,268 @@ describe("solo - 1 node", () => {
...
+
+        test.each(["0x01" + "0".repeat(64), "0x" + "f".repeat(128)])(
+            "nonce: %p",
+            async (nonce, done) => {
+                parcelEncoded[0] = nonce;
+                try {
+                    await node.sendSignedParcelWithRlpBytes(
+                        RLP.encode(parcelEncoded)
+                    );
+                    done.fail();
```
joojis: It's a common practice to return here.

return done.fail()

test/src/integration/basic.test.ts
```diff
@@ -63,6 +105,268 @@ describe("solo - 1 node", () => {
...
+                } catch (e) {
+                    expect(e).toEqual(ERROR.INVALID_RLP_INVALID_LENGTH);
+                }
+                done();
+            }
+        );
+
```
joojis: Could you add an INVALID_NETWORK_ID case? Any 2 characters string except for "tc" will do.

test/src/integration/basic.test.ts
jhs7jhs: Moved `done()` into the `catch`.

test/src/integration/basic.test.ts
jhs7jhs: Added.

jhs7jhs:
I'm still not sure that it is appropriate to add these test cases into `basic.test.ts`.

joojis:
@jhs7jhs How about `verification`?

jhs7jhs:
@joojis It sounds good. I'll move it.

Issue #706, opened 2018-09-12
Chain RPCs on regular key returns non-null value in the block 0


jhs7jhs:
For some `address` and public `key` that `address` has no regular key, `chain_getRegularKey(address, 0)` and `chain_getRegularKeyOwner(key, 0)` return null.
But after setting the regular key of `address` to `key`, `chain_getRegularKey(address, 0)` returns unknown value, and `chain_getRegularKeyOwner(key, 0)` returns `address`.

jhs7jhs:
See the test code in #707: chain.test.ts: 141-187.

jhs7jhs:
It's a bug in the SDK.

PR #707, opened 2018-09-12
Add tests on the chain RPCs using a block number

jhs7jhs:
It cannot be merged because of #706 now.


jhs7jhs:
@joojis Thanks for fixing the bugs. Finished now.

PR #712, opened 2018-09-12
Fix the order of locking to avoid deadlock

jhs7jhs:


sync/src/block/extension.rs
```diff
@@ -276,7 +276,7 @@ impl NetworkExtension for Extension {
                         }
                         _ => {}
                     }
-                    self.dismiss_request(&id, request_id);
+                    request_list.retain(|(i, _)| *i != request_id);
```
Kais-DkM: This one is not a good solution... but let's leave it for now.
We should change this to call function later.

jhs7jhs: Yes, it should be fixed. It's not a good idea...

PR #713, opened 2018-09-13
Do refactoring and simplify the sync code

jhs7jhs:
Removed redundant code, and made the flow clear from the hotfix #712

### DIFF
--- a/sync/src/block/extension.rs
+++ b/sync/src/block/extension.rs
@@ -267,28 +267,35 @@ impl NetworkExtension for Extension {
                 }
             }
             SYNC_EXPIRE_TOKEN_BEGIN...SYNC_EXPIRE_TOKEN_END => {
-                let mut requests = self.requests.write();
-                let mut tokens_info = self.tokens_info.write();
-                let token_info = tokens_info.get_mut(&token).unwrap();
-                if token_info.request_id.is_none() {
-                    return
-                }
+                let (id, request_id) = {
+                    let mut tokens_info = self.tokens_info.write();
+                    let token_info = match tokens_info.get_mut(&token) {
+                        Some(info) => info,
+                        None => return,
+                    };
 
-                let id = token_info.node_id;
-                let request_id = token_info.request_id.unwrap();
-                let request_list = requests.get_mut(&id).unwrap();
-
-                let expired_request = request_list.iter().find(|(r, _)| *r == request_id).cloned();
-                if let Some((request_id, request)) = expired_request {
-                    match request {
-                        RequestMessage::Bodies(hashes) => {
-                            self.body_downloader.lock().reset_downloading(&hashes);
+                    match token_info.request_id {
+                        Some(request_id) => {
+                            token_info.request_id = None;
+                            (token_info.node_id, request_id)
                         }
-                        _ => {}
+                        None => return,
                     }
-                    request_list.retain(|(i, _)| *i != request_id);
+                };
+
+                if let Some(requests) = self.requests.write().get_mut(&id) {
+                    let expired_request = requests.iter().find(|(r, _)| *r == request_id);
+                    if let Some((_, request)) = expired_request {
+                        match request {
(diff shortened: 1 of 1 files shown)


sync/src/block/extension.rs
```diff
@@ -46,12 +46,18 @@ const SYNC_EXPIRE_REQUEST_INTERVAL: i64 = 15000;
 
 const SNAPSHOT_PERIOD: u64 = (1 << 14);
 
+#[derive(Debug, PartialEq)]
+pub struct TokenInfo {
+    id: NodeId,
```
sgkim126: How about `node_Id`?

sync/src/block/extension.rs
```diff
@@ -267,28 +267,36 @@ impl NetworkExtension for Extension {
...
+                    (id, request_id)
+                };
 
-                let expired_request = request_list.iter().find(|(r, _)| *r == request_id).cloned();
-                if let Some((request_id, request)) = expired_request {
-                    match request {
-                        RequestMessage::Bodies(hashes) => {
-                            self.body_downloader.lock().reset_downloading(&hashes);
+                if let Some(requests) = self.requests.write().get_mut(&id) {
+                    let expired_request = requests.iter().find(|(r, _)| *r == request_id).cloned();
```
sgkim126: It's not about this patch, but there is no reason to call `cloned()`.

sync/src/block/extension.rs
jhs7jhs: Good idea.

sync/src/block/extension.rs
```diff
@@ -267,28 +267,36 @@ impl NetworkExtension for Extension {
                 }
             }
             SYNC_EXPIRE_TOKEN_BEGIN...SYNC_EXPIRE_TOKEN_END => {
-                let mut requests = self.requests.write();
-                let mut tokens_info = self.tokens_info.write();
-                let token_info = tokens_info.get_mut(&token).unwrap();
-                if token_info.request_id.is_none() {
-                    return
-                }
+                let (id, request_id) = {
```
sgkim126: It seems that it's not just a refactoring, but changing the behavior.
Is it safe to not lock `requests` while checking `tokens_info`?

jhs7jhs: It is not refactoring, it's my mistake. And, yes, it is safe.

sync/src/block/extension.rs
jhs7jhs: Fixed

sync/src/block/extension.rs
jhs7jhs: There was no reason to lock `tokens_info` after locking `requests`, and the key point of this patch is to make `token_info.request_id.is_none()` as a flag between `on_timeout()` and `on_peer_response()`.

sgkim126: Okay.
@Kais-DkM how about you?

sync/src/block/extension.rs
```diff
@@ -257,28 +267,36 @@ impl NetworkExtension for Extension {
...
+                    let mut tokens_info = self.tokens_info.write();
+                    let token_info = tokens_info.get_mut(&token);
+                    if token_info.is_none() {
+                        return
+                    }
 
-                let id = token_info.0;
-                let request_id = token_info.1.unwrap();
-                let request_list = requests.get_mut(&id).unwrap();
+                    let token_info = token_info.unwrap();
```
Kais-DkM: How about this one?
```
let token_info = match tokens_info.get_mut(&token) {
    Some(info) => info,
    None => return,
};
```

sync/src/block/extension.rs
Kais-DkM: It seems to be safe at a first glance. I'll accept it.

sync/src/block/extension.rs
jhs7jhs: Much better! Leaning a lot +1

Issue #662, opened 2018-09-04
Fix the integration test that synchronizing blocks between 2 nodes


joojis:
Observations:

```
  ● 2 nodes › A-B diverged › nodeA becomes ahead › It should be synced when A-B connected

    Failed: Object {
      "code": -32031,
      "data": "ParcelAlreadyImported",
      "message": "Already Imported",
    }

      84 |
      85 |     describe("nodeA becomes ahead", () => {
    > 86 |       beforeEach(async () => {
         |       ^
      87 |         await nodeA.sendSignedParcel();
      88 |         expect(await nodeA.getBestBlockNumber()).toEqual(await nodeB.getBestBlockNumber() + 1);
      89 |       });

      at Env.beforeEach (node_modules/jest-jasmine2/build/jasmine_async.js:55:24)
      at Suite.describe (src/integration/sync.test.ts:86:7)
      at Suite.describe (src/integration/sync.test.ts:85:5)
```

```
  ● 2 nodes › A-B diverged › nodeA becomes ahead › It should be synced when A-B connected

    CodeChain exited with code 1

      84 |         })
      85 |         .on("close", (code, _signal) => {
    > 86 |           reject(Error(`CodeChain exited with code ${code}`));
         |                  ^
      87 |         });
      88 |
      89 |       const readline = createReadline({ input: this.process!.stderr });

      at ChildProcess.process.on.on (src/helper/spawn.ts:86:18)
```

joojis:
I found that `CodeChain exited with code 1` error can be reproduced(at least in my environment) by the following test:

```
import CodeChain from "../helper/spawn";

let repeat = 50;
while (repeat--) {
test("Start two nodes simultaneously", done => {
    const nodes = [new CodeChain(), new CodeChain()];
    Promise.all(nodes.map(node => node.start()))
        .then(() => Promise.all(nodes.map(node => node.clean())))
        .then(() => done())
        .catch(done.fail);
});
}
```

joojis:
Patch: https://github.com/CodeChain-io/codechain/commit/c7c354b49129ed32291efc59459a7c34800de069

joojis:
The `Already Imported` error disappeared but the other error still exists:

```
    FetchError {
      message: 'request to http://localhost:8125/ failed, reason: connect ECONNREFUSED 127.0.0.1:8125',
      type: 'system',
      errno: 'ECONNREFUSED',
      code: 'ECONNREFUSED' }
```

[...]

PR #715, opened 2018-09-13
Lengthen the timeout to pass the sync tests

jhs7jhs:
See this: https://github.com/CodeChain-io/codechain/issues/662#issuecomment-420877708


jhs7jhs:
Travis does not support the test over 10 min.

PR #727, opened 2018-09-17
Fix some sync problems

sgkim126:


sync/src/block/extension.rs
```diff
@@ -559,9 +559,13 @@ impl Extension {
 
         let mut exists = Vec::new();
         for header in completed {
-            // FIXME: handle import errors
             match self.client.import_header(header.clone().into_inner()) {
                 Err(BlockImportError::Import(ImportError::AlreadyInChain)) => exists.push(header.hash()),
+                // FIXME: handle import errors
```
kseo: What's the plan here?

Kais-DkM: My original plan here was to 1. stop downloading children of this block, and 2. punish the peer that sent this block.
Both of them are not a high priority issue, but they need to be done later.

PR #729, opened 2018-09-18
Make Solo able to set the author

joojis:


core/src/miner/miner.rs
```diff
@@ -537,7 +537,7 @@ impl MinerService for Miner {
     fn set_author(&self, address: Address, password: Option<Password>) -> Result<(), SignError> {
         *self.author.write() = address;
 
-        if self.engine.seals_internally().is_some() {
+        if self.engine.seals_internally().is_some() && self.engine_type() != EngineType::Solo {
```
kseo: Let's change the condition to

`self.engine_type() == EngineType::InternalSealing && self.engine.seals_internally().is_some()`

joojis: Done.

Issue #724, opened 2018-09-14
Handle the situation that two nodes try to connect to each other simultaneously


jhs7jhs:
It happens in the discovery integration tests. You can reproduce the situation sometimes by running the tests.

If two nodes try to connect to each other at the same time, the routing table of each node set the other node to the `establish` state, by
```
if self.connections.connect(token, stream, local_node_id, session, socket_address, self.get_port()) {
    self.routing_table.establish(socket_address);
    Some(token)
}
```
Next, they sends `Sync` message to each other, and stuck. Also, you can find out an error message `Cannot find session`, because of the code below:
```
let session = self
    .routing_table
    .unestablished_session(&remote_addr)
    .ok_or(Error::General("Cannot find session"))?;
```
Notice that it is already established in the routing table.

jhs7jhs:
Fixed in #732

PR #732, opened 2018-09-18
Handle multiple connections between two nodes

jhs7jhs:
Fixing #724.

Two problems:
1. A timer is not working even successfully set by `io.register_timer_once(stream, random_interval).expect("Pull connections must be registered");`
2. Some processes do not die normally, but `main TRACE shutdown  [IoService] Closed.` in every log.

### DIFF
--- a/network/src/p2p/handler.rs
+++ b/network/src/p2p/handler.rs
@@ -97,6 +97,7 @@ pub struct Handler {
 
     listener: Listener,
 
+    establish_lock: Mutex<()>,
     tokens: Mutex<TokenGenerator>,
 
     routing_table: Arc<RoutingTable>,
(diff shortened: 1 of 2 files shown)


jhs7jhs:
Fixed. But I have no idea that why `io.register_timer_once()` is working well in `finally()`.

jhs7jhs:
Two states are added in the routing table: `Establishing` and `WaitingReconnection`.
If a node get a sync message from the opposite node while `Establishing`, it means there are multiple connections so the node will disconnect, change the state to `WaitingReconnection`, and wait reconnection for random duration between 2s~5s. (See `MINIMUM_RECONNECTION_TIME` and `MAXIMUM_RECONNECTION_TIME`)

jhs7jhs:
Found an exception.

jhs7jhs:
The new issues are added (#744, #745) and the discovery test still has problems, so not allowing the test in this patch. But the patch will resolve the issue #742.
@sgkim126

PR #742, opened 2018-09-20
Add the spec of CHKMULTISIG opcode

joojis:


spec/CodeChain-Virtual-Machine.md
```diff
@@ -80,6 +80,13 @@ Leading zeros must be truncated. Note that it is allowed to decode value with le
  1. Pop two values, first one as signature, second one as public key
  1. Verify signature via full transaction message, excluding script parameter
  1. Push true on success, false otherwise.
+* CHKMULTISIG(0x81)
+ 1. Pop one value, the value is the n in the m-of-n Multisig. The maximum value of n is 5.
```
Kais-DkM: Do we need a limit on `n`?

spec/CodeChain-Virtual-Machine.md
```diff
@@ -80,6 +80,13 @@ Leading zeros must be truncated. Note that it is allowed to decode value with le
...
+ 1. Pop n values, which are distinct public keys.
+ 1. Pop one value, the value is the m in the m-of-n Multisig.
+ 1. Pop m values, which are distinct signatures. The signature scheme use the same way as CHKSIG.
+ 1. Verify signatures. The signatures must be ordered the same way as public keys.
```
Kais-DkM: Are these signatures expected to be encoded in our signature format?

Kais-DkM:
I'm not sure if we really need this opcode.
We'd need it if we're going to stick with ECDSA, but AFAIK, our plan is to move to schnorr signatures.
With Schnorr signatures, this opcode would be unnecessary since it can be embedded in `CHKSIG`.

joojis:
@Kais-DkM It's not settled when we're going to use Schnorr and remove ECDSA.
Multisig is a necessary feature for many use cases I think so we can't put off implementing this indefinitely.

And I'm not sure, that we use Schnorr means that we don't need ECDSA Multisig in VM.

spec/CodeChain-Virtual-Machine.md
Kais-DkM: Transactions containing `CHKSIG` with large `m` will be rejected by size limit on parcel.
I think we can leave flexibility here by removing this limit.

Kais-DkM:
I can agree that we need to implement multisig feature as soon as possible.
What I'm worried about is that it will be practically deprecated when shnorr signature comes in.
I think everyone would use embedded multisig (like [MuSig](https://eprint.iacr.org/2018/068)) since it results in much shorter signature, which means much less transaction fee.

I wouldn't block this from being merged, but let's keep in mind that deprecation in blockchain is quite a tough procedure.

spec/CodeChain-Virtual-Machine.md
joojis: Isn't the limit 4MB? It's big enough to make something kind of 1000 of 1000 Multisig script.
AFAIK, ECDSA verification takes hundreds of milliseconds.

joojis: I found the limits of stack size and script size in the spec. I think I can remove this limit then.

GNUp:
It looks good to me.

Issue #668, opened 2018-09-06
Support tagging in a whitelist/blacklist


jhs7jhs:
It is hard to distinguish each IPs in a whitelist/blacklist, so it should be good if tagging is supported in the whitelist/blacklist format.

kseo:
@jhs7jhs Would you add an example of tagging?

jhs7jhs:
Currently, there's no fixed format for it, but I think like this: (address, comment or tag)
```
127.0.0.1, localhost
A.B.C.D, node1
A.B.C.E, node2
...
```

jhs7jhs:
@joojis This will change the JSON RPC format of `net_addToWhitelist`, `net_removeFromWhitelist`, `net_addToBlacklist`, `net_removeFromBlacklist`, `net_getWhitelist, and `net_getBlacklist`.

PR #748, opened 2018-09-21
Support tagging in a whitelist/blacklist

jhs7jhs:
This implements #668.
Only one IP address is allowed in one line, and you can comment/tag by `'#'`.
For example,
Whitelist:
```
127.0.0.1 # Wow
128.0.0.1 # Nice
# This is a pure comment
130.0.0.1
# IP address with no tag
129.0.0.1 #Yes
```
The result of `net_getWhitelist`:
```
{"jsonrpc":"2.0",
 "result":{
    "enabled":true,
    "list":[["127.0.0.1","Wow"],["128.0.0.1","Nice"],["129.0.0.1","Yes"],["130.0.0.1",null]]
}}
```

### DIFF
--- a/codechain/config/mod.rs
+++ b/codechain/config/mod.rs
@@ -17,14 +17,13 @@
 mod chain_type;
 
 use std::fs;
-use std::net::IpAddr;
 use std::str::{self, FromStr};
 use std::time::Duration;
 
 use ccore::{MinerOptions, ShardValidatorConfig, StratumConfig};
 use ckey::PlatformAddress;
 use clap;
-use cnetwork::{NetworkConfig, SocketAddr};
+use cnetwork::{FilterEntry, NetworkConfig, SocketAddr};
 use rpc::{RpcHttpConfig, RpcIpcConfig, RpcWsConfig};
 use toml;
 
@@ -124,7 +123,7 @@ impl Config {
     pub fn network_config(&self) -> Result<NetworkConfig, String> {
         debug_assert!(!self.network.disable.unwrap());
 
-        fn make_ipaddr_list(list_path: Option<&String>, list_name: &str) -> Result<Vec<IpAddr>, String> {
+        fn make_ipaddr_list(list_path: Option<&String>, list_name: &str) -> Result<Vec<FilterEntry>, String> {
             if let Some(path) = list_path {
                 fs::read_to_string(path)
                     .map_err(|e| format!("Cannot open the {}list file {:?}: {:?}", list_name, path, e))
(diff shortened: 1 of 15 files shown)


jhs7jhs:
@sgkim126 
I think a struct for a tagged IP address is needed, but have no idea where to define it. Can I just define it in `filter.rs` and use it in other places, including `control.rs` and `net.rs` of RPC?

sgkim126:
How about add new file under filters directory?

jhs7jhs:
I think it is okay, but the file will only includes a struct like:
```
pub struct FilterEntry {
    addr: IpAddr,
    tag: Option<String>,
}
```

codechain/config/mod.rs
```diff
@@ -112,24 +112,33 @@ impl Config {
     pub fn network_config(&self) -> Result<NetworkConfig, String> {
         debug_assert!(!self.network.disable.unwrap());
 
-        fn make_ipaddr_list(list_path: Option<&String>, list_name: &str) -> Result<Vec<IpAddr>, String> {
+        fn make_ipaddr_list(
+            list_path: Option<&String>,
+            list_name: &str,
+        ) -> Result<Vec<(IpAddr, Option<String>)>, String> {
```
sgkim126: Why do you distinguish the case that there is a comment or not?

jhs7jhs: If there's no tag on an IP address, how would you store it?

sgkim126: An empty string.
IMO, there is no difference between ```8.8.8.8``` and ```8.8.8.8 # ```.

jhs7jhs: I thought the empty string might not be a good idea at that time. I'll think about it after fixing some issues on sync.

jhs7jhs: Changed from `Option<String>` to `String`

jhs7jhs:
@sgkim126 Add `FilterEntry` in `filters/filter.rs`. Is it better to make a new file for the struct?

network/src/filters/filter.rs
```diff
@@ -32,15 +38,22 @@ impl Default for Filter {
...
             enabled: !input_vector.is_empty(),
-            list: input_vector.into_iter().collect(),
+            list: input_vector.into_iter().map(|x| (x.addr, x.tag)).collect(),
         }
     }
 
-    pub fn add(&mut self, addr: IpAddr) {
-        self.list.insert(addr);
+    pub fn add(&mut self, addr: IpAddr, tag: Option<String>) {
+        match tag {
```
sgkim126: How about
```rust
self.list.insert(addr, tag.unwrap_or_default());
```

sgkim126:
Please merge a7ffe13c58fb4f92f6e2102988be6de65741e8c6 onto 95c33d56ebe491fe6b416bcc4cea4c10c14b038c.

network/src/filters/filter.rs
jhs7jhs: I intended to not to change the tag value if the input is `None`.

jhs7jhs:
Merged now.

PR #758, opened 2018-09-27
Add the spec for sendParcel RPC

sgkim126:


joojis:
Shouldn't we mention `unlock` in the description?

sgkim126:
@joojis I referenced [account_sign](https://github.com/CodeChain-io/codechain/blob/master/spec/JSON-RPC.md#account_sign). It also uses the unlocked account if the password is not given. However, it doesn't say about `unlock`.

PR #769, opened 2018-09-28
Add test cases for on-chain synchronization

GNUp:


test/src/integration/onChain.test.ts
```diff
@@ -0,0 +1,215 @@
...
+        nodeA = new CodeChain({ logFlag: true });
+        await nodeA.start();
+    });
+
+    afterEach(async () => {
+        await nodeA.clean();
+    });
+
+    test("OnChain parcel propagation test", async () => {
+        const TH = new TestHelper("0.0.0.0", nodeA.port);
```
joojis: Could you explain the first argument? Is it an address for listening?

test/src/integration/onChain.test.ts
```diff
@@ -0,0 +1,215 @@
...
+            amount: 10000
+        });
+        const signedparcel = parcel.sign({
+            secret: ACCOUNT_SECRET,
+            fee: 10,
+            nonce: 0
+        });
+
+        await TH.sendEncodedParcel([signedparcel.toEncodeObject()]);
+        const block = await sdk.rpc.chain.getBlock(1);
```
joojis: I think we can make this case smaller by making the node stop sealing and then we query pending parcels.

test/src/integration/onChain.test.ts
```diff
@@ -0,0 +1,215 @@
...
+        await TH.end();
+    });
+
+    test("OnChain block propagation test", async () => {
+        const TH = new TestHelper("0.0.0.0", nodeA.port);
+        await TH.establish();
+
+        const sdk = nodeA.sdk;
+
+        // Genesis block
```
joojis: Could you replace it with a function?

We don't need to read the whole genesis block data at this point.

joojis: Also for the block1 and block2.

test/src/integration/onChain.test.ts
GNUp: The IP address and port are for the target node.
A listening IP and port is 0.0.0.0 and 6602.

joojis:
Please update the SDK version of in test-helper dependencies.

image

Release candidate versions can be unpublished accidentally. Sorry for the inconvenience.

test/src/integration/onChain.test.ts
GNUp: I changed it

test/src/integration/onChain.test.ts
GNUp: I made it as function

joojis:
I think you forgot to add `yarn.lock` after updating package.json. Travis is still failing.

q
```diff
@@ -0,0 +1,52 @@
+[1mdiff --git a/test/src/integration/onChainBlock.test.ts b/test/src/integration/onChainBlock.test.ts[m
```
joojis: What is this file?

test/src/integration/onChainBlock.test.ts
```diff
@@ -0,0 +1,407 @@
...
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
```
joojis: I think it's common to put a blank line between the copyright sentences and the imports.

GNUp: I inserted LF between them.

test/src/integration/onChainBlock.test.ts
```diff
@@ -0,0 +1,408 @@
...
+            expect(block1).toEqual(expect.anything());
+            expect(block2).toEqual(expect.anything());
+
+            await TH.end();
+        },
+        30000
+    );
+
+    test(
+        "OnChain invalid timestamp block propagation test",
```
joojis: I think it should tell the difference between the previous one. They have the same description.

PR #770, opened 2018-09-28
Make register_handler synchronous

### LINKED ISSUE #753: Timer setting has a race condition
It can happens in everywhere implementing `NetworkExtension` and `IoHandler`.
It can be checked by running the sync test several times. While watching logs, it seems that `mio` is working but `on_timeout()` is not called.

sgkim126:
It fixes #753.


joojis:
Wow! Some of the sync tests was constantly failing in my environment.
But this patch makes all of the sync tests green. Great. :)

kseo:
Good job!

PR #772, opened 2018-09-28
Change random account test to use PRNG

jhs7jhs:
Using `seedrandom` library for a PRNG.


jhs7jhs:
@sgkim126 It's my mistake. `getRandomIndex()` is only used in the random account test. I'll fix it

PR #790, opened 2018-10-05
Adjust test net configuration

sgkim126:
Please merge it before you reset the test nets.


sgkim126:
@kseo I adjusted the block reward also.

PR #792, opened 2018-10-05
Fix a bug in RLP decoding while TCP stream.

GNUp:
0xf7 prefix means 55 length of list. From the value of 0xf8, the
byte length of length is added behind the prefix.


network/src/p2p/stream.rs
```diff
@@ -81,7 +81,7 @@ impl TryStream {
     fn read_len_of_len(&mut self, mut bytes: Vec<u8>) -> io::Result<(usize, Vec<u8>)> {
         debug_assert_eq!(None, self.read);
         debug_assert_eq!(1, bytes.len());
-        debug_assert!(bytes[0] >= 0xf7);
```
sgkim126: Would you make it `bytes[0] > 0xf7`?

GNUp: I changed it.

PR #796, opened 2018-10-08
Add invalid time stamp block propagation test

GNUp:
There are some issues in sending invalid timestamp onChain blocks.
---
Below is a test scenario.
1. Make block1 and block2.
2. The timestamp in Block2's header should be invalid. Simply make it smaller than block1.
2. Send a `status message` with block2's hash as the bestHash.
3. Send a `header response message` which is [genesisHeader, block1Header, block2Header].
4. Send a `body response message` == [[], []], which is correspond to block1's body and block2's body.
5. Block is accepted from the target node only for block1. So the target node should send a status message with block1's hash as bestHash.
---
And here are problems
1. The target node does not send `status message` if in the above situation. However, if I inserted `info!("55");` then the `status message` come properly.
2. In below `verify()` function, `Ok(verified)` is returned twice although the block2 is invalid.
3. `Header request` does not come from the target node anymore, even if after sending valid block1 and block2. -- Resolved, I misunderstood sync behaviour
```rust

    fn verify(
        verification: Arc<Verification<K>>,
        engine: Arc<CodeChainEngine>,
        ready_signal: Arc<QueueSignal>,
        empty: Arc<SCondvar>,
        more_to_verify: Arc<SCondvar>,
        _id: usize,
    ) {
        loop {
            // wait for work if empty.
            {
                let mut more_to_verify_mutex = verification.more_to_verify_mutex.lock().unwrap();

                if verification.unverified.lock().is_empty() && verification.verifying.lock().is_empty() {
                    empty.notify_all();
                }

                while verification.unverified.lock().is_empty() {
                    more_to_verify_mutex = more_to_verify.wait(more_to_verify_mutex).unwrap();
                }
            }

            // do work.
            let item = {
                // acquire these locks before getting the item to verify.
                let mut unverified = verification.unverified.lock();
                let mut verifying = verification.verifying.lock();

                let item = match unverified.pop_front() {
                    Some(item) => item,
                    None => continue,
                };

                verification.sizes.unverified.fetch_sub(item.heap_size_of_children(), AtomicOrdering::SeqCst);
                verifying.push_back(Verifying {
                    hash: item.hash(),
                    output: None,
                });
                item
            };

            let hash = item.hash();
            let is_ready = match K::verify(item, &*engine, verification.check_seal) {
                Ok(verified) => {
                    let mut verifying = verification.verifying.lock();
                    let mut idx = None;
                    for (i, e) in verifying.iter_mut().enumerate() {
                        if e.hash == hash {
                            idx = Some(i);

                            verification
                                .sizes
                                .verifying
                                .fetch_add(verified.heap_size_of_children(), AtomicOrdering::SeqCst);
                            e.output = Some(verified);
                            break
                        }
                    }

                    if idx == Some(0) {
                        // we're next!
                        let mut verified = verification.verified.lock();
                        let mut bad = verification.bad.lock();
                        VerificationQueue::drain_verifying(
                            &mut verifying,
                            &mut verified,
                            &mut bad,
                            &verification.sizes,
                        );
                        info!("111");
                        true
                    } else {
                        info!("222");
                        false
                    }
                }
                Err(_) => {
                    let mut verifying = verification.verifying.lock();
                    let mut verified = verification.verified.lock();
                    let mut bad = verification.bad.lock();

                    bad.insert(hash.clone());
                    verifying.retain(|e| e.hash != hash);

                    if verifying.front().map_or(false, |x| x.output.is_some()) {
                        VerificationQueue::drain_verifying(
                            &mut verifying,
                            &mut verified,
                            &mut bad,
                            &verification.sizes,
                        );
                        info!("33");
                        true
                    } else {
                        info!("44");
                        false
                    }
                }
            };
            if is_ready {
                info!("55");
                // Import the block immediately
                ready_signal.set_sync();
            }
        }
    }
```

### DIFF
--- a/test/package.json
+++ b/test/package.json
@@ -30,6 +30,7 @@
     "seedrandom": "^2.4.4",
     "ts-jest": "^22.4.6",
     "ts-node": "^7.0.0",
-    "typescript": "^2.9.2"
+    "typescript": "^2.9.2",
+    "codechain-test-helper": "https://github.com/CodeChain-io/codechain-test-helper-js.git#a4a9c26635a52395dfd17733eef66b18a144fd00"
   }
 }
--- a/test/src/integration/onChainBlock.test.ts
+++ b/test/src/integration/onChainBlock.test.ts
@@ -1,0 +1,140 @@
+// Copyright 2018 Kodebox, Inc.
+// This file is part of CodeChain.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import { TestHelper } from "codechain-test-helper/lib/testHelper";
+import { Header } from "codechain-test-helper/lib/cHeader";
+import CodeChain from "../helper/spawn";
+import { H256, U256, H160 } from "codechain-primitives/lib";
+
+describe("Test onChain block communication", async () => {
+    let nodeA: CodeChain;
+    let TH: TestHelper;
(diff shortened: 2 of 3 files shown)


Kais-DkM:
If this test doesn't pass, it means that we have some problem with verfication process related to timestamps.
Could you open an issue for this too?

GNUp:
Let's discuss this at https://github.com/CodeChain-io/codechain/issues/808#issue-369378004

PR #801, opened 2018-10-10
Multi sig

GNUp:


types/src/util/tag.rs
```diff
@@ -0,0 +1,68 @@
...
+pub struct Tag {
+    pub input_scheme: bool,
+    pub output_scheme: bool,
+    pub filter_len: usize,
+    pub filter: Vec<u8>,
+}
+
+impl Tag {
+    pub fn new(mut bitvec: Vec<u8>) -> Result<Tag, HashingError> {
+        let tag = if let Some(tag) = bitvec.pop() {
```
sgkim126:
```rust
let tag = bitvec.pop().ok_or(HashingError::InvalidFilter)?;
```

types/src/util/tag.rs
```diff
@@ -0,0 +1,68 @@
...
+
+pub struct Tag {
+    pub input_scheme: bool,
+    pub output_scheme: bool,
+    pub filter_len: usize,
+    pub filter: Vec<u8>,
+}
+
+impl Tag {
+    pub fn new(mut bitvec: Vec<u8>) -> Result<Tag, HashingError> {
```
sgkim126:
```
pub fn try_new(mut bitvec: Vec<u8>) -> Result<Tag, HashingError> {
```

types/src/util/tag.rs
```diff
@@ -0,0 +1,68 @@
...
+        let tag = if let Some(tag) = bitvec.pop() {
+            tag
+        } else {
+            return Err(HashingError::InvalidFilter)
+        };
+        let input_scheme = (tag & 0x1) == 1;
+        let output_scheme = (tag >> 1 & 0x1) == 1;
+        let filter_len = ((tag >> 2) & 0x3f) as usize;
```
sgkim126: Does `& 0x3f` have a meaning?

types/src/util/tag.rs
```diff
@@ -0,0 +1,68 @@
...
+            output_scheme,
+            filter_len,
+            filter: bitvec,
+        })
+    }
+
+    pub fn get_tag(self) -> Vec<u8> {
+        let filter_len = (self.filter_len << 2) as u8;
+        let output_scheme: u8 = if self.output_scheme {
+            2
```
sgkim126:
```rust
0b10
```

types/src/util/tag.rs
```diff
@@ -0,0 +1,68 @@
...
+
+        Ok(Tag {
+            input_scheme,
+            output_scheme,
+            filter_len,
+            filter: bitvec,
+        })
+    }
+
+    pub fn get_tag(self) -> Vec<u8> {
```
sgkim126: How about making `bitvec` as a member and return the reference to that?

sgkim126: Is it `get_tag` or `into_tag`?
If it's a `get_tag`, it must be `fn get_tag(&self)`.

vm/src/executor.rs
```diff
@@ -229,15 +231,46 @@ pub fn execute(
...
                     Ok(true) => 1,
                     _ => 0,
                 };
                 stack.push(Item(vec![result]))?;
             }
+            Instruction::ChkMultiSig => {
+                // Get n pubkey. If there are more than six pubkeys, return error.
+                let n = stack.pop()?.assert_len(1)?.as_ref()[0] as usize;
+                let mut pubkey: Vec<Public> = Vec::new();
+                for _i in 0..n {
```
sgkim126:
```rust
for _ in 0..n {
```

vm/src/executor.rs
```diff
@@ -229,15 +231,46 @@ pub fn execute(
...
+                    pubkey.push(Public::from_slice(stack.pop()?.assert_len(64)?.as_ref()));
+                }
+
+                // Get m signature. If signatures are more than pubkeys, return error.
+                let m = stack.pop()?.assert_len(1)?.as_ref()[0] as usize;
+                if m > n || m > 6 {
+                    return Err(RuntimeError::InvalidArg)
+                }
+                let mut signatures: Vec<Signature> = Vec::new();
+                for _i in 0..m {
```
sgkim126:
```rust
for _i in 0..m {
```

types/src/util/tag.rs
GNUp: I changed it.

types/src/util/tag.rs
GNUp: I removed it.

types/src/util/tag.rs
GNUp: I changed it.

types/src/util/tag.rs
GNUp: I changed it.

GNUp: I think it is more close to `get_tag`. However, I don't know what is exactly different between them.

vm/src/executor.rs
GNUp: I changed it.

vm/src/executor.rs
GNUp: I changed it to ```rust
for _ in 0..m {
```

File: types/src/util/tag.rs (hunk 5 above)
Comment (sgkim126) [reply types/src/util/tag.rs]: `get_tag(&self)` doesn't take the ownership, but `get_tag(self)` does.
For instance, if `get_tag` takes the ownership and you call `a.get_tag()`, you cannot use `a` after that.
Naming it as `into_*` or implementing `From` is a general way of naming a function that takes the ownership.

File: types/src/util/tag.rs [hunk 8]
```diff
@@ -0,0 +1,51 @@
...
+        let output_scheme = (tag >> 1 & 0x1) == 1;
+        let filter_len = (tag >> 2) as usize;
+
+        if bitvec.len() != filter_len {
+            return Err(HashingError::InvalidFilter)
+        }
+
+        Ok(Tag {
+            sign_all_inputs,
+            output_scheme0b10
```
sgkim126: ?

GNUp: I fixed it.

vm/src/executor.rs
```diff
@@ -229,15 +231,46 @@ pub fn execute(
...
+                let mut signatures: Vec<Signature> = Vec::new();
+                for _ in 0..m {
+                    signatures.push(Signature::from(stack.pop()?.assert_len(SIGNATURE_LENGTH)?.as_ref()));
+                }
+
+                let tag = Tag::try_new(stack.pop()?.as_ref().to_vec())?;
+                let tx_hash = tx.hash_partially(tag, cur, burn)?;
+
+                let mut result = 1;
+                while let Some(sig) = signatures.pop() {
```
joojis: Is it safe when the vector is empty?

How about disallowing `m=0` case?

vm/src/executor.rs
```diff
@@ -229,15 +231,46 @@ pub fn execute(
...
+                // Get n pubkey. If there are more than six pubkeys, return error.
+                let n = stack.pop()?.assert_len(1)?.as_ref()[0] as usize;
+                let mut pubkey: Vec<Public> = Vec::new();
+                for _ in 0..n {
+                    pubkey.push(Public::from_slice(stack.pop()?.assert_len(64)?.as_ref()));
+                }
+
+                // Get m signature. If signatures are more than pubkeys, return error.
+                let m = stack.pop()?.assert_len(1)?.as_ref()[0] as usize;
+                if m > n || m > 6 {
```
joojis: Let's remove the `m > 6` condition. As discussed in https://github.com/CodeChain-io/codechain/pull/742 , the script size limit will limit the size of `m`.

vm/src/executor.rs
GNUp: Okay, I thought that in the `m=0`case, it is a right behaviour to success verifying.

vm/src/executor.rs
GNUp: I removed it.

GNUp: I think it is needed to inform that lock script is invalid. It is better to rename `RuntimeError::invalidArg` to `RuntimeError::invalidLockscript`

PR #814, opened 2018-10-17
Asset compose and decompose

GNUp:


rpc/src/v1/types/transaction.rs
```diff
@@ -69,6 +69,26 @@ pub enum Transaction {
...
+        metadata: String,
+        registrar: Option<PlatformAddress>,
+        inputs: Vec<AssetTransferInput>,
+        output: AssetMintOutput,
+    },
+    #[serde(rename_all = "camelCase")]
+    AssetDecompose {
+        network_id: NetworkId,
+        shard_id: ShardId,
+        world_id: WorldId,
```
joojis: I guess shard_id and world_id are not needed.

state/src/item/asset_scheme.rs
```diff
@@ -53,33 +56,43 @@ impl AssetScheme {
...
         assert_eq!("", &self.metadata);
         assert_eq!(0, self.amount);
         assert_eq!(None, self.registrar);
         self.metadata = metadata;
         self.amount = amount;
         self.registrar = registrar;
+        self.pool = pool;
+    }
+
+    pub fn pool(&self) -> &Vec<Asset> {
```
sgkim126:
```rust
pub fn pool(&self) -> &[Asset] {
```

types/src/transaction/transaction.rs
```diff
@@ -227,6 +275,42 @@ impl Transaction {
...
+                    return Err(Error::EmptyInput)
+                }
+                for input in inputs {
+                    if input.prev_out.amount == 0 {
+                        return Err(Error::ZeroAmount)
+                    }
+                }
+                match output.amount {
+                    Some(amount) if amount == 1 => Ok(()),
+                    _ => Err(Error::ZeroAmount),
```
joojis: It seems we need a different error message here for a non-zero amount

types/src/transaction/transaction.rs
```diff
@@ -403,6 +507,46 @@ impl PartialHashing for Transaction {
...
+                        nonce: *nonce,
+                        metadata: metadata.to_string(),
+                        registrar: *registrar,
+                        inputs: new_inputs,
+                        output: new_output,
+                    }.rlp_bytes(),
+                    &blake128(tag.get_tag()),
+                ))
+            }
             _ => unreachable!(),
```
joojis: It seems AssetDecompose case reaches here.

types/src/transaction/transaction.rs
```diff
@@ -513,6 +559,7 @@ const ASSET_MINT_ID: TransactionId = 0x03;
 const ASSET_TRANSFER_ID: TransactionId = 0x04;
 const SET_WORLD_USERS_ID: TransactionId = 0x05;
 const ASSET_COMPOSE_ID: TransactionId = 0x06;
+const ASSET_DECOMPOSE_ID: TransactionId = 0x7;
```
joojis: 0x07, for the consistency

state/src/impls/shard_level.rs
```diff
@@ -511,7 +511,7 @@ impl<B: Backend + ShardBackend> ShardLevelState<B> {
             .fold(true, |acc, ref asset| acc && (None != asset_scheme.pool().iter().position(|ref x| x == asset)))
             || pool.len() != asset_scheme.pool().len()
         {
-            return Err(TransactionError::InvalidInput.into())
```
sgkim126: Don't you think the order of commit is weird?
It must before `209c8ac0a466dbf9664793d7ff870e292e0ae25a` and that commit must use `InvalidOutput` error.

types/src/transaction/transaction.rs
```diff
@@ -461,6 +461,46 @@ impl PartialHashing for Transaction {
...
+                network_id,
+                shard_id,
+                world_id,
+                nonce,
+                metadata,
+                registrar,
+                inputs,
+                output,
+            } => {
+                if tag.filter_len != 0 {
```
sgkim126: Why must `filter_len` be zero?

state/src/impls/mod.rs
```diff
@@ -17,5 +17,8 @@
 mod shard_level;
 mod top_level;
 
+#[cfg(test)]
```
sgkim126: Why?

types/src/transaction/transaction.rs
```diff
@@ -486,6 +632,38 @@ impl Decodable for Transaction {
...
+                    output: AssetMintOutput {
+                        lock_script_hash: d.val_at(7)?,
+                        parameters: d.val_at(8)?,
+                        amount: d.val_at(9)?,
+                    },
+                    nonce: d.val_at(10)?,
+                })
+            }
+            ASSET_DECOMPOSE_ID => {
+                if d.item_count()? != 6 {
```
sgkim126: 7

sgkim126: And it seems that you need more tests for decomposing.

state/src/impls/shard_level.rs
```diff
@@ -475,6 +480,62 @@ impl<B: Backend + ShardBackend> ShardLevelState<B> {
...
+            let current_amount = sum.get(&output_type).cloned().unwrap_or(0);
+            sum.insert(output_type.clone(), current_amount + output.amount);
+        }
+        let pool: Vec<Asset> = sum.into_iter().map(|(asset_type, amount)| Asset::new(asset_type, amount)).collect();
+        if !pool
+            .iter()
+            .fold(true, |acc, ref asset| acc && (None != asset_scheme.pool().iter().position(|ref x| x == asset)))
+            || pool.len() != asset_scheme.pool().len()
+        {
+            return Err(TransactionError::InvalidInput.into())
```
joojis: I think it's better if the error includes asset type and amount. Like `InvalidAssetAmount` in TransactionError.

state/src/impls/shard_level.rs
```diff
@@ -475,6 +480,62 @@ impl<B: Backend + ShardBackend> ShardLevelState<B> {
...
+
+        let (_, asset_address) = self.check_input_asset(input, sender)?;
+        let script_result = self.check_and_run_input_script(input, transaction, &input.prev_out, false)?;
+
+        match script_result {
+            ScriptResult::Unlocked => {}
+            _ => return Err(TransactionError::FailedToUnlock(asset_address.into()).into()),
+        }
+
+        self.kill_asset(&asset_address);
```
joojis: Isn't it supposed to delete the asset scheme also?

types/src/transaction/transaction.rs
joojis: There is only 1 output for `AssetCompose`. `output = all` and `output = none` can describe all the case.

types/src/transaction/transaction.rs
GNUp: I fixed it and added an RLP test case for decompose transaction.

rpc/src/v1/types/transaction.rs
GNUp: I removed it.

state/src/item/asset_scheme.rs
GNUp: I changed it.

types/src/transaction/transaction.rs
GNUp: I replaced it with `InvalidOutput`

types/src/transaction/transaction.rs
GNUp: I implemented it.

types/src/transaction/transaction.rs
GNUp: I changed it.

state/src/item/asset_scheme.rs
GNUp: It seems `&[Asset]` and `&Vec<Asset>` is the same type. Could you explain why `&[Asset]` is better?

state/src/impls/mod.rs
GNUp: I thought if I tag `mod tests;` with `#[cfg(test)]`, the tests directory will be compiled only in test mode.

types/src/transaction/error.rs
```diff
@@ -216,6 +230,13 @@ impl Display for Error {
             ),
             Error::ZeroAmount => write!(f, "An amount cannot be 0"),
             Error::TooManyOutputs(num) => write!(f, "The number of outputs is {}. It should be 126 or less.", num),
+            Error::EmptyInput => write!(f, "AssetCompose must have at least one input."),
+            Error::InvalidInput => write!(
+                f,
+                "
```
sgkim126: ?

state/src/impls/mod.rs
sgkim126: I meant why do you split tests in the other module?
It makes the test cannot check the private status.

state/src/impls/shard_level.rs
GNUp: I reordered it.

state/src/impls/shard_level.rs
GNUp: I implemented it.

state/src/item/asset_scheme.rs
sgkim126: Slice and Vector are not the same type. They are different as String vs str.
https://www.reddit.com/r/rust/comments/8kujd1/newbie_correct_way_to_pass_vector_of_references/

state/src/impls/shard_level.rs
sgkim126: It seems that comment is missing.

state/src/impls/shard_level.rs
GNUp: I implemented it.

sgkim126:
```
error[E0599]: no variant named `InvalidInput` found for type `ctypes::transaction::Error` in the current scope
   --> state/src/impls/tests/shard_level.rs:727:37
    |
727 |         Ok(TransactionInvoice::Fail(TransactionError::InvalidInput)),
    |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ variant not found in `ctypes::transaction::Error`
```

state/src/impls/mod.rs
GNUp: I thought the source code was too long because of tests code and also the private status can be check by the test function.

sgkim126: Abstract the module if you think it's too long.
It's not the proper way in rust to extract unit tests to another file.
https://doc.rust-lang.org/book/second-edition/ch11-03-test-organization.html

state/src/impls/top_level.rs
```diff
@@ -574,7 +574,7 @@ impl TopLevelState {
         Ok((new_root, db, results))
     }
 
-    fn create_shard_level_state(&mut self, owners: Vec<Address>, users: Vec<Address>) -> StateResult<()> {
+    pub fn create_shard_level_state(&mut self, owners: Vec<Address>, users: Vec<Address>) -> StateResult<()> {
```
sgkim126: You break the abstraction just for testing.
It should be avoided if possible.

GNUp: Okay, I will put back the unit tests to the original place.

PR #815, opened 2018-10-17
Extract some addresses and secrets

sgkim126:


joojis:
@sgkim126 From the looks of it, I was just needed any valid address.

PR #818, opened 2018-10-19
Implement timelock feature

joojis:


core/src/miner/mem_pool.rs
```diff
@@ -78,6 +79,40 @@ impl ParcelOrigin {
...
+            (Some(_), _) => return Ordering::Less,
+            (_, Some(_)) => return Ordering::Greater,
+            _ => (),
+        }
+
+        match (self.timestamp, other.timestamp) {
+            (Some(self_timestamp), Some(other_timestamp)) if self_timestamp != other_timestamp => {
+                return other_timestamp.cmp(&self_timestamp)
+            }
+            (Some(_), _) => return Ordering::Less,
```
sgkim126: Is `ParcelTimeLock { timestamp: Some(timestamp), .. }` less than `ParcelTimeLock { timestamp: None, .. }`?
What's the definition of the order of the timelock?

joojis: No, it should be not. I will change it.

A lower value or `None` should be prioritized.
The basic idea is that miners prefer a shorter one so they can earn the fee earlier.

core/src/client/client.rs
```diff
@@ -423,6 +423,20 @@ impl TransactionInfo for Client {
     fn transaction_parcel(&self, id: TransactionId) -> Option<ParcelAddress> {
         self.transaction_address(id).map(|addr| addr.parcel_address)
     }
+
+    fn transaction_block_number(&self, id: TransactionId) -> Option<BlockNumber> {
```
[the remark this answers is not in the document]
joojis: It's basically no harm because the UTXO fetch will fail, but I will check which error message is shown.

core/src/miner/mem_pool.rs
```diff
@@ -819,8 +925,26 @@ impl MemPool {
...
+        best_block_number: BlockNumber,
+        best_block_timestamp: u64,
+    ) -> bool {
+        if let Some(block_number) = timelock.block {
+            if block_number > best_block_number + 1 {
+                return true
+            }
+        }
+        if let Some(timestamp) = timelock.timestamp {
+            if timestamp > cmp::max(get_time().sec as u64, best_block_timestamp + 1) {
```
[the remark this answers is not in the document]
joojis: The next block's timestamp is set to `cmp::max(get_time().sec as u64, best_block_timestamp + 1)`. Need to use the same rule here.

core/src/miner/mem_pool.rs
```diff
@@ -870,16 +994,26 @@ impl MemPool {
...
         let mut update_last_seq_to = None;
         {
             let by_seq = self.future.by_signer_public.row_mut(&public);
             if by_seq.is_none() {
                 return
             }
             let by_seq = by_seq.expect("None is tested in early-exit condition above; qed");
-            while let Some(order) = by_seq.remove(&current_seq) {
-                // remove also from priority and fee
+            while let Some(order) = by_seq.clone().get(&current_seq) {
```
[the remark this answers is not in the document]
joojis: I fixed it.

core/src/miner/mem_pool.rs
```diff
@@ -78,6 +79,40 @@ impl ParcelOrigin {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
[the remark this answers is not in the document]
joojis: I fixed and added the unit test.

core/src/codechain_machine.rs
```diff
@@ -87,6 +102,68 @@ impl CodeChainMachine {
...
+                ..
+            } => inputs,
+            _ => return Ok(()),
+        };
+        for input in inputs {
+            if let Some(timelock) = &input.timelock {
+                match timelock {
+                    Timelock::Block(value) if value > &header.number() => {
+                        return Err(StateError::Transaction(TransactionError::Timelocked {
+                            timelock: timelock.clone(),
```
sgkim126: How about deriving Copy for Timelock and copying it instead of cloning?

types/src/transaction/error.rs
```diff
@@ -71,6 +76,8 @@ pub enum Error {
     EmptyOutput,
 }
 
+// FIXME: WIP
+const ERROR_ID_TIMELOCKED: u8 = 30u8;
```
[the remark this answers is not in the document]
joojis: Sorry, I was just avoiding the conflicts. I moved it to the last which is `26u8`.

core/src/codechain_machine.rs
joojis: I fixed it. And it's much better.

types/src/transaction/transaction.rs
```diff
@@ -51,10 +51,52 @@ pub struct AssetOutPoint {
...
+#[serde(rename_all = "camelCase", tag = "type", content = "value")]
+pub enum Timelock {
+    Block(u64),
+    BlockAge(u64),
+    Time(u64),
+    TimeAge(u64),
+}
+
+type TimelockType = u8;
+const BLOCK: TimelockType = 0x00;
```
sgkim126: This is a trivial tip, but it is recommended not to use 0 as the ID to be used for RLP.
The zero is converted to `0x80` and it sometimes gives a confusion.

types/src/transaction/transaction.rs
```diff
@@ -469,7 +512,8 @@ fn apply_input_scheme(
 
     if is_sign_single {
         return vec![AssetTransferInput {
-            prev_out: cur.clone(),
+            prev_out: cur.prev_out.clone(),
+            timelock: cur.timelock.clone(),
```
sgkim126:
```rust
            timelock: cur.timelock,
```

types/src/transaction/transaction.rs
```diff
@@ -566,6 +610,7 @@ impl PartialHashing for Transaction {
                         nonce: *nonce,
                         input: AssetTransferInput {
                             prev_out: input.prev_out.clone(),
+                            timelock: input.timelock.clone(),
```
sgkim126:
```rust
            timelock: input.timelock
```

core/src/client/client.rs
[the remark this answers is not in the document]
sgkim126: Currently, I'm working on the allowing duplicate transactions.
I'll make it check whether the transaction success at the commit.

core/src/miner/mem_pool.rs
```diff
@@ -17,6 +17,7 @@
 use std::cmp;
 use std::cmp::Ordering;
 use std::collections::{BTreeSet, HashMap};
+use time::get_time;
```
sgkim126:
```rust
use std::collections::{BTreeSet, HashMap};

use ckey::{public_to_address, Public};
...
use time::get_time;
```

core/src/miner/mem_pool.rs
```diff
@@ -792,8 +831,47 @@ impl MemPool {
...
+            &parcel.insertion_time,
+            timestamp,
+        );
+
+        if Self::should_wait_timelock(&parcel.timelock, parcel.insertion_time, timestamp) {
+            // Check same seq is in current. If it
+            // is than move the following current items to future.
+            let mut moved_to_future_flag = false;
+            let best_block_number = parcel.insertion_time;
+            if let Some(_) = self.current.by_signer_public.get(&signer_public, &seq) {
```
sgkim126:
```rust
            let moved_to_future_flag = if let Some(_) = self.current.by_signer_public.get(&signer_public, &seq) {
                self.move_all_to_future(&signer_public, state_seq);
                true
            } else {
                false
            };
````

File: core/src/miner/mem_pool.rs [hunk 13]
```diff
@@ -792,8 +831,47 @@ impl MemPool {
...
+                moved_to_future_flag = true;
+                self.move_all_to_future(&signer_public, state_seq);
+            }
+
+            check_too_cheap(Self::replace_parcel(
```
sgkim126: Would you explain why it's added here?

core/src/miner/mem_pool.rs
```diff
@@ -792,8 +831,47 @@ impl MemPool {
...
+                    signer_public,
+                    state_seq,
+                    state_seq,
+                    &best_block_number,
+                    timestamp,
+                );
+            }
+
+            let removed = self.future.enforce_limit(&mut self.by_hash, &mut self.local_parcels);
+            check_if_removed(&signer_public, &seq, removed)?;
```
sgkim126: How about add logging here?

core/src/miner/mem_pool.rs
joojis: I edited.

core/src/miner/mem_pool.rs
joojis: Fixed.

core/src/miner/mem_pool.rs
joojis: I added logging code.
```
            cdebug!(MEM_POOL, "Imported parcel to future: {:?}", hash);
            cdebug!(MEM_POOL, "status: {:?}", self.status());
```

core/src/miner/mem_pool.rs
sgkim126: Sorry to bother you, but the below code looks better.
```rust
let moved_to_future_flag = self.current.by_signer_public.get(&signer_public, &seq).is_some();
if moved_to_future_flag {
    self.move_all_to_future(&signer_public, state_seq);
}
```

PR #829, opened 2018-10-25
Remove World

sgkim126:
Currently, there is a world in the shard and world determines who can mint new assets.
It removes world and shard user determines who can mint new assets.
So minting is possible if the address is the user of the Shard.
And there is a free shard that has no users. Anyone can mint in the free shard.

And this patch also removes ShardMetadata, which stores the number of shards, because it has no information.


sgkim126:
This PR is related to https://github.com/CodeChain-io/codechain-sdk-js/pull/242

PR #830, opened 2018-10-25
There is no need to cache cargo registry

sgkim126:


sgkim126:
Currently, the travis cache size is over 4GB.

PR #835, opened 2018-10-26
Check if the filter has trailing zero. fix #813

### LINKED ISSUE #813: Disallow trailing zero bytes in Tag
The trailing zero bytes in Tag are redundant.

GNUp:


sgkim126:
LGTM

types/src/util/tag.rs
```diff
@@ -36,6 +37,11 @@ impl Tag {
             return Err(HashingError::InvalidFilter)
         }
 
+        // Check if the filter has trailing zero
+        if bitvec[0] == 0 {
```
joojis: It seems that you need to handle zero-length `bitvec`.

```
--- tests::executor_tests::chk_sig::sign_single_input_all_output stdout ----
thread 'tests::executor_tests::chk_sig::sign_single_input_all_output' panicked at 'index out of bounds: the len is 0 but the index is 0', /Users/travis/build/rust-lang/rust/src/libcore/slice/mod.rs:2079:10
```

GNUp: I fixed it and added a test case for this.

PR #838, opened 2018-10-29
Reduce the Travis cache size.

sgkim126:


sgkim126:
Because there is no option to remove the old binaries, the size of the target directory grows steadily. It makes Travis slow.

sgkim126:
@jhs7jhs 
Currently, the cache size becomes over 5 GB.
I'll merge it until we find a better option.

PR #842, opened 2018-11-01
Make CodeChain allow the duplicated transactions

sgkim126:
Currently, CodeChain rejects the duplicated transactions, so asset
transactions have meaningless nonce to change the hash of the
transaction.

This patch makes CodeChain allow the duplicated transactions. So the
nonce of asset transactions are removed and changes
chain_getTransactionInvoice to to chain_getTransactionInvoices.

### DIFF
--- a/test/src/integration/mempool.test.ts
+++ b/test/src/integration/mempool.test.ts
@@ -252,11 +252,13 @@ describe("Timelock", () => {
     }
 
     async function checkTx(txhash: H256, shouldBeConfirmed: boolean) {
-        const invoice = await node.sdk.rpc.chain.getTransactionInvoice(txhash);
+        const invoices = await node.sdk.rpc.chain.getTransactionInvoices(
+            txhash
+        );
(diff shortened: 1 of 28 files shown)


sgkim126:
This patch is related to https://github.com/CodeChain-io/codechain-sdk-js/pull/246

test/src/integration/mempool.test.ts
```diff
@@ -331,18 +331,17 @@ describe("Timelock", () => {
             recipient: await node.createP2PKHAddress()
         });
         await node.signTransferInput(tx, 0);
-        await node.sendTransaction(tx, { awaitInvoice: false });
+        try {
+            await node.sendTransaction(tx, { awaitInvoice: false });
+            done.fail();
+        } catch (e) {
+            expect(e.data).toContain("Timelocked");
```
sgkim126: I changed the expected result that depends on the failed transaction.

joojis: Ok.

PR #844, opened 2018-11-02
Do not recompile on non-source changes

sgkim126:


sgkim126:
@majecty IMO, this kind of option needs to https://github.com/CodeChain-io/codechain-agent aslo.

PR #845, opened 2018-11-05
Increase the maximum of the asset amount

sgkim126:
Since we has a plan to support wrapped CCC, the asset amount must be able to cover the amount of the CCC.
So this patch changes the type of asset amount to U256.


sgkim126:
This PR is related to https://github.com/CodeChain-io/codechain-sdk-js/pull/248

Issue #831, opened 2018-10-25
Needs transaction types that convert CCC from/to an assset


kseo:
We need something similar to https://weth.io/

sgkim126:
It depends on https://github.com/CodeChain-io/codechain/pull/845.

jhs7jhs:
Now #852 is merged. Need to consider `chain_getAssetSchemeByType`, `chain_getAssetSchemeByHash` as the conversation on the PR.

jhs7jhs:
Deploying docker-build on https://travis-ci.org/CodeChain-io/codechain/builds/453774165?utm_source=github_status&utm_medium=notification

jhs7jhs:
Now implemented the SDK part. https://github.com/CodeChain-io/codechain-sdk-js/pull/250

jhs7jhs:
Added integration test: #865

PR #852, opened 2018-11-08
Add WrapCCC parcel type/AssetUnwrapCCC tx type

jhs7jhs:
From #831
WrapCCC is the parcel type for wrapping CCC into Wrapped CCC.
AssetUnwrapCCC is the tx type for unwrapping CCC from Wrapped CCC.

### DIFF
--- a/state/src/impls/shard_level.rs
+++ b/state/src/impls/shard_level.rs
@@ -23,7 +23,8 @@ use ckey::Address;
 use cmerkle::{self, Result as TrieResult, TrieError, TrieFactory};
 use ctypes::invoice::Invoice;
 use ctypes::transaction::{
-    AssetMintOutput, AssetTransferInput, AssetTransferOutput, Error as TransactionError, PartialHashing, Transaction,
+    AssetMintOutput, AssetTransferInput, AssetTransferOutput, AssetWrapCCCOutput, Error as TransactionError,
+    InnerTransaction, PartialHashing, Transaction,
 };
 use ctypes::util::unexpected::Mismatch;
 use ctypes::ShardId;
@@ -84,56 +85,72 @@ impl<B: AsHashDB> ShardLevelState<B> {
 
     fn apply_internal<C: ChainTimeInfo>(
         &mut self,
-        transaction: &Transaction,
+        transaction: &InnerTransaction,
         sender: &Address,
         shard_users: &[Address],
         client: &C,
     ) -> StateResult<()> {
         debug_assert_eq!(Ok(()), transaction.verify());
         match transaction {
-            Transaction::AssetMint {
-                metadata,
-                registrar,
+            InnerTransaction::General(transaction) => match transaction {
+                Transaction::AssetMint {
+                    metadata,
+                    registrar,
+                    output:
+                        AssetMintOutput {
+                            lock_script_hash,
+                            amount,
+                            parameters,
+                        },
+                    ..
+                } => Ok(self.mint_asset(
+                    transaction.hash(),
+                    metadata,
+                    lock_script_hash,
+                    parameters,
+                    amount,
+                    registrar,
+                    sender,
+                    shard_users,
+                    Vec::new(),
+                )?),
+                Transaction::AssetTransfer {
+                    burns,
+                    inputs,
+                    outputs,
+                    ..
+                } => {
+                    debug_assert!(outputs.len() <= 512);
+                    self.transfer_asset(&transaction, sender, burns, inputs, outputs, client)
+                }
+
+                Transaction::AssetCompose {
+                    metadata,
+                    registrar,
+                    inputs,
+                    output,
+                    ..
+                } => self.compose_asset(&transaction, metadata, registrar, inputs, output, sender, shard_users, client),
+                Transaction::AssetDecompose {
+                    input,
+                    outputs,
+                    ..
+                } => self.decompose_asset(&transaction, input, outputs, sender, client),
+                Transaction::AssetUnwrapCCC {
+                    burn,
+                    ..
+                } => self.unwrap_ccc(&transaction, sender, burn, client),
+            },
+            InnerTransaction::AssetWrapCCC {
+                parcel_hash,
                 output:
-                    AssetMintOutput {
+                    AssetWrapCCCOutput {
                         lock_script_hash,
                         amount,
                         parameters,
                     },
                 ..
-            } => Ok(self.mint_asset(
-                transaction.hash(),
-                metadata,
-                lock_script_hash,
-                parameters,
-                amount,
-                registrar,
-                sender,
-                shard_users,
-                Vec::new(),
-            )?),
-            Transaction::AssetTransfer {
-                burns,
-                inputs,
-                outputs,
-                ..
-            } => {
-                debug_assert!(outputs.len() <= 512);
-                self.transfer_asset(&transaction, sender, burns, inputs, outputs, client)
-            }
-
-            Transaction::AssetCompose {
-                metadata,
-                registrar,
-                inputs,
-                output,
-                ..
-            } => self.compose_asset(&transaction, metadata, registrar, inputs, output, sender, shard_users, client),
-            Transaction::AssetDecompose {
-                input,
-                outputs,
-                ..
-            } => self.decompose_asset(&transaction, input, outputs, sender, client),
+            } => self.wrap_ccc(parcel_hash, lock_script_hash, parameters, amount),
         }
     }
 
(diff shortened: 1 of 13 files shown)


state/src/impls/shard_level.rs
```diff
@@ -427,6 +443,64 @@ impl<B: AsHashDB> ShardLevelState<B> {
...
+    ) -> StateResult<()> {
+        if !shard_users.is_empty() && !shard_users.contains(sender) {
+            return Err(TransactionError::InsufficientPermission.into())
+        }
+        let asset_scheme_address = AssetSchemeAddress::new_with_zero_suffix(self.shard_id);
+        let mut asset_scheme = self.get_asset_scheme_mut(&asset_scheme_address)?;
+        if asset_scheme.is_null() {
+            // FIXME: Wrapped CCC is minted in here, but the metadata is not well-defined.
+            asset_scheme.init(
+                format!("{\"name\":\"Wrapped CCC\",\"description\":\"Wrapped CCC in shard {}\"}", self.shard_id),
```
```rust
format!("{{\"name\":\"Wrapped CCC\",\"description\":\"Wrapped CCC in shard {}\"}}", self.shard_id),
```

types/src/transaction/transaction.rs
```diff
@@ -295,6 +319,51 @@ impl Transaction {
...
+                Ok(())
+            }
+        }
+    }
+
+    pub fn convert_inner(&self) -> InnerTransaction {
+        InnerTransaction::General(self.clone())
+    }
+
+    pub fn unwrapping_cost(&self) -> U256 {
```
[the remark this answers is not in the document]
jhs7jhs: Changed to `unwrapped_amount`.

jhs7jhs:
@sgkim126 Could you check `fn apply_transaction_for_shard()`?
Currently the receiver of the `AssetUnwrapCCC` is the sender.

types/src/transaction/transaction.rs
```diff
@@ -295,10 +319,63 @@ impl Transaction {
...
+    }
+
+    pub fn verify(&self) -> Result<(), Error> {
+        match self {
+            InnerTransaction::General(transaction) => transaction.verify(),
+            InnerTransaction::AssetWrapCCC {
+                output,
+                ..
+            } => match output.amount {
+                Some(amount) if amount.is_zero() => Err(Error::ZeroAmount),
```
sgkim126: What happens if the `amount` is `None`?

jhs7jhs: It happened because I just tried to use `AssetMintOutput` as the output format. It might be good to make a type for it...

state/src/impls/shard_level.rs
```diff
@@ -427,6 +444,64 @@ impl<B: AsHashDB> ShardLevelState<B> {
...
+        &mut self,
+        parcel_hash: &H256,
+        lock_script_hash: &H160,
+        parameters: &Vec<Bytes>,
+        amount: &Option<U256>,
+        sender: &Address,
+        shard_users: &[Address],
+    ) -> StateResult<()> {
+        if !shard_users.is_empty() && !shard_users.contains(sender) {
+            return Err(TransactionError::InsufficientPermission.into())
```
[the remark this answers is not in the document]
jhs7jhs: I think it can be deleted. We can regard wrapped CCC assets as minted ones.

types/src/transaction/transaction.rs
jhs7jhs: I've changed the type to `AssetWrapCCCOutput`, newly defined one.

types/src/transaction/transaction.rs
```diff
@@ -146,6 +150,25 @@ pub struct AssetMintOutput {
...
+        parcel_hash: H256,
+        output: AssetWrapCCCOutput,
+    },
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AssetWrapCCCOutput {
+    pub lock_script_hash: H160,
+    pub parameters: Vec<Bytes>,
+    pub amount: U256,
```
sgkim126: @joojis What about asset composition? Does it allow infinite quantity?

jhs7jhs:
@sgkim126 There might be some bugs so I'll check more.

types/src/transaction/transaction.rs
joojis: What is that mean? infinite quantity?

Inputs of the composition can be `U256` since `Asset`'s amount is U256.

sgkim126: Because the current `AssetCompose` transaction reuses `AssetMintOutput`, the type of `amount` is `Option<U256>`.

joojis: Currently, all the AssetCompose transactions verify that `amount == 1`.

https://github.com/CodeChain-io/codechain/blob/bbd0f62c70c9ec79214521ce140e03a16e55c809/types/src/transaction/transaction.rs#L271

jhs7jhs:
@sgkim126 Also, I'm concerned about some behaviors, sorry for this, I think it would be good to ask someone instead of reading the code,
1. Is `InnerTransaction::AssetWrapCCC` stored in the DB?
2. How do `chain_getTransaction` and `chain_getTransactionInvoices` work on the hash of parcels with `WrapCCC` action?
3. Is it good to mint wrapped CCC when the first WrapCCC parcel is done? (`chain_getAssetSchemeByType`, `chain_getAssetSchemeByHash`)

types/src/transaction/transaction.rs
sgkim126: Does it mean that the amount of composed asset is always 1?
If that how about add new output type that has no amount?

state/src/impls/top_level.rs
```diff
@@ -481,10 +537,15 @@ impl TopLevelState {
...
         let mut shard_level_state = ShardLevelState::from_existing(shard_id, self.db.clone(), shard_root)?;
 
-        let invoice = shard_level_state.apply(transaction, sender, &shard_users, client)?;
-        let (new_root, db) = shard_level_state.drop();
+        let invoice = shard_level_state.apply(&transaction.clone().into(), sender, &shard_users, client)?;
         match &invoice {
             Invoice::Success => {
+                let unwrapped_amount = transaction.unwrapped_amount();
+                if !unwrapped_amount.is_zero() {
+                    self.add_balance(sender, &unwrapped_amount)?;
```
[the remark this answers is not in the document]
sgkim126: It must be after `self.db = db;` statement.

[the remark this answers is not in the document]
sgkim126: https://github.com/CodeChain-io/codechain/pull/854 will make it share `db` and it will fix this constraint. But the current implementation replaces `db`, so changing top level state must be done after finish the shard level state changing.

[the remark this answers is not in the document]
jhs7jhs: Thanks, I will fix this.

sgkim126:
> @sgkim126 Also, I'm concerned about some behaviors, sorry for this, I think it would be good to ask someone instead of reading the code,
> 
> 1. Is `InnerTransaction::AssetWrapCCC` stored in the DB?
I think it will not.
The parcels are stored in the DB.

> 2. How do `chain_getTransaction` and `chain_getTransactionInvoices` work on the hash of parcels with `WrapCCC` action?
I think chain_getTransaction/TransactionInvocies should not work for `WrapCCC`.
`InnerTransaction::AssetWrapCCC` type must be the intermediate type for implementation. User won't know whether it exist.

> 3. Is it good to mint wrapped CCC when the first WrapCCC parcel is done? (`chain_getAssetSchemeByType`, `chain_getAssetSchemeByHash`)
I'll think about it. Actually, I didn't consider the RPCs.

state/src/impls/top_level.rs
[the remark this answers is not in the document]
sgkim126: Sorry. It must be moved to `apply_transaction` method. The current position executed several time if the inputs of the `UnwrapCCC` transaction are in multiple shards.

types/src/transaction/transaction.rs
joojis: Yes.

And like this?
```
AssetComposeOutput {
    pub lock_script_hash: H160,
    pub parameters: Vec<Bytes>,
}
```
Sounds fair to me.

jhs7jhs:
@sgkim126 Sorry for that, I've missed while changing the code.

state/src/impls/top_level.rs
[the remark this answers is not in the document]
jhs7jhs: Yes, that was what I was concerned about, but I thought it is okay because `transaction.related_shards()` only return one shard on `AssetUnwrapCCC`.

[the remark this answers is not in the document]
jhs7jhs: But it's not proper to put in there by its meaning. I'll fix it

types/src/transaction/transaction.rs
jhs7jhs: Also sounds fair to me. Those outputs are quite different from the others. But this one is not considered in this PR.

sgkim126:
@jhs7jhs Would add a test case to check UnwrapCCC increases the balance and WrapCCC decreases the balance?

types/src/transaction/transaction.rs
sgkim126: Okay. Let's do it in the other PR.

jhs7jhs:
@sgkim126 So, is it okay with the current design? I'll add test cases if so.

sgkim126:
I think the only left issue is whether to create an asset scheme for WrappedCCC or not.
But I do not think this will change the design much, so let's go ahead and discuss it later.

jhs7jhs:
Added basic test cases on wrapping/unwrapping CCCs.
Also, `PartialHashing` was not implemented on `AssetUnwrapCCC`, so I've implmented it. A tag should satisfy `tag.sign_all_inputs && tag.sign_all_outputs`.

types/src/transaction/transaction.rs
```diff
@@ -530,12 +620,35 @@ impl PartialHashing for Transaction {
                     &blake128(tag.get_tag()),
                 ))
             }
+            Transaction::AssetUnwrapCCC {
+                network_id,
+                burn,
+            } => {
+                if !tag.sign_all_inputs || !tag.sign_all_outputs {
```
sgkim126: Why does it need to be sign_all_inputs?

sgkim126: @joojis

state/src/impls/top_level.rs
```diff
@@ -465,6 +533,11 @@ impl TopLevelState {
                 return Err(ParcelError::InconsistentShardOutcomes.into())
             }
         }
+
+        let unwrapped_amount = transaction.unwrapped_amount();
+        if !unwrapped_amount.is_zero() {
```
sgkim126: Can unwrapped_amount be zero?

types/src/transaction/transaction.rs
joojis: I just wanted to disallow creating anothor valid transaction by manipulating the signature tag.
And that's the default value that the SDK creates.

state/src/impls/top_level.rs
jhs7jhs: On other transaction types can be zero.

jhs7jhs: Currently, only on `AssetUnwrapCCC` type will return non-zero value. I don't think it should be a function.

PR #853, opened 2018-11-08
Add structured logger

majecty:


util/logger/src/logger.rs
```diff
@@ -86,6 +88,13 @@ impl Log for Logger {
             let log_target = record.target();
             let log_message = record.args();
             eprintln!("#{} {} {} {} {}  {}", instance_id, timestamp, thread_name, log_level, log_target, log_message);
+
+            slogger.log(structured_logger::Log {
+                level: format!("{}", log_level),
```
sgkim126: Use to_string.

util/logger/src/structured_logger.rs
```diff
@@ -0,0 +1,86 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use parking_lot::Mutex;
+use std::sync::mpsc::{channel, Receiver, Sender};
+use std::sync::Arc;
+
+use serde::Serialize;
```
sgkim126: Move parking_lot here.

util/logger/src/structured_logger.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub timestamp: String,
+}
+
+impl StructuredLogger {
+    pub fn create() -> StructuredLogger {
+        let (sender, receiver) = channel();
+        StructuredLogger {
+            sender: Arc::new(Mutex::new(sender)),
+            receiver: Arc::new(Mutex::new(receiver)),
+            enabled: Arc::new(Mutex::new(false)),
```
sgkim126: How about use AtomicBool?

util/logger/src/structured_logger.rs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        let serialized_log = match to_value(log) {
+            Ok(value) => value,
+            Err(e) => {
+                eprintln!("Cannot serialize log : {}", e);
+                return
+            }
+        };
+        let sender = self.sender.lock().clone();
```
sgkim126:
```rust
let sender = Arc::clone(&self.sender.lock().clone())
```

util/logger/src/structured_logger.rs
```diff
@@ -0,0 +1,86 @@
...
+        if let Err(_) = sender.send(serialized_log) {
+            eprintln!("No receiver of structurred error");
```
sgkim126: ?

util/logger/src/structured_logger.rs
majecty: I don't know well about atomic ordering.
miner's sealing_enabled always use Ordering::Relaxed.
Is it OK to use only Ordering::Relaxed to `enabled`?

util/logger/src/structured_logger.rs
majecty: My intention was cloning Sender.
The cloned sender is removed after this block?
I think that the Arc is not needed.

util/logger/src/structured_logger.rs
sgkim126: Use `SeqCst`, if it doesn't harm the performance.

PR #854, opened 2018-11-08
Shard StateDB rather than cloning.

sgkim126:


sgkim126:
I'll merge it after merging #852.

PR #872, opened 2018-11-16
Use Uint instead of u64 for RPC

sgkim126:
And it makes the json serialization of Uint be the string, since
the JavaScript number type doesn't give a enough precision.


sgkim126:
I changed the seq and blanace to u64 in the previous patch. And it breaks the integration test of the SDK since the precision of JavaScript number is not enough to represent 64-bit integer.

PR #873, opened 2018-11-16
Fix bugs about wrap/unwrap

sgkim126:


sgkim126:
The first commit is just clean-up code.
The second commit fixes a bug caused by #869 
And the third commit fixes a bug from #852

PR #876, opened 2018-11-19
Adds RPCs requested by #841

### LINKED ISSUE #841: Needs additional RPCs
I need some RPCs for implementing the applications like the wallet or explorer.
It is currently hardcoded according to each chain consensus.
- [ ] Mining rewards
- [ ] Address list on the genesis
- [ ] Finalization time

sgkim126:
It fixes #841


sgkim126:
@joojis How about create `engine` category and move `getCoinbase` and `getBlockReward`?
They are related to the engine rather than the chain.

joojis:
@sgkim126 I think it's good to separate the category. A little concern is that I think the word `engine` is natural to the core developers only. But, still, it is ok.

PR #886, opened 2018-11-22
Move height increase logic in tendermint

majecty:
When new block is imported, tendermint increase its block height a
`is_proposal` function. Change the code to use ChainNotify instead.


core/src/consensus/tendermint/mod.rs
```diff
@@ -320,6 +326,7 @@ impl Tendermint {
     }
 
     fn to_next_height(&self, height: Height) {
+        assert_eq!(true, height >= self.height.load(AtomicOrdering::SeqCst));
```
```rust
assert!(height >= self.height.load(AtomicOrdering::SeqCst), "{} < {}", height, self.height.load(AtomicOrdering::SeqCst));
```

PR #887, opened 2018-11-22
Fix integration tests

jhs7jhs:
1. Remove U256
2. Remove unused import statements
3. Remove `{logFlag: true}` from creating CodeChain instances. 
4. Remove timeout field from `test.each`, and moved it to `jest.setTimeout()`
5. Fixed integration tests on transactions


jhs7jhs:
Merged from #889.

PR #889, opened 2018-11-23
Change the type of parcel fee from U256 to U64 in onchainparcel test

GNUp:


jhs7jhs:
@GNUp I think this should be merged into #887, because of the SDK.

jhs7jhs:
No, it's flaky. I think it's okay to close this PR.

PR #890, opened 2018-11-23
Update the version of testhelper

GNUp:
fix this [bug](https://github.com/CodeChain-io/codechain/pull/889#issuecomment-441154000)

### DIFF
--- a/test/package.json
+++ b/test/package.json
@@ -26,7 +26,7 @@
   },
   "dependencies": {
     "codechain-sdk": "https://github.com/jhs7jhs/codechain-sdk-js.git#build",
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
```
jhs7jhs: Why is this file created?

GNUp:
Sorry I didn't check it. I removed it.

PR #891, opened 2018-11-23
Make a Tendermint's block contain the previous block's signature

majecty:
Now seal's content is
   [current_view, prev_block_view, prev_block_signatures].


sgkim126:
Would you resolve #879 first?

core/src/blockchain/block_info.rs
```diff
@@ -14,17 +14,110 @@
...
 #[derive(Debug, Clone, PartialEq)]
-pub enum BlockLocation {
-    /// It's part of the canon chain.
-    CanonChain,
-    /// It's not a part of the canon chain.
-    Branch,
+pub enum BestBlockChanged {
+    /// Cannonical chain is appended.
+    CanonChainAppended {
+        best_block: Vec<u8>,
```
sgkim126: Why the type of `best_block` is `Vec<u8>`?

core/src/consensus/mod.rs
```diff
@@ -213,7 +212,7 @@ pub trait ConsensusEngine<M: Machine>: Sync + Send {
     }
 
     /// Broadcast a block proposal.
-    fn broadcast_proposal_block(&self, _block: SealedBlock) {}
+    fn broadcast_proposal_block(&self, _verified_block_data: &[u8]) {}
```
Kais-DkM: Why did you change this?

core/src/blockchain/block_info.rs
```diff
@@ -14,17 +14,110 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+use std::option::Option;
```
Kais-DkM: Do you need to import this?

core/src/blockchain/block_info.rs
```diff
@@ -14,17 +14,110 @@
...
+            } => best_block,
+            BestBlockChanged::BranchBecomingCanonChain {
+                best_block,
+                ..
+            } => best_block,
+            BestBlockChanged::None => return None,
+        };
+
+        let block_view = BlockView::new(block);
+        Some(block_view.header().hash())
```
Kais-DkM: `BlockView` has `fn hash` too.

core/src/blockchain/body_db.rs
```diff
@@ -162,40 +176,53 @@ impl BodyDB {
...
+            best_block_hash
+        } else {
+            return HashMap::new()
+        };
+        let block = match best_block_changed.best_block() {
+            Some(block) => block,
+            None => return HashMap::new(),
+        };
+
+        // FIXME: is the block is new best block?
```
Kais-DkM: I can't think of the cases that `block` doesn't become a new best block for now.
Do you have any situation in mind?

core/src/blockchain/route.rs
```diff
@@ -120,26 +120,73 @@ pub struct ImportRoute {
...
+                ImportRoute {
+                    retracted,
+                    enacted,
+                    omitted,
+                }
+            }
+        }
+    }
+
+    pub fn new_from_best_hash_changed(new_block_hash: H256, best_header_changed: &BestHeaderChanged) -> Self {
```
Kais-DkM: I think `new_from_best_header_changed` would be a better name for this.

core/src/blockchain/route.rs
```diff
@@ -120,26 +120,73 @@ pub struct ImportRoute {
...
+            BestBlockChanged::None => ImportRoute {
+                retracted: vec![],
+                enacted,
+                omitted,
+            },
+            BestBlockChanged::BranchBecomingCanonChain {
+                tree_route,
+                ..
+            } => {
+                enacted.extend(tree_route.enacted.iter());
```
Kais-DkM: According to `TreeRoute`'s documentation, elements of `enacted` are sorted in ascending order by their block number.
Let's put the best header at the tail to be consistent with original implementation.

Kais-DkM:
It seems there are some cases where the highest scored block cannot be referred to as "best" block.
Can I have more details about those situations?

majecty:
> It seems there are some cases where the highest scored block cannot be referred to as "best" block.
> Can I have more details about those situations?

@Kais-DkM 

In the previous Tendermint implementation, If a block proposer collects more than 2/3 of pre-commits, the proposer propagates a block which contains pre-commits. If the proposer does not propagate the block, the Tendermint consensus stopped. 

After this patch, the next block contains the previous block's signatures to remove block propagation step in the previous implementation. But if we apply this patch, the last block has no signatures of itself, so the last block should not be the best block.

Kais-DkM:
@majecty 
My understanding of the current situation is like this:
* Tendermint needs to inform other blocks for collecting prevotes/precommits, and original implementation sent full block data for this.
* Even though block data is sent to other peers, they didn't import these blocks until they receive it again.

To optimize this, this patch seems to use something like import-then-validate strategy. Accordingly, the highest scored block can be invalid, so best block doesn't necessarily match with the highest scored block.

I see the following problems here:
* `BlockChain` must contain only valid blocks in its storage. Every invalid block is rejected on verification and discarded forever after that. I think it's reasonable to think some parts in our code relies on that, but our new highest scored block has possibility of being invalid.
* Currently proposed block is consensus-specific state, so it seems bit unnatural for me to delegate these responsibilities to blockchain. I think consensus extension is the right place for managing those state.

majecty:
@Kais-DkM 

> To optimize this, this patch seems to use something like import-then-validate strategy. Accordingly, the highest scored block can be invalid, so best block doesn't necessarily match with the highest scored block.

Not for optimization, Its about Byzantine fault. If the proposer does not propagate the block(which signatures are aggregated), the Tendermint network stopped. Any other nodes could not propose a new block. Only one Byzantine node could stop the network.

`BlockChain` contains blocks which are not included in the canonical chain as branch blocks. When other blocks are imported the branch blocks could be included in the canonical chain. The latest block of Tendermint consensus is similar to the branch blocks. The block is saved but not in the canonical chain until a new block is imported. I think it is not strange to think that the latest block of the Tendermint is a branch block.

It's a good idea to put all code about propose block into the Tendermint consensus extension. However, it is not possible to implement all the code there. A proposal block should be validated like other blocks. And, to finalize the best block in the tendermint consensus, the next block(highest scored block) is needed. So we should save the highest block somewhere. It is to difficult to save a block in a disk in the Tendermint extension.

Kais-DkM:
> The block is saved but not in the canonical chain until a new block is imported. I think it is not strange to think that the latest block of the Tendermint is a branch block.

The difference is that branch blocks are valid blocks that pass the entire validation process, but proposal block does not. It remains invalid until it collects 2/3 precommits. Actually, it won't be a big problem if the tendermint consensus is going well. But if it fails to collect enough precommit and a new block is proposed, this invalid block will live forever in the storage as if it was a valid block.

>So we should save the highest block somewhere. It is to difficult to save a block in a disk in the Tendermint extension.

Is there any reason for storing this block in a disk?

Kais-DkM:
Also, I agree that it is really hard to execute full verification inside the consensus extension without sending it to the verification queue. How about considering `client::import_verified_blocks` or equivalently `importer::import_verified_blocks` as a starting point?

majecty:
@Kais-DkM 

> 
> The difference is that branch blocks are valid blocks that pass the entire validation process, but proposal block does not. It remains invalid until it collects 2/3 precommits. Actually, it won't be a big problem if the tendermint consensus is going well. But if it fails to collect enough precommit and a new block is proposed, this invalid block will live forever in the storage as if it was a valid block.
> 

It is the same with the PoW consensus. PoW's branch block also saved in the disk and it is not removed. It will live forever in the storage as if it was a canonical chain. And creating a proposal is not that easy. only a valid order of proposer could create and propagate a proposal. Also, If a proposer proposes more than a block, then the proposer will be published(this is not implemented, but Seongchan will implement it). So if you worried about the data storage, it should be implemented in another place to delete branch blocks of PoW consensus and Tendermint consensus.

> Is there any reason for storing this block in a disk?

If we did not store the proposal block, a node could lose the best block's signature.

Example) 
If we do not save the proposal block.

1. node A received a proposal with height `h`.
2. node A changed the best block to a block whose height is `h - 1`.
3. node A rebooted.
4. node A thinks it has a finalized block at height `h - 1`, but it does not have signatures for them.

Kais-DkM:
After a conversation, we concluded that the goal of this PR is a correct behavior.
I can now agree that non-best highest scored block is safe to be stored.

core/src/consensus/tendermint/mod.rs
```diff
@@ -1143,87 +1168,64 @@ mod tests {
...
+            current_view: &0,
+            last_block_signatures: &vec![signature0],
+        }
+        .seal_fields();
+        header.set_seal(seal);
 
         // One good signature is not enough.
         match engine.verify_block_external(&header) {
             Err(Error::Engine(EngineError::BadSealFieldSize(_))) => {}
-            _ => panic!(),
```
sgkim126: Why?

core/src/consensus/solo/mod.rs
```diff
@@ -60,7 +60,7 @@ where
     }
 
     fn generate_seal(&self, _block: &M::LiveBlock, _parent: &M::Header) -> Seal {
```
sgkim126: What does `Some` means?

core/src/consensus/mod.rs
```diff
@@ -106,9 +97,11 @@ pub trait ConsensusEngine<M: Machine>: Sync + Send {
     /// It is fine to require access to state or a full client for this function, since
     /// light clients do not generate seals.
     fn generate_seal(&self, _block: &M::LiveBlock, _parent: &M::Header) -> Seal {
```
sgkim126: Is there a use case of this default implementation?

core/src/consensus/mod.rs
```diff
@@ -215,7 +208,7 @@ pub trait ConsensusEngine<M: Machine>: Sync + Send {
 
     /// Called when proposal block is verified.
     /// Consensus many hold the verified proposal block until it should be imported.
-    fn on_verified_proposal(&self, _header: &Header) {}
+    fn on_verified_proposal(&self, _verified_block_data: &[u8]) {}
```
sgkim126: Should we pass the bytes? Is there no abstraction of Block?

core/src/consensus/mod.rs
```diff
@@ -54,16 +54,7 @@ use crate::scheme::CommonParams;
...
-#[derive(Debug, PartialEq, Eq)]
-pub enum Seal {
-    /// Proposal seal; should be broadcasted, but not inserted into blockchain.
-    Proposal(Vec<Bytes>),
-    /// Regular block seal; should be part of the blockchain.
-    Regular(Vec<Bytes>),
-    /// Engine does generate seal for this block right now.
-    None,
-}
+pub type Seal = Option<Vec<Bytes>>;
```
sgkim126: What is the available format of the Seal?

core/src/consensus/mod.rs
majecty: The caller site has a pre-verified block(`PreverifiedBlock`) and closed block(`ClosedBlock`).

```
let pre_verified_block = ...;
let closed_block = verify(pre_verified_block);
on_verified_proposal(&pre_verified_block.bytes);
```

We can't use the closed block because the closed block's bytes are not the same with the pre-verified block's.
We can't use the pre-verified block because the pre-verified block's type says it is not verified yet. Receiving pre-verified block in `on_verified_proposal` is awkward.

There is also encoded::Block type, I don't know when to use the encoded::Block yet.
I'll check it now.

core/src/consensus/solo/mod.rs
majecty: Before this commit, the seal type is Regular OR Proposal OR None. Now there only remains Regular. So I changed the seal type to Option.

core/src/consensus/tendermint/mod.rs
majecty: I'll remove it now.

core/src/consensus/mod.rs
majecty: Since PoW consensus uses an extra seal, it uses the default implementation of generate_seal.

core/src/consensus/mod.rs
sgkim126: How about create an Enum has three types?

majecty: It's a good idea. I'll start right now.

core/src/consensus/mod.rs
majecty: I think `encoded::Block` is enough.

`BlockView` is a wrapper of `rlp<'a>`, and provide getter functions. `encoded::Block` owns `Vec<u8>` and provide `BlockView`'s getters.

Kais-DkM:
ACK. I don't have any additional comments for now.

core/src/consensus/tendermint/message.rs
```diff
@@ -157,10 +141,9 @@ pub fn consensus_view(header: &Header) -> Result<View, ::rlp::DecoderError> {
     UntrustedRlp::new(view_rlp.as_slice()).as_val()
 }
 
-/// Proposal signature.
-pub fn proposal_signature(header: &Header) -> Result<Signature, ::rlp::DecoderError> {
-    UntrustedRlp::new(header.seal().get(1).expect("seal passed basic verification; seal has 3 fields; qed").as_slice())
-        .as_val()
+pub fn previous_block_view(header: &Header) -> Result<View, ::rlp::DecoderError> {
+    let view_rlp = header.seal().get(1).expect("seal passed basic verification; seal has 3 fields; qed");
```
sgkim126: Isn't the second item of tendermint seal `cur_view`?

sgkim126: Ignore the previous comment, I found that `seal_fields()` makes `cur_view` the first item.

core/src/consensus/mod.rs
```diff
@@ -44,27 +44,48 @@ use ctypes::util::unexpected::{Mismatch, OutOfBounds};
...
+        match self {
+            Seal::None => None,
+            Seal::Solo => Some(Vec::new()),
+            Seal::SimplePoA(signature) => Some(vec![::rlp::encode(signature).into_vec()]),
+            Seal::Tendermint {
+                prev_view,
+                cur_view,
+                signatures,
+            } => Some(vec![
+                ::rlp::encode(cur_view).into_vec(),
```
sgkim126: It made me confused.
How about swapping the order of `cur_view` and `prev_view`?

core/src/consensus/tendermint/mod.rs
```diff
@@ -74,12 +77,13 @@ pub struct Tendermint {
     lock_change: RwLock<Option<ConsensusMessage>>,
     /// Last lock view.
     last_lock: AtomicUsize,
-    /// Bare hash of the proposed block, used for seal submission.
```
sgkim126: What's the difference between bare hash and hash?

core/src/consensus/tendermint/mod.rs
```diff
@@ -195,6 +204,16 @@ impl Tendermint {
         self.check_above_threshold(aligned_count).is_ok()
     }
 
+    fn has_enough_precommit_votes(&self, block_hash: H256) -> bool {
+        let vote_step = VoteStep::new(
+            self.height.load(AtomicOrdering::SeqCst),
```
sgkim126: Why don't you use `self.height()`?

core/src/consensus/tendermint/mod.rs
```diff
@@ -254,6 +282,8 @@ impl Tendermint {
         self.view.store(0, AtomicOrdering::SeqCst);
         *self.lock_change.write() = None;
         *self.proposal.write() = None;
+        *self.proposal_parent.write() = Default::default();
```
sgkim126: This lock is unlocked when the statement is finished.
Are we okay if the other thread looks the partially updated states?

core/src/consensus/tendermint/mod.rs
```diff
@@ -435,13 +479,7 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
         let seal_length = header.seal().len();
         let expected_seal_fields = self.seal_fields(header);
         if seal_length == expected_seal_fields {
```
sgkim126: How about check whether the seal is a Tendermint seal?

core/src/miner/miner.rs
```diff
@@ -540,41 +540,34 @@ impl Miner {
...
                 *self.next_mandatory_reseal.write() = Instant::now() + self.options.reseal_max_period;
-                {
-                    let mut sealing_work = self.sealing_work.lock();
-                    sealing_work.queue.push(block.clone());
-                    sealing_work.queue.use_last_ref();
+                if self.engine.is_proposal(block.header()) {
+                    block
+                        .lock()
+                        .seal(&*self.engine, seal.clone())
+                        .map(|sealed| {
```
```rust
                        .map(|sealed| {
                            self.engine.proposal_generated(&sealed);
                            self.engine.broadcast_proposal_block(sealed);
                        })
                        .unwrap_or_else(|e| {
                            cwarn!(MINER, "ERROR: seal failed when given internally generated seal: {}", e);
                        }).is_ok()
```

core/src/consensus/tendermint/mod.rs
```diff
@@ -658,24 +696,44 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
...
+        };
+
+        // if next header is imported, current header is not a proposal
+        if c.block_header(&BlockId::Number(number + 1)).map(|next| next.parent_hash() == header.hash()).unwrap_or(false)
+        {
             return false
         }
-        true
+
+        !self.has_enough_precommit_votes(header.hash())
```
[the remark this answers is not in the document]
majecty: When proposer creates a block, the `miner` queries to the engine whether the block(just generated) is a proposal or not. At that time, the block has no pre-commit votes.

core/src/consensus/tendermint/mod.rs
majecty: Bare hash does not contain seal.
This was used when a proposal block's seal is different from the imported block to get the same hash by not including the seal in the hash.
Now we use the same block as a proposal block and an imported block we do not need the bare hash.

core/src/consensus/mod.rs
majecty: Good

core/src/consensus/tendermint/mod.rs
majecty: I don't know well. There are too many variables which could be changed by other threads.

PR #896, opened 2018-11-26
Fix Tendermint thread bug

majecty:
Before this commit:
Two or more threads could call `handle_valid_message` at the same time.
If there are enough pre-commits, `to_step(Step::Commit)` may be called
twice. This stopped the node.

After this commit:
Only one thread could call `handle_valid_message` at a time.

This patch is not thoroughly tested. Should run Tendermint network more than
a few days to check this bug.

### DIFF
--- a/core/src/consensus/tendermint/mod.rs
+++ b/core/src/consensus/tendermint/mod.rs
@@ -28,7 +28,7 @@ use cnetwork::{Api, NetworkExtension, NetworkService, NodeId, TimeoutHandler, Ti
 use ctypes::machine::WithBalances;
 use ctypes::util::unexpected::{Mismatch, OutOfBounds};
 use ctypes::BlockNumber;
-use parking_lot::{Mutex, RwLock};
+use parking_lot::{Mutex, ReentrantMutex, RwLock};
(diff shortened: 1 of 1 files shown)


core/src/consensus/tendermint/mod.rs
```diff
@@ -734,6 +736,7 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
                 return Err(EngineError::DoubleVote(sender))
             }
             ctrace!(ENGINE, "Handling a valid {:?} from {}.", message, sender);
+            let _guard = self.step_change_lock.lock();
```
sgkim126: Is it okay just locking here?
It seems that there are many paths to call `to_step` method.

majecty: The problem was that `to_next_height` was called twice in `handle_valid_message`, so height is increased twice. I fixed it by not calling `handle_valid_message` concurrently.
As you say, although we locked the `handle_valid_message`, `to_step` is called concurrently and maybe it causes some problem. I'll look into what happens if the `to_step` is called concurrently.

Issue #898, opened 2018-11-27
Tendermint lock previous block's prevote after committed


majecty:
Current behavior

In Tendermint consensus, if the node received a new "prevote" message on height `h` after a node transit to "commit" state on height `h`, the node is locked on the message's block.

Desirable behavior

A node should not lock on a block which already committed.

PR #899, opened 2018-11-27
Tendermint: do not lock on the height `h` if a block is committed in the height `h`

majecty:
Previous this commit

In Tendermint consensus, if the node received a new "prevote" message on
height h after a node transit to "commit" state on height h, the node is
locked on the message's block.

After this commit

A node locks on the height `h` which is already committed.

This fixes https://github.com/CodeChain-io/codechain/issues/898


jhs7jhs:
@majecty I've fixed hgith to height. Can you fix the commit name?

majecty:
@jhs7jhs  I fixed it.

Issue #779, opened 2018-10-02
Use mocha instead of jasmine for test runner


joojis:
It seems that mocha is able to know whether the test failed in afterEach().

- https://github.com/mochajs/mocha/issues/797
- https://mochajs.org/api/runner.js.html

But I haven't found a spec or documentation for `currentTest` field. It needs to be confirmed.

jhs7jhs:
@joojis It seems it's possible to do so. https://stackoverflow.com/questions/22031765/how-might-i-check-if-any-mocha-tests-failed-from-within-an-after-block

jhs7jhs:
I've tried to find out the way to detect test failures in `afterEach` in Jest, but it's still not possible.
There's no way to escape from this issue. I'll try to migrate the integration test.

PR #900, opened 2018-11-27
Migrate from Jest to Mocha and apply tslint

jhs7jhs:
Resolve: #779 
The code was not clear with tslint while using Jest, so I've put applying tslint with the PR.

### DIFF
--- a/test/jest.config.js
+++ b/test/jest.config.js
@@ -1,8 +1,0 @@
-module.exports = {
-  roots: ["<rootDir>/src"],
-  transform: {
-    "^.+\\.tsx?$": "ts-jest"
-  },
-  testRegex: "/src/.*\\.test\\.tsx?$",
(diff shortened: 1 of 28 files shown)


jhs7jhs:
There's another way to set timeout. Changing the code.

jhs7jhs:
Moved `this.timeout()` to `it().timeout()`. But `BeforeEach` and `AfterEach` does not have it, so there's still some lines using `this.timeout()`.

jhs7jhs:
Also, the most of the code does not use arrow functions anymore.
Check this: https://mochajs.org/#arrow-functions

joojis:
I'm just wondering, have you tried to check whether a test fails in `after()`?

jhs7jhs:
@joojis 
I heard it's possible to check, but the problem is that we've decided to migrate to Mocha because of some logging problems, and the test cases with `before()` and `after()` are quite hard to read its log files.

joojis:
@jhs7jhs Ok

PR #914, opened 2018-11-30
Fix the transaction type on RPC

jhs7jhs:
1. Removed `hash` field from the type. It is currently not used in the SDK, and not documented in the spec.
2. Added types to support `number`(by `usize`) and `U64`(by `Uint`).
3. Added simple test case for the transaction. Test cases will be added more after this PR.

### DIFF
--- a/rpc/src/v1/types/action.rs
+++ b/rpc/src/v1/types/action.rs
@@ -20,18 +20,16 @@ use ctypes::parcel::Action as ActionType;
 use ctypes::ShardId;
 use primitives::{Bytes, H160};
 
-use super::Transaction;
+use super::{Transaction, TransactionWithHash};
 
-#[derive(Debug, Deserialize, Serialize)]
+#[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase", tag = "action")]
 pub enum Action {
-    /// Transaction, can be either asset mint or asset transfer
     AssetTransaction {
         transaction: Transaction,
     },
     Payment {
         receiver: PlatformAddress,
-        /// Transferred amount.
         amount: Uint,
     },
     SetRegularKey {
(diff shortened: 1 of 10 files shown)


joojis:
Removing `hash` may affect the explorer or the indexer.

jhs7jhs:
@joojis Then the transaction type should be separated, and I think it is weird. Also it should be written in the spec.

jhs7jhs:
@joojis @xpdlf1004 Changed the spec. Please check them please.
@sgkim126 Could you review the core code part?

test/src/integration/chain.test.ts
```diff
@@ -388,8 +388,34 @@ describe("chain", function() {
...
+            data.data.output.lockScriptHash
+        }`;
+
+        await node.sdk.rpc
+            .sendRpcRequest("chain_executeTransaction", [
+                data,
+                faucetAddress.value
+            ])
+            .then(result => {
+                console.log(result);
```
sgkim126: Why test prints it to stdout?

rpc/src/v1/types/action.rs
```diff
@@ -55,36 +53,67 @@ pub enum Action {
     Custom(Bytes),
 }
 
-impl Action {
+#[derive(Debug, Deserialize, Serialize)]
```
sgkim126: You don't need to derive `Deserizlie` for it.

rpc/src/v1/types/action.rs
```diff
@@ -20,18 +20,16 @@ use ctypes::parcel::Action as ActionType;
 use ctypes::ShardId;
 use primitives::{Bytes, H160};
 
-use super::Transaction;
+use super::{Transaction, TransactionWithHash};
 
 #[derive(Debug, Deserialize, Serialize)]
```
sgkim126: You don't need to derive `Serizlie` for it.

test/src/integration/chain.test.ts
jhs7jhs: Oops, I'll fix this.

rpc/src/v1/types/action.rs
jhs7jhs: Fixed.

rpc/src/v1/types/action.rs
jhs7jhs: Fixed.

jhs7jhs:
@sgkim126 Resolved some conflicts. By the way, can clippy check unused trait derivations? I think there may be some unused `Serialize`/`Deserialize`.

sgkim126:
@jhs7jhs AFAIK, no.

rpc/src/v1/types/transaction_with_hash.rs
```diff
@@ -0,0 +1,134 @@
...
+                network_id,
+                burn,
+            } => TransactionWithHash::AssetUnwrapCCC {
+                network_id,
+                burn,
+                hash,
+            },
+        }
+    }
+}
```
sgkim126: Would you install https://github.com/editorconfig/editorconfig-vscode?
https://github.com/CodeChain-io/codechain/blob/master/.editorconfig#L10
There mustbe new line at the end since we set `insert_final_newline`.

PR #917, opened 2018-11-30
Fix the order of operation by parentheses

sgkim126:
@GNUp I added parentheses at 7bbb80c86151b6525bbd33cdd3eb33895b7d69e5 and I found the original code has a bug.
`+` operator has a higher precedence than `&` operator, so `0x10 + v[0]` is executed before `& 15u8` is done. Since `0x10` is `0b1_0000` and `15` is `0b1111`, the adding `0x10` is meaningless.
I think you intended `0x10 + (v[0] & 15u8)`.


GNUp:
@sgkim126 Sorry, it was my mistake. Thanks

PR #921, opened 2018-11-30
Register tendermint's chain_notify

majecty:


codechain/run_node.rs
```diff
@@ -231,6 +232,8 @@ pub fn run_node(matches: ArgMatches) -> Result<(), String> {
     let miner = new_miner(&config, &scheme, ap.clone())?;
     let client = client_start(&config.operating, &scheme, miner.clone())?;
 
+    scheme.engine.register_chain_notify(client.client().borrow());
```
sgkim126: Using `as_ref` instead of `borrow`.

majecty:
@sgkim126 I fixed it.

Issue #517, opened 2018-08-01
Check the "burns" field in the "AssetTransferTransaction" when queuing a parcel


majecty:
Solve this issue after the https://github.com/CodeChain-io/codechain-sdk-js/issues/103 is solved.

kseo:
@majecty What do you mean by "checking the burns field"?

majecty:
If we turn on the stratum, the crash is not shown.
However, if we turn off the stratum, it occurred.
I'm struggling with this problem.

majecty:
@kseo I'll check whether is there any duplicated inputs among  `inputs` field and `burns` field.

jhs7jhs:
@majecty When do we need to check it?

majecty:
I want to check below things.
* Whether inputs and burns have some entries or not whose prev_out is the same.
* Whether some burn entries exist or not which have the same prev_out.

PR #922, opened 2018-11-30
Check previous outputs of burns/inputs duplicated in tx 

jhs7jhs:
Resolves: #517


types/src/transaction/transaction.rs
```diff
@@ -492,6 +493,20 @@ impl HeapSizeOf for Transaction {
     }
 }
 
+fn is_prev_out_duplicated(burns: &[AssetTransferInput], inputs: &[AssetTransferInput]) -> Result<(), Error> {
```
sgkim126: How about renaming it to `verify_duplicated_prev_out`?
`is_*` seems to return a boolean.

types/src/transaction/transaction.rs
```diff
@@ -492,6 +493,20 @@ impl HeapSizeOf for Transaction {
...
+    let mut prev_out_set = HashSet::new();
+    for input in inputs.iter().chain(burns) {
+        let prev_out = (input.prev_out.transaction_hash, input.prev_out.index);
+        if !prev_out_set.insert(prev_out) {
+            return Err(Error::DuplicatedPreviousOutput {
+                transaction_hash: input.prev_out.transaction_hash.clone(),
```
sgkim126: `H256` derives `Copy`. No need to call `clone`.

types/src/transaction/transaction.rs
jhs7jhs: Changed to `check_duplication_in_prev_out`.

types/src/transaction/transaction.rs
jhs7jhs: Fixed.

Issue #457, opened 2018-07-23
Change unit test of stratum to use unused ports


Kais-DkM:
Some unit tests in `stratum` crate opens network socket.
e.g. https://github.com/CodeChain-io/codechain/blob/420b63e97f275e043ff4367140547372ba2da219/stratum/src/lib.rs#L455
While it works and serves its own purpose, it's not a good practice to depend on external environment in unit test.
For example, if socket is already open by other application, the test will fail even if there's nothing wrong with its logic.
We should remove such tests, and find other way to test functionalities in those files.

jhs7jhs:
@Kais-DkM I think it is inevitable by the structure of `Stratum`, but we can ensure the unit test works by checking socket addresses and using a socket address which is not used now.

Kais-DkM:
@jhs7jhs Then what would be the correct operation when the port is already occupied?

jhs7jhs:
@Kais-DkM Just find another port to open.

Kais-DkM:
@jhs7jhs It seems we have to choose a range of ports for testing. Do you have any value in mind?

Kais-DkM:
I agree with that. It seems good enough if there's no problem.

PR #923, opened 2018-11-30
Change unit test of stratum to use unused ports

jhs7jhs:
Resolves: #457 
This design is inevitable, because if ranges are overlapping each other, then there will be a race condition while finding the port due to parallel testing, so each test case will be able to choose the same port.
If there's a better way, please let me know.


stratum/src/lib.rs
```diff
@@ -364,15 +364,30 @@ mod tests {
...
+    fn get_available_test_addr(start: u32, end: u32) -> SocketAddr {
+        let addr = (start..end)
+            .map(|port| SocketAddr::from_str(&format!("127.0.0.1:{}", port)).unwrap())
+            .find(|addr| match TcpListener::bind(addr) {
+                Ok(v) => {
+                    drop(v);
+                    true
+                }
+                Err(_) => false,
+            })
```
Kais-DkM: How about `TcpListener::bind(addr).is_ok()`?
I think socket will be automatically dropped after this statement.

jhs7jhs: This code is redundant. I'll fix this. I was trying to fix the port problem described above, because I think the port is still bound.

PR #933, opened 2018-12-04
Add signature to Tendermint's proposal message

majecty:
This PR depends on #891

### DIFF
--- a/core/src/consensus/tendermint/message.rs
+++ b/core/src/consensus/tendermint/message.rs
@@ -17,12 +17,13 @@
 use std::cmp;
 
 use ccrypto::blake256;
-use ckey::Signature;
+use ckey::{public_to_address, recover, Address, Signature};
 use primitives::{Bytes, H256};
 use rlp::{Decodable, DecoderError, Encodable, RlpStream, UntrustedRlp};
 
 use super::super::vote_collector::Message;
 use super::{BlockHash, Height, Step, View};
+use crate::error::Error;
 use crate::header::Header;
 
 /// Complete step of the consensus process.
@@ -81,7 +82,7 @@ const MESSAGE_ID_PROPOSAL_BLOCK: u8 = 0x02;
 #[derive(Debug, PartialEq)]
 pub enum TendermintMessage {
     ConsensusMessage(Bytes),
-    ProposalBlock(Bytes),
+    ProposalBlock(Signature, Bytes),
 }
 
 impl Encodable for TendermintMessage {
@@ -92,9 +93,10 @@ impl Encodable for TendermintMessage {
                 s.append(&MESSAGE_ID_CONSENSUS_MESSAGE);
                 s.append(message);
             }
-            TendermintMessage::ProposalBlock(bytes) => {
-                s.begin_list(2);
+            TendermintMessage::ProposalBlock(signature, bytes) => {
+                s.begin_list(3);
                 s.append(&MESSAGE_ID_PROPOSAL_BLOCK);
+                s.append(signature);
                 s.append(bytes);
             }
         }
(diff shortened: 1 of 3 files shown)


core/src/miner/miner.rs
```diff
@@ -537,40 +537,33 @@ impl Miner {
...
                 *self.next_mandatory_reseal.write() = Instant::now() + self.options.reseal_max_period;
-                {
-                    let mut sealing_work = self.sealing_work.lock();
-                    sealing_work.queue.push(block.clone());
-                    sealing_work.queue.use_last_ref();
+                if self.engine.is_proposal(block.header()) {
+                    block
+                        .lock()
+                        .seal(&*self.engine, seal.clone())
+                        .map(|sealed| {
```
```rust
.map(|sealed| {
    self.engine.proposal_generated(&sealed);
    self.engine.broadcast_proposal_block(sealed);
})
.map_err(|e| {
    cwarn!(MINER, "ERROR: seal failed when given internally generated seal: {}", e);
}).is_ok();
```

majecty:
@sgkim126 Could you review this PR.
You only have to check the last commit.

core/src/consensus/tendermint/message.rs
```diff
@@ -133,6 +143,23 @@ impl ConsensusMessage {
...
+        let height = header.number();
+        let view = consensus_view(header)?;
+        Ok(ConsensusMessage {
+            signature,
+            block_hash: Some(header.hash()),
+            vote_step: VoteStep::new(height as Height, view, Step::Propose),
+        })
+    }
+
+    pub fn verify(&self) -> Result<Address, Error> {
```
sgkim126: Returning `Address` from `verify` function looks weird. How about `verifier` or `signer`?

core/src/consensus/tendermint/message.rs
```diff
@@ -233,7 +260,7 @@ mod tests {
 
     #[test]
     fn encode_and_decode_tendermint_message_2() {
-        rlp_encode_and_decode_test!(TendermintMessage::ProposalBlock(vec![1u8, 2u8]));
+        rlp_encode_and_decode_test!(TendermintMessage::ProposalBlock(Default::default(), vec![1u8, 2u8]));
```
sgkim126: How about
```rust
rlp_encode_and_decode_test!(TendermintMessage::ProposalBlock(Signature::random(), vec![1u8, 2u8]));
```

core/src/consensus/tendermint/message.rs
majecty: OK. I'll use the `signer`.

core/src/consensus/tendermint/message.rs
majecty: Good!

majecty:
@sgkim126 I fixed them.

core/src/consensus/tendermint/mod.rs
```diff
@@ -998,13 +1013,59 @@ impl NetworkExtension for TendermintExtension {
...
+                // This block borrows bytes
+                {
+                    let block_view = BlockView::new(&bytes);
+                    let header_view = block_view.header();
+                    let number = header_view.number();
+                    ctrace!(ENGINE, "Proposal received for {}-{:?}", number, header_view.hash());
+
+                    let message = match ConsensusMessage::new_proposal(signature, &header_view) {
+                        Ok(message) => message,
+                        Err(_) => {
```
```rust
Err(err) => {
    cwarn!(ENGINE, "Invalid proposal received: {:?}", err);
    return
}
```

core/src/consensus/tendermint/mod.rs
```diff
@@ -998,13 +1013,59 @@ impl NetworkExtension for TendermintExtension {
...
+                            cwarn!(ENGINE, "Invalid proposal received");
+                            return
                         }
+                    };
+
+                    let author_from_signature = match message.signer() {
+                        Ok(address) => address,
+                        Err(_) => {
```
```rust
Err(err) => {
    cwarn!(ENGINE, "Proposal verification failed: {:?}", err);
    return
}
```

core/src/consensus/tendermint/mod.rs
```diff
@@ -998,13 +1013,59 @@ impl NetworkExtension for TendermintExtension {
...
+                            ENGINE,
+                            "Proposal author({}) not matched with header({})",
+                            author_from_signature,
+                            header_view.author()
+                        );
+                        return
+                    }
+
+                    if t.votes.is_old_or_known(&message) {
+                        cinfo!(ENGINE, "Proposal is already known");
```
sgkim126: I think it should be `cdebug`.

sync/src/block/extension.rs
```diff
@@ -321,7 +321,6 @@ impl ChainNotify for Extension {
         }
         let mut headers_to_download: Vec<_> = enacted
             .into_iter()
-            .chain(new_highest_header.into_iter())
```
sgkim126: Would you explain about this change?

majecty: The purpose of this PR is checking whether the proper proposer creates the proposal or not. Since, If a node downloads a proposal block using block sync, the node could not verify the signature(the block does not contain the signature), I made the block extension not to download the highest block which doesn't be included in the canonical chain(that is proposal block).

sgkim126: Is it safe also on other consensus engines?

PR #934, opened 2018-12-04
Import proposal block early

majecty:
This PR depends on #891, #933 

If a node would import a proposal block early, the node could propagate the previous block shortly, because the previous block became the best block when the proposal block is imported.

### DIFF
--- a/core/src/consensus/tendermint/mod.rs
+++ b/core/src/consensus/tendermint/mod.rs
@@ -51,7 +51,7 @@ use crate::consensus::EngineType;
 use crate::encoded;
 use crate::error::{BlockError, Error};
 use crate::header::Header;
-use crate::views::BlockView;
+use crate::views::{BlockView, HeaderView};
 use crate::BlockId;
 use ChainNotify;
 
@@ -785,6 +785,10 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
         c.block_header(&BlockId::Number(prev_height as BlockNumber))
             .expect("Previous height's block should be imported")
             .hash()
+    }
+
+    fn get_best_block_from_highest_score_header(&self, header: &HeaderView) -> H256 {
+        header.parent_hash()
     }
 }
 
--- a/core/src/consensus/tendermint/mod.rs
+++ b/core/src/consensus/tendermint/mod.rs
@@ -48,7 +48,6 @@ use crate::block::*;
 use crate::client::{Client, EngineClient};
 use crate::codechain_machine::CodeChainMachine;
 use crate::consensus::EngineType;
-use crate::encoded;
 use crate::error::{BlockError, Error};
 use crate::header::Header;
 use crate::views::{BlockView, HeaderView};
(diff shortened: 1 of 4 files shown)


majecty:
@sgkim126 Could you review this PR?
You only need to check last two commits.

sgkim126:
Don't we need to call `mark_as_good`?

core/src/consensus/tendermint/mod.rs
```diff
@@ -141,6 +140,34 @@ impl Tendermint {
...
+        if prev_score < new_score {
+            ctrace!(ENGINE, "future proposal changed");
+            *self.future_proposal.write() = Some(future_proposal);
+        } else {
+            ctrace!(ENGINE, "future proposal not changed");
+        }
+    }
+
+    pub fn future_proposal_prev_hash(&self) -> Option<H256> {
+        if let Some(future_proposal) = self.future_proposal.try_read_for(std::time::Duration::new(1, 0)) {
```
sgkim126: Why do you use `try_read_for`?
It should be used only when there is a fallback when it failed to acquire a lock.

core/src/consensus/tendermint/mod.rs
```diff
@@ -279,14 +293,38 @@ impl Tendermint {
...
+                        *self.proposal.write() = Some(*hash);
+                        *self.proposal_parent.write() = header.parent_hash();
+                    } else {
+                        ctrace!(ENGINE, "Proposal is received but not imported");
+                        // Proposal is received but is not verified yet.
+                        // Wait for verification.
+                        return
+                    }
+                } else {
+                    self.update_sealing()
```
sgkim126: When does `get_block_hashes` return None?

majecty:
@sgkim126 We don't need to call `mark_as_good`.
`mark_as_good` is used to remove the block/header from the verification queue.
If we import the block/header, it is removed from the verification queue while importing it.

core/src/consensus/tendermint/mod.rs
majecty: I'll fix it.

core/src/consensus/tendermint/mod.rs
majecty: `self.votes.get_block_hashes(&vote_step).first()` is true when a node received proposal message before it entered to the "propose" step.
If a node didn't receive the proposal before enter the propose step, the value is None.

majecty:
@sgkim126 I fixed them.

core/src/consensus/tendermint/mod.rs
sgkim126: Is it okay to call `update_sealing`?
Even if the node is not a proposer?
Or is it guaranteed that the node is a proposer when there is no received proposal?

sgkim126:
> mark_as_good

If then, why does `mark_as_good` exist?
Is there a case that a block should not be imported even it is good?

majecty:
Before this commit, the proposal block is not imported although verification was successful. Then it was needed to remove the proposal from the verification queue.

Now there is no case that a block is not imported although verification was successful.

core/src/consensus/tendermint/mod.rs
majecty: It has been like this before.

I agree with that we do not need to call update_sealing if the node is not a proposer.
Fixing this behavior should be another PR.

PR #935, opened 2018-12-05
Query to the consensus engine which block a node mine on

majecty:
Since CodeChain saves the highest block hash in the DB, the client does
not need to ask it to the engine.

This PR depends on #891

### DIFF
--- a/core/src/consensus/tendermint/mod.rs
+++ b/core/src/consensus/tendermint/mod.rs
@@ -777,6 +777,15 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
     fn register_chain_notify(&self, client: &Client) {
         client.add_notify(Arc::downgrade(&self.chain_notify) as Weak<ChainNotify>);
     }
+
+    fn get_block_hash_to_mine_on(&self, _best_block_hash: H256) -> H256 {
+        let c = self.client.read().as_ref().and_then(|weak| weak.upgrade()).expect("Client should be exist");
+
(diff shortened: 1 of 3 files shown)


majecty:
@sgkim126 I changed the code to query the block which mine on to the consensus engine.
Could you review again?

core/src/consensus/tendermint/mod.rs
```diff
@@ -762,6 +762,18 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
     fn register_chain_notify(&self, client: &Client) {
         client.add_notify(Arc::downgrade(&self.chain_notify) as Weak<ChainNotify>);
     }
+
+    fn get_block_hash_to_mine_on(&self, _best_block_hash: H256) -> H256 {
+        let c = match self.client.read().as_ref().and_then(|weak| weak.upgrade()) {
+            Some(c) => c,
+            None => return Default::default(),
```
sgkim126: When does it happen?
And is it okay just return zero hash?

majecty: This could happen when the codechain exits.
The program will be killed soon, so returning default is ok.

sgkim126: If then, how about using `expect` so it will be killed or returning `Option`?
I want to make the meaning clear.

majecty: Ok.

PR #960, opened 2018-12-10
Do not run integration tests parallel

sgkim126:


sgkim126:
It becomes flakier on Travis.

PR #975, opened 2018-12-13
 Skip to prevote step if all pre-commits are collected

majecty:


core/src/consensus/tendermint/mod.rs
```diff
@@ -435,6 +440,13 @@ impl Tendermint {
                 ctrace!(ENGINE, "Transition to {:?} triggered.", step);
                 self.to_step(step);
             }
+        } else if vote_step.step == Step::Precommit
+            && self.height() - 1 == vote_step.height
```
sgkim126: Is it okay with the genesis block?
The hight of the genesis block is 0.

jhs7jhs: `self.height()` has positive value always.

sgkim126: You are right. it's always a positive value.
I checked that it's the height to vote, not the current height.

sgkim126:
@majecty The change in the second commit is already applied to master.
https://github.com/CodeChain-io/codechain/commit/a9957f982b2ddf4f284eaad77546aacd4460d7c2#diff-1c1e4a7a7118105177e1800187ceca34R124

PR #977, opened 2018-12-14
Go to Prevote step when proposal block arrives/is created

jhs7jhs:
Should be merged after #975 
I'm not really sure about this logic because there are too many parts related to proposal.


core/src/consensus/tendermint/mod.rs
```diff
@@ -534,6 +544,7 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
             consensus_view(header).expect("This block is generated by this machine."),
             header
         );
+        self.to_step(Step::Prevote);
```
sgkim126: When is `proposal_generated` called?
Is it a function that called on the block proposer?
If then, I think you should describe it on the commit message.

core/src/consensus/tendermint/mod.rs
```diff
@@ -455,8 +456,17 @@ impl Tendermint {
         let height = proposal.number() as Height;
         let view = consensus_view(proposal).expect("Imported block is already verified");
         if current_height == height && self.view.load(AtomicOrdering::SeqCst) == view {
-            *self.proposal.write() = Some(proposal.hash());
-            *self.proposal_parent.write() = *proposal.parent_hash();
+            {
+                let mut proposal_guard = self.proposal.write();
+                if proposal_guard.is_none() {
```
sgkim126: How about lock `proposal_parent` here?

core/src/consensus/tendermint/mod.rs
```diff
@@ -455,8 +456,17 @@ impl Tendermint {
         let height = proposal.number() as Height;
         let view = consensus_view(proposal).expect("Imported block is already verified");
         if current_height == height && self.view.load(AtomicOrdering::SeqCst) == view {
-            *self.proposal.write() = Some(proposal.hash());
-            *self.proposal_parent.write() = *proposal.parent_hash();
+            {
+                let mut proposal_guard = self.proposal.write();
```
sgkim126: How about lock `proposal_parent` here?

core/src/consensus/tendermint/mod.rs
```diff
@@ -319,6 +319,7 @@ impl Tendermint {
                     if let Some(header) = c.block_header(&BlockId::Hash(*hash)) {
                         *self.proposal.write() = Some(*hash);
                         *self.proposal_parent.write() = header.parent_hash();
+                        self.to_step(Step::Prevote);
```
sgkim126: Why do we call it here?
We already change the current step in `on_imported_proposal` and `proposal_generated`.

jhs7jhs: It's also my question. There's a code `t.votes.vote(message, author_from_signature);` in `on_message()`. But I'm not sure `to_step(Step::Propose)` called at that time. I cannot find any relations.

jhs7jhs:
I'm sorry, I forgot to put WIP tag. There are too many functions which are called from outside, and even there are functions always called together. This code should be refactored.

core/src/consensus/tendermint/mod.rs
majecty: Since the proposal block could be received in the previous height's prevote or precommit step, we need the `to_step` here.

core/src/consensus/tendermint/mod.rs
sgkim126: It should be described on the commit message since it changes the behavior.
The further reader will expect this commit only changes the behavior when the proposal block arrives, but it also changes the behavior when the proposal block is generated.

jhs7jhs: Oh, I see. I wrote "when proposal block **arrives**". I'll fix this.
Also, I think there should be some comments for readers. It's too hard to figure out where those callbacks are called. Even I added less than 100 lines of code, it takes too much time :(

core/src/consensus/tendermint/mod.rs
jhs7jhs: Is there any reason to split `self.proposal` and `self.proposal_parent`? I don't think it would improve performance.

sgkim126: I agree.
Let's merge two locks in the other commit.

jhs7jhs: Okay.

PR #978, opened 2018-12-14
Add timeout token counter in Tendermint

jhs7jhs:
There could be a race condition between `step()` and functions which call `to_step()` except `step()`. `to_step()` can be called twice.

### DIFF
--- a/core/src/consensus/tendermint/mod.rs
+++ b/core/src/consensus/tendermint/mod.rs
@@ -98,6 +98,8 @@ pub struct Tendermint {
     chain_notify: Arc<TendermintChainNotify>,
     /// Lock for changing step. We use ReentrantMutex because handle_valid_message called recursively.
     pub step_change_lock: ReentrantMutex<()>,
+    /// The token counter for timeout. If timeout is called with the token which is less than this value, it should be discarded.
+    timeout_token_counter: AtomicUsize,
 }
 
 impl Tendermint {
@@ -125,6 +127,7 @@ impl Tendermint {
             chain_notify: Arc::new(chain_notify),
             machine,
             step_change_lock: ReentrantMutex::new(()),
+            timeout_token_counter: AtomicUsize::new(0),
         });
         engine.extension.register_tendermint(Arc::downgrade(&engine));
         engine.chain_notify.register_tendermint(Arc::downgrade(&engine));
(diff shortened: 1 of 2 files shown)


core/src/consensus/tendermint/mod.rs
```diff
@@ -295,7 +298,10 @@ impl Tendermint {
     }
 
     fn to_step(&self, step: Step) {
-        self.extension.set_timer_step(step);
+        // Should be protected by step_change_lock
+        let counter = self.timeout_token_counter.load(AtomicOrdering::SeqCst);
+        self.extension.set_timer_step(step, counter);
```
majecty: I think that we should pass the increased counter value.
The function's name is `set_timer_step`, so we need to pass the value to be set which is the increased counter.

core/src/consensus/tendermint/mod.rs
```diff
@@ -1032,8 +1042,8 @@ impl TendermintExtension {
             let view = t.view.load(AtomicOrdering::SeqCst);
             let du = Duration::seconds(view as i64);
 
-            api.clear_timer(ENGINE_TIMEOUT_TOKEN).expect("Timer clear succeeds");
-            api.set_timer_once(ENGINE_TIMEOUT_TOKEN, self.timeouts.timeout(&step) + du).expect("Timer set succeeds");
+            api.clear_timer(ENGINE_TIMEOUT_TOKEN + counter).expect("Timer clear succeeds");
+            api.set_timer_once(ENGINE_TIMEOUT_TOKEN + counter, self.timeouts.timeout(&step) + du).expect("Timer set succeeds");
```
majecty: You write code that `clear_timer` and `set_timer_once` receives the same counter value.
But should we clear the timer with the previous counter and set new timer with the new value?

Maybe the code should be like below.

```
api.clear_timer(ENGINE_TIMEOUT_TOKEN + counter - 1).expect("Timer clear succeeds");
api.set_timer_once(ENGINE_TIMEOUT_TOKEN + counter, self.timeouts.timeout(&step) + du).expect("Timer set succeeds");
```

jhs7jhs: Oh, I've missed this. I'll fix this.

core/src/consensus/tendermint/mod.rs
jhs7jhs: I think it's much better. I'll fix.

jhs7jhs:
@majecty That's exact scenario that I thought at that time. I think this situation is hard to find in normal, well-connected environment.

majecty:
LGTM

PR #986, opened 2018-12-19
Make each parcel types have different min costs

sgkim126:


sgkim126:
The cost value is not determined yet.

PR #987, opened 2018-12-20
Make key generation deterministic in tests

sgkim126:


jhs7jhs:
I think it's good to use `SmallRng` instead of `XorShiftRng`, because `SmallRng` is included in `rand`. (`rand::rngs::SmallRng`)

sgkim126:
@jhs7jhs 

https://docs.rs/rand/0.6.1/rand/rngs/struct.SmallRng.html
> Reproducibility of output from this generator is however not required, thus future library versions may use a different internal generator with different output. Further, this generator may not be portable and can produce different output depending on the architecture.

I want to make the test deterministic but `SmallRng` is not reproducible and platform dependent.

PR #991, opened 2018-12-20
Fix a bug in ChkMultiSig

sgkim126:
Currently, the public keys must be continuously matched with the signatures. But non-continuous public keys are also valid if the order is correct.


sgkim126:
@joojis I re-pushed since I had forgotten to run rustfmt.

PR #969, opened 2018-12-12
Save Tendermint's state to the DB

majecty:
Tendermint should save votes to the persistent DB not to send
DoubleVote. Save backup data when step is changed or the node sends a
vote.

PR #992, opened 2018-12-20
Cleanup tendermint code

majecty:
This PR depends on https://github.com/CodeChain-io/codechain/pull/969

Each commit is simple. Please check it commit by commit except for the first commit.


sgkim126:
@majecty Please resolve conflict.s

core/src/consensus/tendermint/mod.rs
```diff
@@ -880,13 +875,12 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
             return false
         }
 
-        let c = match self.client.read().as_ref().and_then(|weak| weak.upgrade()) {
```
sgkim126: It seems that it changes the behavior.
Is it okay?

core/src/consensus/tendermint/mod.rs
```diff
@@ -480,36 +478,33 @@ impl Tendermint {
     }
 
     fn backup(&self) {
-        if let Some(c) = self.client.read().as_ref().and_then(|weak| weak.upgrade()) {
```
sgkim126: It seems that it changes the behavior.
Is it okay?

core/src/consensus/tendermint/mod.rs
```diff
@@ -919,7 +919,6 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
         {
             self.signer.write().set(ap, address, password);
         }
-        self.to_step(Step::Propose);
```
sgkim126: If then, when it starts?

core/src/consensus/tendermint/mod.rs
majecty: Yes. The client should valid in all places.

core/src/consensus/tendermint/mod.rs
majecty: Yes. The client should be valid in all places.

core/src/consensus/tendermint/mod.rs
majecty: The tendermint starts when the timer expires.
NetworkExtension for TendermintExtension's `on_initialize` set timer and the timer calls `to_step` function.

sgkim126:
@majecty Would you extract the last commit to another PR. That is not cleanup code.

majecty:
@sgkim126 I removed it.

PR #993, opened 2018-12-20
Add chain_executeVM() RPC call

### LINKED ISSUE #965: Add executeVM RPC method
```
executeVM(script: [u8]) => bool
```
This method is needed when checking if the script is valid.

jhs7jhs:
Resolves #965.


rpc/src/v1/impls/chain.rs
```diff
@@ -252,4 +254,46 @@ where
         let transaction_type = ::std::result::Result::from(transaction).map_err(errors::core)?;
         Ok(self.client.execute_transaction(&transaction_type, sender_address).map_err(errors::core)?)
     }
+
+    fn execute_vm(&self, tx: Transaction, params: Vec<Vec<BytesArray>>, indices: Vec<usize>) -> Result<Vec<String>> {
```
sgkim126: How about implementing this function on `Client` instead of making RPC depend on `codechain_vm`.

sgkim126: I agree.

PR #996, opened 2018-12-21
Change tendermint vote message format

sgkim126:


core/src/consensus/tendermint/message.rs
```diff
@@ -127,20 +127,28 @@ impl Decodable for TendermintMessage {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Hash, Default, RlpDecodable, RlpEncodable)]
+pub struct VoteOn {
+    pub step: VoteStep,
```
majecty: It's a shame to have `step.step` because VoteStep has a field named `step`.
But I couldn't think a good alternative.

sgkim126: I agree.

PR #998, opened 2018-12-22
Reset Tendermint's proposal field when the round is increased

### LINKED ISSUE #983: Tendermint should reset the proposal variable when round is changed
Tendermint's `proposal` variable means current round's proposal block's hash.
However, this value is not set to none when the round is changed. So the variable has the old value which is different from the original intention.

Effect:
Tendermint's `generate_seal` check the `proposal` and do not create a proposal if it is a `some` value. The purpose of the code is that does not make redundant proposal block. However, if the `proposal` could be set from the previous round, proposer miss the block creation.

majecty:
Since the proposal field means current round's proposal hash, a node
does not create a new proposal if the value is set. However, the
proposal value was not reset when the round is changed. So there was a
bug which a node could not create a proposal if the node received a
proposal in the previous rounds.

This fixes #983


sgkim126:
LGTM. @jhs7jhs How about you?

PR #1001, opened 2018-12-24
Upgrade sdk and test helper for integration tests

sgkim126:


test/src/integration/chain.test.ts
```diff
@@ -399,9 +399,6 @@ describe("chain", function() {
         });
 
         const data = tx.toJSON();
-        data.data.output.lockScriptHash = `0x${
-            data.data.output.lockScriptHash
-        }`;
```
joojis: Was this a workaround for the bug of `Transaction.toJSON()`?

sgkim126: Yes. I think so.

PR #1003, opened 2018-12-24
Small fixes in Tendermint

majecty:


core/src/consensus/tendermint/mod.rs
```diff
@@ -595,6 +603,7 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
         let view = self.view();
 
         let (last_block_hash, last_block_view) = &*self.last_confirmed_view.read();
+        assert_eq!(last_block_hash, &parent.hash());
```
majecty: This makes the test fail.
Can we ignore assert_eq! in the test environment?

@sgkim126 @jhs7jhs @foriequal0  Do you have any idea?

sgkim126: It's possible to use `#[cfg(not(test))]` but I don't recommend.

majecty: I found that I can fix this problem by making test client to use the Tendermint test scheme.
I've fixed now.

core/src/consensus/tendermint/mod.rs
```diff
@@ -454,6 +453,16 @@ impl Tendermint {
     pub fn on_imported_proposal(&self, proposal: &Header) {
         let _guard = self.step_change_lock.lock();
 
+        self.votes.throw_out_old(&VoteStep {
```
sgkim126: Do we have a proposal for genesis block?
If not, I think it must do an early return if the proposal number is 0 and the below code should be `(proposal.number() - 1) as usize`

majecty: No, we don't have a proposal for genesis block.
OK. I"ll fix it.

majecty:
@sgkim126 I fixed it.

PR #1005, opened 2018-12-26
Rename to_step to on_timeout in the Tendermint

majecty:


core/src/consensus/mod.rs
```diff
@@ -207,7 +207,7 @@ pub trait ConsensusEngine<M: Machine>: Sync + Send {
     fn populate_from_parent(&self, _header: &mut M::Header, _parent: &M::Header) {}
 
     /// Trigger next step of the consensus engine.
```
sgkim126: Would you fix the comment too?

majecty: I fixed it. Could you recheck it?

PR #1008, opened 2018-12-27
Fix the bug that causes inconsistent multiple reads of votes

majecty:
There was a crash which caused by the inconsistent result of
has_enough_aligned_votes.

ex)
```rust
if block_hash.is_none() && has_enough_aligned_votes() {
  // ...
} else if has_enough_aligned_votes() {
  // The below assertion could fail.
  // has_enough_aligned_votes could return different value with the
  // previous call
  assert!(block_hash.is_some())
}
```

### DIFF
--- a/core/src/consensus/tendermint/mod.rs
+++ b/core/src/consensus/tendermint/mod.rs
@@ -400,11 +400,12 @@ impl Tendermint {
             Some(lock) => *vote_step > lock.on.step,
             None => true,
         };
+        let has_enough_aligned_votes = self.has_enough_aligned_votes(message);
         let lock_change = is_newer_than_lock
             && vote_step.height == self.height()
             && vote_step.step == Step::Prevote
             && message.on.block_hash.is_some()
-            && self.has_enough_aligned_votes(message);
+            && has_enough_aligned_votes;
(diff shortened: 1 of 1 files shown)


core/src/consensus/tendermint/mod.rs
```diff
@@ -400,23 +400,24 @@ impl Tendermint {
             Some(lock) => *vote_step > lock.on.step,
             None => true,
         };
+        let has_enough_aligned_votes = self.has_enough_aligned_votes(message);
         let lock_change = is_newer_than_lock
             && vote_step.height == self.height()
             && vote_step.step == Step::Prevote
             && message.on.block_hash.is_some()
-            && self.has_enough_aligned_votes(message);
+            && has_enough_aligned_votes;
```
sgkim126: It changes the order of function calls.
Is it okay?

jhs7jhs: It does not cause a deadlock, so I think it's okay.

sgkim126: I was also worried about inconsistent data.
But it seems okay since other conditions are related to the message rather than the states of the extension.

PR #1013, opened 2018-12-28
Change signer public to signer index on Tendermint messages

jhs7jhs:
`VoteCollector` now stores index(`usize`) instead of `Public`.
`EngineError::NotAuthorized` is split to two: `BlockNotAuthorized` and `MessageNotAuthorized`. Because messages only have the index of the signer while blocks has the author field.

I'm concerned about two points:
1. Is it good to put the indices of the signers of the precommits on the block? This will improve the speed of verification because it'll use `verify` instead of `recover`.
2. Is `validators` are thread-safe? (Search for `self.validators` and `t.validators`.)

### DIFF
--- a/core/src/consensus/mod.rs
+++ b/core/src/consensus/mod.rs
@@ -45,7 +45,7 @@ use ctypes::util::unexpected::{Mismatch, OutOfBounds};
 use primitives::{Bytes, H256, U256};
 
 use self::epoch::{EpochVerifier, NoOp, PendingTransition};
-use self::tendermint::types::View;
+use self::tendermint::types::{BitSet, View};
 use crate::account_provider::AccountProvider;
 use crate::block::SealedBlock;
 use crate::codechain_machine::CodeChainMachine;
@@ -64,6 +64,7 @@ pub enum Seal {
         prev_view: View,
         cur_view: View,
         precommits: Vec<SchnorrSignature>,
+        precommit_bitset: BitSet,
     },
     None,
 }
@@ -78,10 +79,12 @@ impl Seal {
                 prev_view,
                 cur_view,
                 precommits,
+                precommit_bitset,
             } => Some(vec![
                 ::rlp::encode(prev_view).into_vec(),
                 ::rlp::encode(cur_view).into_vec(),
                 ::rlp::encode_list(precommits).into_vec(),
+                ::rlp::encode(precommit_bitset).into_vec(),
             ]),
         }
     }
@@ -256,6 +259,10 @@ pub trait ConsensusEngine<M: Machine>: Sync + Send {
         None
     }
 
+    fn signer_index(&self, _bh: &H256) -> Option<usize> {
+        None
+    }
+
     fn register_network_extension_to_service(&self, _: &NetworkService) {}
 
     fn score_to_target(&self, _score: &U256) -> U256 {
@@ -324,8 +331,19 @@ pub type PendingTransitionStore<'a> = Fn(H256) -> Option<PendingTransition> + 'a
 /// Voting errors.
 #[derive(Debug)]
 pub enum EngineError {
-    /// Signature or author field does not belong to an authority.
-    NotAuthorized(Address),
+    /// Precommit signatures or author field does not belong to an authority.
+    BlockNotAuthorized(Address),
+    /// The signature cannot be verified with the signer of the message.
+    MessageWithInvalidSignature {
+        height: u64,
+        signer_index: usize,
+        address: Address,
+    },
+    /// The validator on the given height and index is from the future or does not exist.
+    ValidatorNotExist {
+        height: u64,
+        index: usize,
+    },
     /// The same author issued different votes at the same step.
     DoubleVote(Address),
     /// The received block is from an incorrect proposer.
@@ -342,7 +360,19 @@ impl fmt::Display for EngineError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         use self::EngineError::*;
         let msg = match self {
-            NotAuthorized(address) => format!("Signer {} is not authorized.", address),
+            BlockNotAuthorized(address) => format!("Signer {} is not authorized.", address),
+            MessageWithInvalidSignature {
+                height,
+                signer_index,
+                address,
+            } => format!("The {}th validator({}) on height {} is not authorized.", signer_index, address, height),
+            ValidatorNotExist {
+                height,
+                index,
+            } => format!(
+                "The {}th validator on height {} does not exist. (out of bound or from the future)",
+                index, height
+            ),
             DoubleVote(address) => format!("Author {} issued too many blocks.", address),
             NotProposer(mis) => format!("Author is not a current proposer: {}", mis),
             UnexpectedMessage => "This Engine should not be fed messages.".into(),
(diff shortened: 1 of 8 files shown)


sgkim126:
1. I think the block should not use the index since there is no validator set on PoW.
2. `ValidatorSet` is marked as `Send + Sync`. It should be thread-safe. If it's not, it's a bug.

core/src/consensus/tendermint/mod.rs
```diff
@@ -680,24 +680,35 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
...
                 },
             };
-            let address = match self.votes.get(&precommit) {
-                Some(a) => a,
-                None => public_to_address(&recover(&precommit.signature, &precommit_hash)?),
+            let index = match self.votes.get_index(&precommit) {
+                Some(idx) => idx,
+                None => {
+                    let public = recover(&precommit.signature, &precommit_hash)?;
+                    match self.validators.get_index(header.parent_hash(), &public) {
```
```rust
self.validators.get_index(header.parent_hash(), &public).ok_or_else(|| {
    let address = public_to_address(&public);
    Err(EngineError::BlockNotAuthorized(address.to_owned()))
})?;

File: core/src/consensus/mod.rs [hunk 2]
```diff
@@ -324,8 +328,10 @@ pub type PendingTransitionStore<'a> = Fn(H256) -> Option<PendingTransition> + 'a
 /// Voting errors.
 #[derive(Debug)]
 pub enum EngineError {
-    /// Signature or author field does not belong to an authority.
-    NotAuthorized(Address),
+    /// Precommit signatures or author field does not belong to an authority.
+    BlockNotAuthorized(Address),
+    /// The height or the signer index of message does not belong to an authority.
+    MessageNotAuthorized(u64, usize),
```
sgkim126: How about adding `Option<Address>`?

jhs7jhs:
@sgkim126 For question 1, there's `Seal::Tendermint` which is for Tendermint:
```
Seal::Tendermint {
    prev_view: View,
    cur_view: View,
    precommits: Vec<Signature>,
}
```
It might be able to put indices. I'm not sure it's okay to do so.

core/src/consensus/tendermint/mod.rs
jhs7jhs: Fixed.

jhs7jhs:
I'm going to use `BitSet` in #1004 for add indices in Tendermint sealing.

majecty:
I also think it will be better to use the same BitSet struct.

core/src/consensus/mod.rs
jhs7jhs: I've added `Option<Address>` on `MessageNotAuthorized`, and changed an error message.

sgkim126:
@jhs7jhs BitSet was added.

core/src/consensus/mod.rs
jhs7jhs: Spilt into `MessageNotAuthorized` and `ValidatorNotExist`.

core/src/consensus/mod.rs
```diff
@@ -324,8 +331,12 @@ pub type PendingTransitionStore<'a> = Fn(H256) -> Option<PendingTransition> + 'a
 /// Voting errors.
 #[derive(Debug)]
 pub enum EngineError {
-    /// Signature or author field does not belong to an authority.
-    NotAuthorized(Address),
+    /// Precommit signatures or author field does not belong to an authority.
+    BlockNotAuthorized(Address),
+    /// The height or the signer index of message does not belong to an authority.
+    MessageNotAuthorized(u64, usize, Address),
```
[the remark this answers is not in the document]
jhs7jhs: It doesn't matter, just I've followed code nearby. I'll fix this.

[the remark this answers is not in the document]
jhs7jhs: Fixed.

core/src/consensus/mod.rs
```diff
@@ -324,8 +331,12 @@ pub type PendingTransitionStore<'a> = Fn(H256) -> Option<PendingTransition> + 'a
...
+    /// The validator on the given height and index is from the future or does not exist.
+    ValidatorNotExist(u64, usize),
```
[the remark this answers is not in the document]
jhs7jhs: Fixed.

core/src/consensus/tendermint/mod.rs
```diff
@@ -726,35 +720,53 @@ impl TendermintInner {
...
+                );
+                return Err(BlockError::InvalidSeal.into())
+            }
+            while !precommit_bitset.is_set(bitset_index) {
+                bitset_index += 1;
+            }
+            precommit_bitset.reset(bitset_index);
+
+            // This message is a dummy message to check address. signer_index value does not affect.
             let precommit = ConsensusMessage {
```
sgkim126: It seems that only signature is used.
How about replacing it to `let signature = rlp.as_val();`

sgkim126: If type inference doesn't work,
```rust
let signature: SchnorrSignature = rlp.as_val()?;
```

core/src/consensus/vote_collector.rs
```diff
@@ -69,31 +69,31 @@ impl<M: Message> StepCollector<M> {
...
+                    author_index: message.signer_index(),
                     vote_one: previous,
                     vote_two: message,
                 })
             } else {
                 self.block_votes
                     .entry(message.block_hash())
-                    .or_insert_with(HashMap::new)
-                    .insert(message.signature(), message.signer_public());
+                    .or_insert_with(BTreeMap::new)
```
sgkim126: You can use [or_default](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_default)

core/src/consensus/tendermint/types.rs
```diff
@@ -111,6 +119,13 @@ impl BitSet {
 
         self.0[array_index] |= 1u8 << bit_index;
     }
+
+    pub fn reset(&mut self, index: usize) {
+        let array_index = index / 8;
+        let bit_index = index % 8;
+
+        self.0[array_index] &= 255u8 ^ (1 << bit_index);
```
sgkim126: `0b1111_1111` or `0xFF`

core/src/consensus/tendermint/mod.rs
jhs7jhs: I forgot to remove this. I'll fix this. Type inference should work as I think.

core/src/consensus/tendermint/types.rs
jhs7jhs: Changed to `0b1111_1111`.

core/src/consensus/tendermint/mod.rs
jhs7jhs: Fixed.

jhs7jhs:
By the way, I've checked this in local environment and 3-node network.

core/src/consensus/mod.rs
jhs7jhs: `MessageNotAuthorized` is changed to `MessageWithInvalidSignature`.

sgkim126:
@majecty how about you?

jhs7jhs:
@sgkim126 Actually, @majecty asked me for changing the error name.
Also sorry for dismissing again, there was a bug in checking precommit signatures.

PR #1014, opened 2018-12-28
Use Schnorr signature scheme in consensus algorithms

jhs7jhs:
Tested in local.


jhs7jhs:
@sgkim126 Tested. It works well!

PR #1016, opened 2018-12-28
Use coarse-grained lock in the Tendermint

majecty:
This PR depends on #1004.
Only the last commit is this PR's work.


sgkim126:
LGTM. Would you resolve the conflict?

PR #1018, opened 2018-12-28
Add metric logger

majecty:
The metric logger prints logs/sec at every second.

I think that we should turn on/off using some environment variable.

The sample result.
```
Metric at : 2018-12-28 21:50:10
Metric ckey::ecdsa::recover: cur 36385 changed 0 speed 0
Metric ckey::ecdsa::sign: cur 6 changed 0 speed 0
Metric ckey::schnorr::sign: cur 109 changed 0 speed 0
Metric ckey::schnorr::verify: cur 106 changed 0 speed 0
Metric mem_pool::update_future::all_seqs_from_sender: cur 73562040 changed 1406636 speed
1406636
Metric miner::add_parcels_to_pool: cur 10001 changed 0 speed 0
Metric miner::local_parcel::mark_invalid: cur 2738 changed 191 speed 191
Metric miner::prepare_block::parcel::push: cur 40001 changed 0 speed 0
Metric miner::prepare_block::parcel::remove_invalid: cur 2738 changed 191 speed 191
Metric parcel::recover_public: cur 36385 changed 0 speed 0
Metric parcel::recover_public::from_signed_parcel: cur 36385 changed 0 speed 0
Metric tendermint::on_message: cur 1460 changed 0 speed 0
Metric tendermint::request_messages: cur 239 changed 0 speed 0
Metric tendermint::send_message: cur 493 changed 0 speed 0
Metric tendermint::send_proposal_block: cur 104 changed 0 speed 0
Metric tendermint::send_state: cur 636 changed 0 speed 0
```

### DIFF
--- a/util/logger/src/lib.rs
+++ b/util/logger/src/lib.rs
@@ -27,9 +27,11 @@ extern crate time;
 
 mod logger;
 mod macros;
+mod metric;
 mod structured_logger;
 
 use log::SetLoggerError;
+use metric::Metric;
 
 pub use logger::Config as LoggerConfig;
 use logger::Logger;
@@ -48,4 +50,8 @@ use lazy_static::lazy_static;
 
 lazy_static! {
     pub static ref slogger: StructuredLogger = StructuredLogger::create();
+}
+
+lazy_static! {
+    pub static ref metric_logger: Metric = Metric::new();
 }
--- a/util/logger/src/metric.rs
+++ b/util/logger/src/metric.rs
@@ -1,0 +1,122 @@
+// Copyright 2018 Kodebox, Inc.
+// This file is part of CodeChain.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use std::collections::BTreeMap;
+use std::sync::Arc;
+use std::thread;
+use std::time::Instant;
+
+use parking_lot::RwLock;
+use time;
+
+pub type MetricKey = &'static str;
+
+pub struct MetricEntry {
+    pub value: u64,
+    pub prev: u64,
+}
+
+impl MetricEntry {
+    fn new_with(initial: u64) -> Self {
+        MetricEntry {
+            value: initial,
+            prev: 0,
+        }
+    }
+
+    fn changed(&self) -> u64 {
+        self.value - self.prev
+    }
+
+    fn inc(&mut self) {
+        self.value += 1;
+    }
+
+    fn reset_prev(&mut self) {
+        self.prev = self.value;
+    }
+}
+
+pub struct Metric {
+    inner: Arc<RwLock<MetricInner>>,
+}
+
+pub struct MetricInner {
+    table: BTreeMap<MetricKey, MetricEntry>,
+    prev_print_time: Instant,
+}
+
+impl Default for Metric {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Metric {
+    pub fn new() -> Self {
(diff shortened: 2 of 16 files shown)


util/logger/src/metric.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use std::collections::BTreeMap;
+use std::sync::Arc;
+use std::thread;
+use std::time::Instant;
+
+use parking_lot::RwLock;
+use time;
+
+pub type MetricKey = String;
```
sgkim126: Please try to use `&'static str`

util/logger/src/metric.rs
```diff
@@ -0,0 +1,122 @@
...
+        self.value += 1;
+    }
+
+    fn reset_prev(&mut self) {
+        self.prev = self.value;
+    }
+}
+
+pub struct Metric {
+    inner: Arc<RwLock<MetricInner>>,
```
sgkim126: Make `Metric` have `RwLock<BtreeMap<MetricKey, AtomicUsize>>`.

majecty: Since `start_thread(&self)` function needs to clone `inner` to use it in the other thread, I added the `Arc`. Could you know any other method to use `inner` or `self` in other thread?

util/logger/src/metric.rs
majecty: Good! I fixed them

util/logger/src/metric.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+impl MetricInner {
+    pub fn increase(&mut self, key: &'static str) {
+        self.table.entry(key).or_insert_with(|| MetricEntry::new_with(1)).inc();
+    }
+
+    pub fn print(&mut self) {
+        let timestamp = time::strftime("%Y-%m-%d %H:%M:%S %Z", &time::now()).unwrap();
+        println!("Metric at : {}", timestamp);
```
sgkim126: Should it use `println` instead of log macros?

util/logger/src/metric.rs
sgkim126: What I commented was making `increase` function hold read lock to make multiple threads can increase the count.
So I think `Arc<RwLock<...>>` is okay.

util/logger/src/metric.rs
```diff
@@ -0,0 +1,122 @@
...
+            })),
+        }
+    }
+
+    pub fn start_thread(&self) {
+        let inner = Arc::clone(&self.inner);
+        thread::Builder::new()
+            .name("Metric logger".to_string())
+            .spawn(move || loop {
+                thread::sleep(::std::time::Duration::new(1, 0));
```
sgkim126: Use from_secs.
https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_secs

util/logger/src/lib.rs
```diff
@@ -49,3 +51,7 @@ use lazy_static::lazy_static;
 lazy_static! {
     pub static ref slogger: StructuredLogger = StructuredLogger::create();
 }
+
+lazy_static! {
+    pub static ref metric_logger: Metric = Metric::new();
```
sgkim126: How about passing `Metric` to the modules which use it instead of declaring global one.

util/logger/src/metric.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub fn new() -> Self {
+        Self {
+            inner: Arc::new(RwLock::new(MetricInner {
+                table: BTreeMap::new(),
+                prev_print_time: Instant::now(),
+            })),
+        }
+    }
+
+    pub fn start_thread(&self) {
```
sgkim126: How about renaming it to `start` and making it return `Self`.
Current code seems fragile because this function must be called only once.

util/logger/src/lib.rs
majecty: Are you saying about passing the metric variable as a function argument?
Then calculating the metric of very nested code became very difficult.

util/logger/src/metric.rs
majecty: Thanks!

util/logger/src/metric.rs
majecty: Then which log level it should be? info?

util/logger/src/metric.rs
majecty: Aha, I understand it. It would be great.

util/logger/src/metric.rs
majecty: Good!

majecty:
It's old and we don't need metric log now.

PR #1021, opened 2019-01-02
Refactor the mem pool code

jhs7jhs:
1. `update_future` was not efficient enough, so tried to reduce the time. But it is not possible to reduce the time complexity because of `by_signer_public`. Lazy evaluation is possible, but I'm not sure how it'll be efficient in here.
2. There were some redundant expect statements, so I've deleted.


jhs7jhs:
There's a problem with the first commit. I'm figuring out now.

PR #1022, opened 2019-01-02
Show the existing account information in case of failure

HoOngEe:
resolves: #795

### DIFF
--- a/keystore/src/keystore.rs
+++ b/keystore/src/keystore.rs
@@ -67,7 +67,7 @@ impl SimpleSecretStore for KeyStore {
     fn insert_account(&self, secret: Secret, password: &Password) -> Result<Address, Error> {
         let keypair = KeyPair::from_private(secret.into()).map_err(|_| Error::CreationFailed)?;
         if self.has_account(&keypair.address())? {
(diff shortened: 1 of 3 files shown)


keystore/src/keystore.rs
```diff
@@ -67,7 +67,7 @@ impl SimpleSecretStore for KeyStore {
     fn insert_account(&self, secret: Secret, password: &Password) -> Result<Address, Error> {
         let keypair = KeyPair::from_private(secret.into()).map_err(|_| Error::CreationFailed)?;
         if self.has_account(&keypair.address())? {
-            Err(Error::AlreadyExists)
+            Err(Error::AlreadyExists(keypair.address()))
```
sgkim126: `Address` is not suitable.
The error message is printed to users but user knows only `PlatformAddress` not `Address`.
Please find the way to make `PlatformAddress`.

jhs7jhs:
I think the keystore should not know which network is used when it's accessed. I recommend to split error types.

majecty:
@HoOngEe If we don't have an easy way to get the network id, how about closing this PR?

PR #1023, opened 2019-01-02
Make scheme be able to export Arc<Tendermint>

majecty:


sgkim126:
Would you explain why it is needed?

majecty:
This PR was for Tendermint's unit test.

As a result of offline communication, since this PR seems to change structure too much for the test, seulgi and I decided that to get Tendermint struct without changing other structure.

If it is too hard, I'll try to test the Tendermint in an integration test.

core/src/scheme/scheme.rs
```diff
@@ -295,7 +295,24 @@ impl Scheme {
...
+            load_from(x, engine).map_err(fmt_err)
+        })
+    }
+
+    pub fn load_tendermint<R>(reader: R) -> Result<(Self, Arc<Tendermint>), String>
+    where
+        R: Read, {
+        cjson::scheme::Scheme::load(reader).map_err(fmt_err).and_then(|x| {
+            let machine = Self::machine(&x.engine, CommonParams::from(x.params.clone()));
+            let tendermint = match x.engine.clone() {
```
foriequal0: I think we can most of the `#[derive(..., Clone)]` making
```rust
match &x.engine {
  cjson::scheme::Engine::Tendermint(tendermint) => Tendermint::new(tendermint.params.clone().into(), machine),
}
```

PR #1028, opened 2019-01-03
Reduce last_confirmed_view lock range

majecty:
When using RwLock like below, if the some_reference_value is a reference
type, the value holds read lock.
let some_reference_value = last_confirmed_view().read().blahblah();

This commit reduce unintended extended lock range.


sgkim126:
How about removing all internal locks instead of it?

majecty:
Good! I'll close this.

PR #1029, opened 2019-01-03
Update Parcel documentation

sgkim126:
0. No Parcel anymore. It's Transaction.
1. The nonce is renamed to seq.
2. No distinguish between transaction and asset transaction.
3. Timelock/Expiration of the Transaction is removed.
4. The type of the seq is changed.
5. The type of the fee is changed.
6. Add approvals field to some transactions.
7. Add some new transaction types.
8. Call the fee payer as payer instead of sender.
9. Registrar becomes approver.


sgkim126:
Because it doesn't change the code, there is no need to wait.
I'll merge it.

PR #1033, opened 2019-01-03
Do not use RESEAL_MIN_TIMER in the Tendermint 

majecty:


core/src/miner/miner.rs
```diff
@@ -578,7 +578,9 @@ impl MinerService for Miner {
     fn set_author(&self, address: Address, password: Option<Password>) -> Result<(), SignError> {
         self.params.write().author = address;
 
-        if self.engine_type() == EngineType::InternalSealing && self.engine.seals_internally().is_some() {
+        let engine_type = self.engine_type();
+        let internal_sealing_engine = engine_type == EngineType::PoA || engine_type == EngineType::PBFT;
+        if internal_sealing_engine && self.engine.seals_internally().is_some() {
```
sgkim126: How about adding a method `seals_internally()` to EngineType?
```rust
if self.engine_type().seals_internally() && self.engine.seals_internally().is_some() {
```
And it's more readable if the next patch adds `EngineType::use_reseal_min_timer()`.

majecty: Good!

majecty: `engine_type().seals_internally` seems confusing.
`seals_internally` has some meanings which slightly different in code.

1. The engine create block seal without external calculation. (PBFT, SimplePoA, Solo)
2. The engine needs signer's private key to seal. (PBFT, SimplePoA)

So I'll add `EngineType::need_signer_key` instead of `engine_type().seals_internally`

core/src/client/client.rs
```diff
@@ -262,7 +262,7 @@ impl TimeoutHandler for Client {
             }
             RESEAL_MIN_TIMER_TOKEN => {
                 // Checking self.ready_parcels() for efficiency
-                if !self.ready_parcels().is_empty() {
+                if self.engine().engine_type() != EngineType::PBFT && !self.ready_parcels().is_empty() {
```
sgkim126: Please add a method `use_reseal_min_timer` or `ignore_reseal_min_timer`

majecty: I fixed it.

PR #1041, opened 2019-01-03
Trigger RESEAL_MIN_TIMER when block is imported

majecty:


test/src/helper/promise.ts
```diff
@@ -17,3 +17,36 @@
...
+        const result = await p;
+
+        this.waiting[key] -= 1;
+        if (this.waiting[key] === 0) {
+            delete this.waiting[key];
+        }
+        return result;
+    }
+
+    public checkFulfil() {
```
sgkim126: What does this method?
It just prints log and returns nothing?

majecty: @sgkim126 
Hyunsik is right. When running the Tendermint tests, there were many test failures caused by connection error. If we could check the reason for the timeout, we could save a lot of time.

jhs7jhs:
@sgkim126 I'm not sure about triggering the reseal timer in `chain_new_blocks()`. When the transactions of the new block is removed from the mempool? I think we need to trigger the timer when there are some transactions moved from the future pool to the current pool.

sgkim126:
@jhs7jhs AFAIK, they are removed on `chain_new_blocks`. `chain_new_blocks` is called right before calling `Chain::new_blocks()` in `import_sealed_block` and `import_verified_block`.
The future transactions are moved in `remove_old` method, but I'm not sure whether there is no other code path that moves future transactions.

test/src/helper/promise.ts
sgkim126: How about making `checkFulfilled` throw an error if there is a waiting promise?

sgkim126: And please use template string instead of string concatenation.

test/src/helper/promise.ts
```diff
@@ -17,3 +17,46 @@
...
+        }
+        this.waiting = {};
+
+        if (timeoutJobs.length > 0) {
+            console.error(
+                `Timeout period is expired while waiting ${timeoutJobs.join(
+                    ", "
+                )}`
+            );
+            throw new Error("");
```
sgkim126: Why do you throw an empty string?
How about throw
```js
`Timeout period is expired while waiting ${timeoutJobs.join(
                    ", "
                )}`
```

majecty: I fixed it.

PR #1043, opened 2019-01-04
[WIP] Define ParcelPool in MemPool

jhs7jhs:
Moved current/future of `MemPool` to `ParcelPool`.
This is an initial job for refactoring on the memory pool. The current/future parcel sets are not capsulized well, so there are many lines which make readers confused. I think it's possible to move them to one struct.


jhs7jhs:
I think it's hard to split for now. I'll reopen this if it is good to go.

PR #1060, opened 2019-01-07
Update copyright

majecty:


kseo:
You don't need to update copyrights all at once.

majecty:
@kseo 
Then, how do we change the copyright?

kseo:
@majecty Just update the copyright of the files you modified.

sgkim126:
How about making this PR update the files that have been modified since January 1, and then change others individually?

jhs7jhs:
Moved to #1139.

PR #1061, opened 2019-01-07
[WIP] Creates a block in a dedicated thread

majecty:
Currently, the Tendermint calls `update_sealing` function which creates a
block. The `update_sealing` function is the cause of the deadklock because
it acquires both Tendermint lock and importer lock.

Using a dedicated thread will fix the deadlock, and also the dedicated
thread approach is more natural than concurrently creating a block.


core/src/client/client.rs
```diff
@@ -694,17 +697,17 @@ impl ReopenBlock for Client {
...
 
 impl PrepareOpenBlock for Client {
-    fn prepare_open_block(&self, author: Address, extra_data: Bytes) -> OpenBlock {
+    fn prepare_open_block(&self, parent_block: BlockId, author: Address, extra_data: Bytes) -> OpenBlock {
         let engine = &*self.engine;
         let chain = self.block_chain();
-        let h = engine.get_block_hash_to_mine_on(chain.best_block_hash());
-        let latest_header = &chain.block_header(&h).expect("h is best block hash: so its header must exist: qed");
+        let h = self.block_hash(&parent_block).expect("parent exist always");
+        let parent_header = chain.block_header(&h).expect("parent exist always");
```
```rust
assert_eq!(h, parent_header.hash());
```

core/src/consensus/tendermint/types.rs
```diff
@@ -26,6 +26,42 @@ use super::message::VoteStep;
...
+    ProposeWaitBlockGeneration {
+        parent_hash: H256,
+    },
+    Prevote,
+    Precommit,
+    Commit,
+}
+
+impl InternalStepState {
+    pub fn to_step(&self) -> Step {
```
sgkim126: Please implement `From<InternalStepState> for Step`

core/src/consensus/tendermint/types.rs
```diff
@@ -26,6 +26,42 @@ use super::message::VoteStep;
...
+                ..
+            } => Step::Propose,
+            InternalStepState::Prevote => Step::Prevote,
+            InternalStepState::Precommit => Step::Precommit,
+            InternalStepState::Commit => Step::Commit,
+        }
+    }
+}
+
+impl From<&Step> for InternalStepState {
```
sgkim126:
```rust
impl From<Step> for InternalStepState 
```

core/src/consensus/tendermint/types.rs
```diff
@@ -26,6 +26,42 @@ use super::message::VoteStep;
 pub type Height = usize;
 pub type View = usize;
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
+pub enum InternalStepState {
```
sgkim126: How about just `TendermintState`?

core/src/client/client.rs
majecty: Do we need the assert?
`chain.block_header` has the responsibility to satisfy the assertion.

core/src/consensus/tendermint/types.rs
majecty: I use this `to_step` function to use reference of `InternalStepState`.
If we need the `InternalStepState` value after converting it to `Step`, should we copy the `InternalStepState` value?

core/src/consensus/tendermint/types.rs
majecty: Good!

core/src/consensus/tendermint/types.rs
sgkim126: I thought From is better since the copying cost of `InternalStepState` is cheap.
But now, implementing `to_step` also looks good.

core/src/client/client.rs
sgkim126: I think adding `debug_assert_eq` and using `h` at line 706 is better.

majecty:
@sgkim126 I changed the code to use the pre-calculated header's hash in the `prepare_open_block` function.

core/src/consensus/tendermint/types.rs
```diff
@@ -71,6 +107,21 @@ impl Encodable for Step {
     }
 }
 
+impl From<TendermintState> for Step {
+    fn from(s: TendermintState) -> Self {
+        match s {
```
sgkim126: How about `s.to_step()`?

majecty: Good

majecty:
@sgkim126 
I made the miner not to create a block although new transactions are received when the engine is PBFT.

PR #1063, opened 2019-01-07
Make IoService handle only one Handler

sgkim126:
Current implementation allows multiple Handler be registered. But
CodeChain needs only one handler. So this patch removes multiple handler
feature.

### DIFF
--- a/util/io/src/service.rs
+++ b/util/io/src/service.rs
@@ -15,7 +15,6 @@
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
 use crossbeam::deque;
-use limited_table::LimitedTable;
 use mio::deprecated::{EventLoop, EventLoopBuilder, Handler, Sender};
 use mio::timer::Timeout;
 use mio::*;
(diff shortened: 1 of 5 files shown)


sgkim126:
FYI, the current implementation is just verbose and not useful because the handlers registered in the same IoService should receive the same message type.

util/io/src/service.rs
```diff
@@ -569,12 +472,9 @@ where
     /// Regiter an IO handler with the event loop.
     pub fn register_handler(&self, handler: Arc<IoHandler<Message> + Send>) -> Result<(), IoError> {
         let h = Arc::clone(&handler);
-        let handler_id =
-            self.handlers.write().insert(handler).unwrap_or_else(|| panic!("Too many handlers registered"));
-        h.initialize(&IoContext::new(
-            IoChannel::new(self.event_loop_channel.clone(), Arc::downgrade(&self.handlers)),
-            handler_id,
-        ))?;
+        assert!(self.handler.read().is_none());
```
majecty: Good assertion!

PR #1070, opened 2019-01-09
Enhance Tendermint unit tests

majecty:
* I added an integration test which creating a few non-empty blocks.
* I used `PromiseExpect` to log which async job has caused a timeout error.

When I run the integration test 10 times, the test failed two times because of connection error.


foriequal0:
The new test seems to be not deterministic. But it would be laborious, so merge this for now, and fix it when it is failing too often.

PR #1071, opened 2019-01-09
Update mem usage

HoOngEe:
update mem_usage calculation with stack consideration using mem::size_of


core/src/miner/mem_pool.rs
```diff
@@ -115,11 +116,11 @@ impl TransactionOrder {
     fn for_transaction(item: &MemPoolItem, seq_seq: u64) -> Self {
         let rlp_bytes_len = rlp::encode(&item.tx).to_vec().len();
         let fee = item.tx.fee;
-        ctrace!(MEM_POOL, "New tx with size {}", item.tx.heap_size_of_children());
+        ctrace!(MEM_POOL, "New tx with size {}", size_of_val(&item.tx) + item.tx.heap_size_of_children());
```
sgkim126: Extract `size_of_val(&item.tx) + item.tx.heap_size_of_children()` to a variable. You don't need to calculate it twice.

PR #1072, opened 2019-01-09
Make the verification succeed even if the block's timestamp is not changed

majecty:
Since in the Tendermint consensus the network could create a few blocks in a second, we should allow blocks with the same timestamp.

JinGyeong previously said that for timelock verification, only the mem_pool.rs:1077 should be fixed.


jhs7jhs:
@majecty Is `verify_transfer_timelock()` function in `codechain_machine.rs` okay?

joojis:
@jhs7jhs I checked the `verify_transfer_timelock()` and I see no problem.

Also, it's okay for me.

sgkim126:
@kseo How about you? Do you have any concerns?

jhs7jhs:
@sgkim126 @joojis Merge if it is okay to go.

PR #1074, opened 2019-01-09
Add a tc for transfer-not-allowed case

HoOngEe:
New tc for the case of the unallowed output lock script hash.


jhs7jhs:
@HoOngEe How about "Add unit test case on limited asset"

core/src/miner/mem_pool.rs
```diff
@@ -115,11 +116,11 @@ impl TransactionOrder {
     fn for_transaction(item: &MemPoolItem, seq_seq: u64) -> Self {
         let rlp_bytes_len = rlp::encode(&item.tx).to_vec().len();
         let fee = item.tx.fee;
-        ctrace!(MEM_POOL, "New tx with size {}", item.tx.heap_size_of_children());
+        ctrace!(MEM_POOL, "New tx with size {}", size_of_val(&item.tx) + item.tx.heap_size_of_children());
         Self {
             seq_height: item.seq() - seq_seq,
             fee,
-            mem_usage: item.tx.heap_size_of_children(),
+            mem_usage: size_of_val(&item.tx) + item.tx.heap_size_of_children(),
```
sgkim126: I think this change is not related to this change.

state/src/impls/shard_level.rs
```diff
@@ -1057,6 +1057,51 @@ mod tests {
         ]);
     }
 
+    #[test]
+    fn mint_and_transfer_not_allowed() {
```
sgkim126: Would you add a test for a successful case also?

HoOngEe: tc added for the case of transfer_allowed

state/src/impls/shard_level.rs
```diff
@@ -1057,6 +1057,103 @@ mod tests {
...
+        let sender = address();
+        let mut state_db = RefCell::new(get_temp_state_db());
+        let mut shard_cache = ShardCache::default();
+        let mut state = get_temp_shard_state(&mut state_db, SHARD_ID, &mut shard_cache);
+        let mut allowed_script_hashes = Vec::new();
+
+        let metadata = "metadata".to_string();
+        let lock_script_hash = H160::from("b042ad154a3359d276835c903587ebafefea22af");
+        let random_lock_script_hash = H160::random();
+        allowed_script_hashes.push(lock_script_hash);
```
```rust
let allowed_script_hashes = vec![lock_script_hash, random_lock_script_hash];
```

Issue #1069, opened 2019-01-09
ResealTimer::set_min_timer causes a crash


majecty:
I found that `reseal_timer.schedule_once` in the `ResealTimer::set_min_timer` causes crash.
I suspect that the cause of the crash is https://github.com/CodeChain-io/codechain/pull/1041 's modification.

@sgkim126 @foriequal0 
How about adding an atomic version of `cancle_and_schedule_once`?

sgkim126:
@foriequal0 
I have three questions.
1. Is the order of schedule and cancel in the same thread guaranteed?
2. What happens if the scheduled token is scheduled again.
3. What happens if the canceled token is canceled again.

foriequal0:
@sgkim126

1. Yes
  `Schedule::schedule` function puts a `ScheduleStateControl` to the `SchedulerInner` (with lock), and `Schedule::cancel` function finds `ScheduleStateControl` from it, and it should be sequential. But even if it succeeds to fire cancel signal, it can race with on_timeout scheduling (callback could be called)

2. Before the token is expired, it should return Err(TokenAlreadyScheduled), otherwise, it should succeed.

3. It just returns `false`.

sgkim126:
@majecty I discussed with @foriequal0 
IMO, the atomic `cancle_and_schedule_once` works in the same if
1. the order in the same transaction is guaranteed
2. it's okay to schedule already scheduled token
3. it's okay to cancel already canceled token
And currently, the only error returned by `schedule_once` is `AlreadyScheduled`.
So I think it's okay if you just ignore the error of `schedule_once` function.

majecty:
OK, I'll fix it by ignoring AlreadyScheduled error.

PR #1075, opened 2019-01-09
Ignore the TokenAlreadyScheduled error of the reseal_timer

majecty:
This fixes https://github.com/CodeChain-io/codechain/issues/1069


core/src/client/client.rs
```diff
@@ -286,18 +286,32 @@ impl ResealTimer for Client {
...
-            reseal_timer
+            match reseal_timer
                 .schedule_once(self.importer.miner.get_options().reseal_max_period, RESEAL_MAX_TIMER_TOKEN)
-                .expect("Reseal max timer set succeeds");
+            {
+                Ok(_) => {}
+                Err(TimerScheduleError::TokenAlreadyScheduled) => {
+                    // Since set_max_timer could be called in multi thread, ignore the TokenAlreadyScheduled error
+                }
+                Err(_) => unreachable!("Reseal max timer set succeeds"),
```
foriequal0: We can't paraphrase the error message as "Expect ... succeeds, but failed". Should we change it as a negative statement?

core/src/client/client.rs
```diff
@@ -286,18 +286,32 @@ impl ResealTimer for Client {
...
+            match reseal_timer
                 .schedule_once(self.importer.miner.get_options().reseal_min_period, RESEAL_MIN_TIMER_TOKEN)
-                .expect("Reseal min timer set succeeds");
+            {
+                Ok(_) => {}
+                Err(TimerScheduleError::TokenAlreadyScheduled) => {
+                    // Since set_min_timer could be called in multi thread, ignore the TokenAlreadyScheduled error
+                }
+                Err(_) => {
+                    unreachable!("Reseal min timer set succeeds");
```
foriequal0: Same as above. A negative statement?

core/src/client/client.rs
majecty: How about this?
`unreachable!("Reseal max timer should not fail but failed with {:?}", err)`

foriequal0: Looks Good

majecty:
@sgkim126 @foriequal0 I changed the error message.

PR #1083, opened 2019-01-10
Redesign the mempool

jhs7jhs:
`TransactionSet` is removed. The new design will add/remove transactions efficiently than before.


core/src/miner/mem_pool.rs
```diff
@@ -236,274 +236,82 @@ impl MemPoolItem {
...
-                .remove(&order.hash)
-                .expect("hash is in `by_priorty`; all hashes in `by_priority` must be in `by_hash`; qed");
-
-            let min = removed.get(&sender).map_or(seq, |val| cmp::min(*val, seq));
-            removed.insert(sender, min);
-            removed
-        }))
+    fn add(&mut self, fee: u64) {
+        ctrace!(MEM_POOL, "Fee counter add {}", fee);
+        *self.counter.entry(fee).or_insert(0) += 1;
```
sgkim126:
```rust
*self.counter.entry(fee).or_default() += 1;
```

core/src/miner/mem_pool.rs
```diff
@@ -611,45 +488,142 @@ impl MemPool {
...
+                    }
+                    QueueTag::Future => {
+                        self.future.remove(&old_order);
+                        self.future_memory_usage -= old_order.mem_usage;
+                    }
+                    QueueTag::New => unreachable!(),
+                }
+            }
+
+            to_insert.entry(signer_public).or_insert_with(Vec::new).push(seq);
```
sgkim126:
```rust
to_insert.entry(signer_public).or_default().push(seq);
```

core/src/miner/miner.rs
```diff
@@ -264,29 +266,43 @@ impl Miner {
...
+                seq: client.latest_seq(&a),
+                balance: client.latest_balance(&a),
+            }
+        };
+
+        let insertion_results = mem_pool.add(to_insert, insertion_time, insertion_timestamp, &fetch_account);
+        let mut result_idx = 0;
+
+        let results = intermediate_results
+            .into_iter()
```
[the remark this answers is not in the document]
jhs7jhs: Fixed.

core/src/miner/mem_pool.rs
jhs7jhs: Fixed. This will require type declaration on `to_insert`.

jhs7jhs:
Made new structs: `CurrentQueue` and `FutureQueue`.

jhs7jhs:
@sgkim126 I've added a logic on checking local transactions.

core/src/miner/mem_pool.rs
```diff
@@ -611,45 +525,139 @@ impl MemPool {
...
+
+            if let Some(seq_list) = to_insert.get(&public) {
+                self.add_new_orders_to_queue(public, seq_list, new_next_seq);
+            }
+
+            self.by_signer_public.clear_if_empty(&public);
+        }
+
+        // FIXME: Enforcing limit too late?
+        // The logic above will allow memory usage more than the limit for a moment.
```
sgkim126: I think it's okay.
Actually, we don't enforce that the mempool uses less than the limitation because of the local transactions.

sgkim126: For the same reason, I think it's okay not to call `enforce_limit` in `set_limit`.

jhs7jhs: But the original behavior is to call `enforce_limit`. I'm not sure...
+ I forgot to not to include local/retracted transactions to the memory limit. I'll fix this.

core/src/miner/mem_pool.rs
```diff
@@ -673,138 +726,269 @@ impl MemPool {
...
     }
 
     /// Removes invalid transaction identified by hash from pool.
     /// Assumption is that this transaction seq is not related to client seq,
     /// so transactions left in pool are processed according to client seq.
     ///
     /// If gap is introduced marks subsequent transactions as future
-    pub fn remove<F>(&mut self, tx_hash: &H256, fetch_seq: &F, current_time: PoolingInstant, timestamp: u64)
-    where
+    pub fn remove<F>(
```
majecty: I'm happy to see the `remove` function get many transaction hashes, not just a tx_hash.

core/src/miner/mem_pool.rs
```diff
@@ -611,45 +559,150 @@ impl MemPool {
...
+                to_insert
+                    .get(&public)
+                    .and_then(|v| self.check_new_transactions(public, v, next_seq, current_time, current_timestamp))
+                    .unwrap_or_else(|| self.check_transactions(public, next_seq, current_time, current_timestamp))
+            };
+
+            let is_this_account_local = new_local_accounts.contains(&public);
+            // Need to update transactions because of height/origin change
+            if current_seq != first_seq || is_this_account_local {
+                self.update_orders(public, current_seq, new_next_seq, is_this_account_local);
```
majecty: Great!

PR #1091, opened 2019-01-11
Use camelCase for JSON-RPC

sgkim126:


sgkim126:
I checked the test and integration test passed on one platform each.
I'll merge it since it's not a platform dependent.

PR #1096, opened 2019-01-11
Hide the highest block's data

majecty:
I fixed the code because some wrong information was saved when the highest block is changed.

How to change:

A node should not be able to find BlockHash from the highest block's number.
A node should not be able to find a transaction's id from the transaction's hash which is in the highest block.
A node should not be able to find invoices which in the highest block.

A node should be able to find BlockDetail from the highest block's hash.
A node should be able to find Header/Block from the highest block's hash.

### DIFF
--- a/core/src/blockchain/headerchain.rs
+++ b/core/src/blockchain/headerchain.rs
@@ -75,7 +75,6 @@ impl HeaderChain {
                     number: genesis.number(),
                     total_score: genesis.score(),
                     parent: genesis.parent_hash(),
-                    children: vec![],
                 };
 
                 let mut batch = DBTransaction::new();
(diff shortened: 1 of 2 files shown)


core/src/blockchain/headerchain.rs
```diff
@@ -190,8 +190,10 @@ impl HeaderChain {
         match best_header_changed {
             BestHeaderChanged::None => (),
             BestHeaderChanged::CanonChainAppended {
-                ..
+                best_header,
             } => {
+                let header_view = HeaderView::new(best_header);
+                hashes.insert(header_view.number(), header_view.hash());
                 hashes.insert(new_header.number(), new_header.hash());
```
sgkim126: Isn't `new_header` the highest header?
Is it okay that `new_hash_entries` returns a hash of the highest one?

majecty: Removing the `new_header` is included in this commit(https://github.com/CodeChain-io/codechain/pull/1096/commits/43de1349ad32bf8e0a5e5fe8164f5b28183a156b).

sgkim126: How about merging it with 43de134?

sync/src/block/extension.rs
```diff
@@ -440,8 +440,18 @@ impl Extension {
...
 
     fn create_headers_response(&self, start_number: BlockNumber, max_count: u64) -> ResponseMessage {
+        let highest_header = self.client.highest_header();
         let headers = (0..max_count)
-            .map(|number| self.client.block(&BlockId::Number(start_number + number)))
+            .map(|number| {
+                let height = start_number + number;
+                let block_id = if highest_header.number() == height {
+                    // We could not get highest block with height
+                    BlockId::Hash(highest_header.hash())
```
jhs7jhs: How about add a comment about relation between `highest_header_number` and `best_block_number` in here?
Currently, every consensus except Tendermint ensures `best_block_num = highest_block_num` and Tendermint ensures `best_block_num + 1 >= highest_block_num`.

majecty: Good

majecty: I added a comment.

majecty:
@jhs7jhs @sgkim126 I fixed conflicts. Could you approve this PR again?

PR #1105, opened 2019-01-14
Reorganize integration tests

sgkim126:
https://doc.rust-lang.org/book/ch11-03-test-organization.html


vm/tests/chk_multi_sig.rs
```diff
@@ -14,22 +14,31 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+extern crate codechain_crypto as ccrypto;
+extern crate codechain_key as ckey;
+extern crate codechain_types as ctypes;
+extern crate codechain_vm as cvm;
+extern crate primitives;
```
Kais-DkM: `vm/tests/common.rs` seems to have the same `extern crate` lines.
Do we need them in this file too?

sgkim126: I checked that these lines can be removed if we put a module in a subdirectory.

PR #1122, opened 2019-01-16
Remove the previous block hash from last_confirmed_view in Tendermint

jhs7jhs:


jhs7jhs:
@majecty There's a statement which uses `last_confirmed_view.1` in #1146, so I've modified a bit. Could you check this again?

PR #1146, opened 2019-01-18
Request pre-commits when in the commit state in Tendermint

jhs7jhs:
Should be merged after #1144


jhs7jhs:
Removing `do-not-merge` tag because #1144 is merged.

Issue #1148, opened 2019-01-21
Split `asset_type` to `H160` and `ShardId`.


sgkim126:
Currently, `asset_type` is a 256-bit hash with the prefix(`'S'` and shard id).
Actually, it is the same with the address of the asset scheme.
But we don't need to use the address as an asset type.
It is sufficient to be able to compute the address from the asset type.
So I propose to use the blake160 hash of the mint transaction tracker as an asset type and add a `shard_id` field.

PR #1164, opened 2019-01-23
Change the definition of the asset type

jhs7jhs:
Resolves: #1148 / Resolves: #1166 

Old: (H256 of tracker of mint tx) + (shard id)
New: H160 of tracker of mint tx only.

`AssetSchemeAddress` was exactly same as the asset type in the old definition.
From the new definition, `AssetSchemeAddress` is calculated from the shard id and the blake256 hash of the asset type.

`shard_id` field is added to each struct which has `asset_type` field.

### DIFF
--- a/rpc/src/v1/impls/chain.rs
+++ b/rpc/src/v1/impls/chain.rs
@@ -20,14 +20,15 @@ use ccore::{
     AssetClient, BlockId, EngineInfo, ExecuteClient, MinerService, MiningBlockChainClient, RegularKey, RegularKeyOwner,
     Shard, SignedTransaction, TextClient, UnverifiedTransaction,
 };
+use ccrypto::Blake;
 use cjson::bytes::Bytes;
 use cjson::uint::Uint;
 use ckey::{public_to_address, NetworkId, PlatformAddress, Public};
-use cstate::{AssetSchemeAddress, FindActionHandler};
+use cstate::FindActionHandler;
 use ctypes::invoice::Invoice;
 use ctypes::transaction::{Action, ShardTransaction as ShardTransactionType};
 use ctypes::{BlockNumber, ShardId};
-use primitives::{Bytes as BytesArray, H256};
+use primitives::{Bytes as BytesArray, H160, H256};
 use rlp::{DecoderError, UntrustedRlp};
 
 use jsonrpc_core::Result;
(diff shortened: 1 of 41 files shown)


rpc/src/v1/impls/chain.rs
```diff
@@ -124,16 +124,20 @@ where
         shard_id: ShardId,
         block_number: Option<u64>,
     ) -> Result<Option<AssetScheme>> {
-        let address = AssetSchemeAddress::new(tracker, shard_id);
-        self.get_asset_scheme_by_type(address.into(), block_number)
+        let block_id = block_number.map(BlockId::Number).unwrap_or(BlockId::Latest);
+        let address = AssetSchemeAddress::new_from_tracker(tracker, shard_id);
+        self.client.get_asset_scheme(address, block_id).map_err(errors::transaction_state)
```
sgkim126: How about making `get_asset_scheme` receive the `asset_type` and `shard_id` instead of the asset scheme address?

jhs7jhs: I think it's a good idea. Also I'll look at other files using `AssetSchemeAddress`.

jhs7jhs: Now `AssetSchemeAddress` is only used in codechain-state. I'll delete it from the SDK too.

jhs7jhs:
Now passed all tests in my local environment.

types/src/transaction/error.rs
```diff
@@ -229,17 +227,23 @@ impl Encodable for Error {
             Error::CannotBurnCentralizedAsset => tag_with(s, ERROR_ID_CANNOT_BURN_CENTRALIZED_ASSET),
             Error::CannotComposeCentralizedAsset => tag_with(s, ERROR_ID_CANNOT_COMPOSE_CENTRALIZED_ASSET),
             Error::InvalidDecomposedInput {
-                address,
+                asset_type,
+                shard_id,
                 got,
-            } => tag_with(s, ERROR_ID_INVALID_DECOMPOSED_INPUT).append(address).append(got),
+            } => tag_with(s, ERROR_ID_INVALID_DECOMPOSED_INPUT).append(asset_type).append(shard_id).append(got),
```
sgkim126: You should change `fn list_length_for(tag: u8)`

jhs7jhs: Fixed. There are various error types, and every implementation is different from others...

PR #1165, opened 2019-01-23
Declare types for rpc

sgkim126:


state/Cargo.toml
```diff
@@ -21,6 +21,4 @@ primitives = { git = "https://github.com/CodeChain-io/rust-codechain-primitives.
 rlp = { path = "../util/rlp" }
 rlp_derive = { path = "../util/rlp_derive" }
 rustc-hex = "1.0"
-serde = "1.0"
-serde_derive = "1.0"
```
majecty: I'm happy to see "serde" is removed from the state module.

PR #1177, opened 2019-01-24
Add chain_getPendingTransactionsCount API

ScarletBlue:


spec/JSON-RPC.md
```diff
@@ -1264,6 +1265,34 @@ No parameters
...
 [Back to **List of methods**](#list-of-methods)
 
+## chain_getPendingTransactionsCount
+Returns a count of the transactions that are in the current transaction queue.
+
+### Params
+No parameters
+
+### Returns
+`Count`
```
majecty: Let's use `U64`

sgkim126: How about using `number`?
I think it gives enough precision for the number of transactions.

ScarletBlue: I've made the changes to `number`.

PR #1180, opened 2019-01-24
Skip unit tests if only e2e tests are changed

sgkim126:


sgkim126:
@foriequal0 PR updated

PR #1182, opened 2019-01-24
Fix flaky e2e tests

### LINKED ISSUE #1183: Fix flaky e2e tests.
Recent days, the failure rates of tests which use [TestHelper](https://github.com/CodeChain-io/codechain-test-helper-js) has been increased.

sgkim126:
Fix #1183


sgkim126:
The main reason is that the test helper signed with the wrong nonce.

PR #1187, opened 2019-01-25
Fix Tendermint proposal block importing logic

jhs7jhs:
- `engine.proposal_generated()` should be called first, and `chain.import_sealed_block()` should be called next.

- `chain.import_sealed_block()` will call `new_blocks()` and then `on_imported_proposal()`. So the empty proposal timer should be triggered in `on_imported_proposal()`, not `proposal_generated()`.

- `proposal_generated()` must not change the Tendermint state to the prevote state, nor broadcast the proposal block.


core/src/consensus/tendermint/mod.rs
```diff
@@ -557,8 +557,27 @@ impl TendermintInner {
...
-            if self.step.is_propose() {
-                self.move_to_step(Step::Prevote);
+            let block = match self.step {
+                TendermintState::Propose => {
+                    self.move_to_step(Step::Prevote);
+                    return
+                }
+                TendermintState::ProposeWaitImported {
+                    ref block,
+                } => {
```
majecty: Let's add an assertion that checks proposal.hash == block.hash

PR #1189, opened 2019-01-25
Fix the bug that a node does not request unknown votes

majecty:


core/src/consensus/tendermint/mod.rs
```diff
@@ -1037,7 +1037,19 @@ impl TendermintInner {
                 return Ok(())
             }
 
-            if message.on.step.height == self.height && message.on.step.view == self.view {
+            let current_vote_step = if self.step.is_commit() {
```
jhs7jhs: I think this one can not be modified if #1188 is merged.

PR #1199, opened 2019-01-29
No need to use sudo-enabled Travis worker

sgkim126:
We used sudo-enabled workers because container-based workers don't give
enough memory to compile CodeChain. But after the machine spec has been
changed, container-based works have enough memory.


sgkim126:
See https://github.com/travis-ci/docs-travis-ci-com/commit/aa5c5686833db746c855af488a9e0dc2b427ca12#diff-dfdcbe4cee20a496db3112fd77473d03 for more information.

PR #1203, opened 2019-01-29
Save votes from received proposal blocks

### LINKED ISSUE #1192: Save votes from received proposal blocks
When a node received the _h+1th_ proposal in the height _h_, and moved to the _h+1_, the node does not have precommits in its `votes` data structure. The node should add precommits in the proposal block to the `votes`.

majecty:
Close #1192

### DIFF
--- a/core/src/consensus/tendermint/types.rs
+++ b/core/src/consensus/tendermint/types.rs
@@ -18,11 +18,13 @@ use std::cmp::PartialEq;
 use std::fmt;
 use std::ops::Sub;
 
-use primitives::H256;
+use ckey::SchnorrSignature;
+use primitives::{Bytes, H256};
 use rlp::{Decodable, DecoderError, Encodable, RlpStream, UntrustedRlp};
 
 use super::message::VoteStep;
 use crate::block::{IsBlock, SealedBlock};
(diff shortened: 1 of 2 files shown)


core/src/consensus/tendermint/types.rs
```diff
@@ -275,3 +315,38 @@ impl PeerState {
...
+            &self.seal.get(2).expect("block went through verify_block_basic; block has .seal_fields() fields; qed"),
+        )
+    }
+
+    pub fn signatures(&self) -> Result<Vec<(usize, SchnorrSignature)>, Error> {
+        let precommits = self.precommits();
+        let bitset = self.bitset()?;
+
+        let bitset_iter = bitset.true_index_iter();
+        let signatures: Vec<SchnorrSignature> =
```
sgkim126:
```rust
        let signatures = precommits.iter().map(|rlp| rlp.as_val::<SchnorrSignature>());
        Ok(bitset_iter
            .zip(signatures)
            .map::<Result<(usize, SchnorrSignature), Error>, _>(|(a, b)| Ok((a, b?)))
            .collect::<Result<_, _>>()?)
 ```

PR #1211, opened 2019-01-30
Fix the bug that restoring Tendermint causes crash because the extension is not initialized

majecty:
Now the Tendermint does not use the extension in the `restore` function.
Because the network extension has an initial timeout, the `restore` does not need to create its own.


sgkim126:
Should we call `set_timer_step` on `restore`?

majecty:
@sgkim126 I changed it.

PR #1213, opened 2019-01-30
Load the Tendermint's backup data by calling `handle_message`

majecty:
I found bugs on the backup and recovery of the Tendermint.
There have been a few things that have not been handled while editing other parts of the Tendermint.
I judged that rather than adding the modifications to the backup, calling normal path codes is easier.


sgkim126:
@majecty Would you explain a bit more?

majecty:
@sgkim126 I added the description.

majecty:
Thanks. You're right. It is used only in `restore`. I'll change the name

majecty:
I fixed the conflict.

Issue #930, opened 2018-12-03
No way to get a composed asset scheme when composing and decomposing in the same block


sgkim126:
`AssetDecompose` removes the asset scheme.
When composing and decomposing parcel handled in the same block, there is no way to find the scheme of the composed asset.
It causes a problem on the [Explorer](https://github.com/CodeChain-io/codechain-explorer) since it must show all asset scheme that once existed.

Keeping the scheme of the decomposed asset can be a solution.
But I'll dig more whether it has no side-effect.

PR #1214, opened 2019-01-30
Do not kill asset scheme when decomposing

jhs7jhs:
Resolves: #930.
I'm sure that the only side effect is that it will take more memory/database spaces. This can be resolved indirectly by giving high fee on compose/decompose transactions.
Leaving `kill_asset_scheme()` as dead code for `git blame` and a future usage.

### DIFF
--- a/state/src/impls/shard_level.rs
+++ b/state/src/impls/shard_level.rs
@@ -637,7 +637,6 @@ impl<'db> ShardLevelState<'db> {
         self.check_and_run_input_script(input, transaction, None, false, sender, approvers, client)?;
 
         self.kill_asset(&asset_address);
-        self.kill_asset_scheme(&asset_scheme_address);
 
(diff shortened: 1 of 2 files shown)


state/src/impls/shard_level.rs
```diff
@@ -714,6 +713,7 @@ impl<'db> ShardLevelState<'db> {
         self.cache.remove_asset(account);
     }
 
+    #[allow(dead_code)]
     fn kill_asset_scheme(&mut self, account: &AssetSchemeAddress) {
```
sgkim126: I think it's fine to remove this method.

jhs7jhs: I'll remove it.

jhs7jhs: Removed.

Issue #1229, opened 2019-02-02
Panic at "add_transactions_to_pool"


majecty:
CodeChain Version : df2a74b6c134b7c04cac6a28038c3324dee8f358

```
====================

stack backtrace:
   0:     0x5640ee8d0fd6 - backtrace::backtrace::libunwind::trace::h6f69f7a8523b3b48
                        at /home/majecty/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.7/src/backtrace/libunwind.rs:53
                         - backtrace::backtrace::trace::hfcb48dcb3ac891e8
                        at /home/majecty/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.7/src/backtrace/mod.rs:42
   1:     0x5640ee8cb583 - backtrace::capture::Backtrace::new_unresolved::h3e5684a38718154b
                        at /home/majecty/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.7/src/capture.rs:88
   2:     0x5640ee8cb4dd - backtrace::capture::Backtrace::new::h448ec18ea4124027
                        at /home/majecty/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.7/src/capture.rs:63
   3:     0x5640ed7ea31d - panic_hook::panic_hook::ha4aecdc1727731f8
                        at util/panic_hook/src/lib.rs:53
   4:     0x5640ed7eacc7 - core::ops::function::Fn::call::h50bcd1f0a1621bec
                        at /rustc/9fda7c2237db910e41d6a712e9a2139b352e558b/src/libcore/ops/function.rs:78
   5:     0x5640ef2983d9 - std::panicking::rust_panic_with_hook::h8cbdfe43764887be
                        at src/libstd/panicking.rs:495
   6:     0x5640ef297e81 - std::panicking::continue_panic_fmt::h3d3c5a833c00a5e1
                        at src/libstd/panicking.rs:398
   7:     0x5640ef297d65 - rust_begin_unwind
                        at src/libstd/panicking.rs:325
   8:     0x5640ef2b39dc - core::panicking::panic_fmt::h4d67173bc68f6d5a
                        at src/libcore/panicking.rs:95
   9:     0x5640ef2b3990 - core::panicking::panic_bounds_check::h620ba7e4e283471a
                        at src/libcore/panicking.rs:71
[...]

PR #1231, opened 2019-02-02
Fix the out of index crash

majecty:
The insertion_results only contains successful results, and the
intermediate_results contains both successful and failed results. So
when accessing the insertion_results by index, we should not count
failed results.


majecty:
This will fix #1229

PR #1238, opened 2019-02-07
Add a test case for the pr#1231

HoOngEe:
The `add_transactions_to_pool` function rejected some 
transactions early before the `mem_pool.add` function call. 
Before the pull request 1231 is merged, there are chances of
crashing with out-of-index in insertion_result vector. 
This commit added a test case which occurs that sort of crash.


sgkim126:
@majecty ping

Issue #1253, opened 2019-02-08
Implement RPC to get an error hint


sgkim126:
Currently, the failed invoice contains RuntimeError to indicate the reason for failure.
However, this makes maintaining code and revising the spec difficult because the order of verification should be kept.
Thus, I propose to remove RuntimeError from an invoice. This will simply make the invoice a boolean.

And it should add a new RPC that returns the RuntimeError that the node has. I think the specification below will suffice.

---
## chain_getErrorHint
Gets a hint to find out why the transaction failed.
It returns `null` if the transaction is not involved in the chain or succeeded.
If the transaction failed, this should return the reason for the transaction failing.
It's better to return the human-readable string; however, the message is not guaranteed.

Other nodes may return a different message at the same time.
And it does not guarantee the consistency of the values returned by the same node.

### Params
 1. transaction hash - `H256`
 2. block number: `number` | `null`

### Returns
`null` | `string` - `null` if there is no hint, `string` if the transaction failed.

Errors: `Invalid Params`

sgkim126:
https://github.com/CodeChain-io/codechain/pull/1258 It changed the type of Invoice to boolean.
The RPCs are not implemented yet.

sgkim126:
I removed `chain_getResultsByTracker` and renamed `chain_getResult` to `chain_getErrorHint`.
The RPC will not distinguish whether a transaction was successful or was not executed.

sgkim126:
Closed by #1270

PR #1258, opened 2019-02-11
Remove the failure reason from Invoice

sgkim126:


sgkim126:
It's related to https://github.com/CodeChain-io/codechain-sdk-js/pull/312

sgkim126:
@xpdlf1004 @joojis It makes the type of invoice to boolean.
So the current errorType field will be removed. Instead of it, I'll implement an RPC to get the hint of transaction failure. I'll change codechain-indexer to provide it.

sgkim126:
See https://github.com/CodeChain-io/codechain/issues/1253 for more information

Issue #1247, opened 2019-02-08
The Tendermint mistakes some valid votes for old votes


majecty:
CodeChain Version : fa87b07ca662ade09065a92e35ae70f61ae7efa4

```
#235917509 2019-02-04 20:22:34  Client Worker #1 DEBUG engine  Transitioning to height 2262.
#235917509 2019-02-04 20:22:34  Client Worker #1 TRACE timer  cancel(TimerName(tendermint), ScheduleId(4, 28))
#235917509 2019-02-04 20:22:34  Client Worker #1 TRACE timer  schedule(TimerName(tendermint), ScheduleId(4, 29), after: 10s, repeat: None)
#235917509 2019-02-04 20:22:34  Client Worker #1 TRACE engine  Broadcast state VoteStep { height: 2262, view: 0, step: Prevote } Some(0x82fcd5d50855ebbdc73565e72cde4030bff1d9645d4ef5e4a92234db3e0f9b89) [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
#235917509 2019-02-04 20:22:34  Client Worker #1 TRACE engine  move_to_step: Prevote.
[...]

PR #1262, opened 2019-02-12
Remove votes after inserting votes from the new proposal

majecty:
Since the `VoteCollector` should always have at least one vote, we
should insert new votes before deleting old votes.

Fix https://github.com/CodeChain-io/codechain/issues/1247


core/src/consensus/vote_collector.rs
```diff
@@ -138,6 +138,7 @@ impl<M: Message + Default + Encodable + Debug> VoteCollector<M> {
     pub fn throw_out_old(&self, vote_round: &M::Round) {
         let mut guard = self.votes.write();
         let new_collector = guard.split_off(vote_round);
+        debug_assert!(false, new_collector.is_empty());
```
sgkim126:
```rust
debug_assert!(!new_collector.is_empty());
```

PR #1267, opened 2019-02-12
Add test case for rejected-by-limit-and-reaccepted

HoOngEe:
Test case for the situation where some transactions rejected to enter the mempool because of the mempool size limit but accepted by next request.

### DIFF
--- a/test/src/e2e.long/mempool.test.ts
+++ b/test/src/e2e.long/mempool.test.ts
@@ -18,6 +18,7 @@ import { expect } from "chai";
 import "mocha";
 import { wait } from "../helper/promise";
 import CodeChain from "../helper/spawn";
+import { SignedTransaction } from "codechain-sdk/lib/core/classes";
 
 const BASE = 200;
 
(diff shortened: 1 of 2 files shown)


test/src/e2e.long/mempool.test.ts
```diff
@@ -80,6 +81,73 @@ describe("Memory pool size test", function() {
...
+                (await nodeB.sdk.rpc.chain.getPendingTransactions()).length > 0
+            ) {
+                await wait(500);
+                counter += 1;
+            }
+            await wait(500 * (counter + 1));
+
+            await nodeB.sdk.rpc.devel.stopSealing();
+
+            rejectedTransactions.forEach(async (tx: SignedTransaction) => {
```
sgkim126:
```ts
Promise.all(rejectedTransactions.map((tx: SignedTransaction) => nodeB.sdk.rpc.chain.sendSignedTransaction(tx));
```

sgkim126: `forEach` doesn't guarantee the promises are resolved.

test/src/e2e.long/mempool.test.ts
```diff
@@ -80,6 +81,73 @@ describe("Memory pool size test", function() {
...
+
+            counter = 0;
+            while (
+                (await nodeB.sdk.rpc.chain.getPendingTransactions()).length <
+                rejectedTransactions.length
+            ) {
+                await wait(500);
+                counter += 1;
+            }
+            await wait(500 * (counter + 1));
```
sgkim126: Why do you wait here?

test/tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "target": "es2015",
     "types": ["node"],
     "module": "commonjs",
+    "lib": ["dom", "es2017"],
```
sgkim126: Why do you add dom?

test/src/e2e.long/mempool.test.ts
```diff
@@ -80,6 +81,73 @@ describe("Memory pool size test", function() {
...
+
+            pendingTransactions = await nodeB.sdk.rpc.chain.getPendingTransactions();
```
sgkim126: Don't reuse the variable.

test/src/e2e.long/mempool.test.ts
```diff
@@ -80,6 +81,73 @@ describe("Memory pool size test", function() {
...
+            pendingTransactionHashes = pendingTransactions.map(
```
sgkim126: ditto

test/src/e2e.long/mempool.test.ts
```diff
@@ -80,6 +81,73 @@ describe("Memory pool size test", function() {
...
+            await nodeB.sdk.rpc.devel.startSealing();
+
+            counter = 0;
+            while (
+                (await nodeB.sdk.rpc.chain.getPendingTransactions()).length > 0
+            ) {
+                await wait(500);
+                counter += 1;
+            }
+            await wait(500 * (counter + 1));
```
sgkim126: Why do you wait?

HoOngEe: wait removed

test/src/e2e.long/mempool.test.ts
HoOngEe: new variables are introduced

test/tsconfig.json
HoOngEe: removed

test/src/e2e.long/mempool.test.ts
HoOngEe: Fixed

test/src/e2e.long/mempool.test.ts
HoOngEe: wait removed

test/src/e2e.long/mempool.test.ts
```diff
@@ -80,6 +81,75 @@ describe("Memory pool size test", function() {
...
+            const sent = [];
+            for (let i = 0; i < sizeLimit * 2; i++) {
+                sent.push(
+                    await nodeA.sendPayTx({
+                        seq: i,
+                        awaitInvoice: false
+                    })
+                );
+            }
+            let counter = 0;
```
sgkim126: This isn't used.

test/src/e2e.long/mempool.test.ts
```diff
@@ -80,6 +81,75 @@ describe("Memory pool size test", function() {
...
+            while (
+                (await nodeB.sdk.rpc.chain.getPendingTransactions()).length > 0
+            ) {
+                await wait(500);
+                counter += 1;
+            }
+
+            await nodeB.sdk.rpc.devel.stopSealing();
+
+            Promise.all(
```
sgkim126: Sorry. I missed using `await`. It should be `await Promise.all(` if you want to wait to send.
But it seems that you don't need to wait for it. Because you count the number of pending transactions on line 131.

HoOngEe: Then I think It's better to add await in front of Promise.all and remove the waiting until the number of pending transactions and rejected transactions become equal.
The last part of test check all rejected transactions are now in pending transactions, it automatically check the number of two arrays are same.

sgkim126: Would you check when RPC responses for chain_sendSignedTransaction?

PR #1270, opened 2019-02-13
Implement RPC getErrorHint

### LINKED ISSUE #1253: Implement RPC to get an error hint

HoOngEe:
Implement new RPC which get error string from invoice.

It closes #1253

### DIFF
--- a/rpc/src/v1/impls/chain.rs
+++ b/rpc/src/v1/impls/chain.rs
@@ -189,6 +189,14 @@ where
         Ok(self.client.balance(address, block_id.into()).map(Into::into))
     }
 
+    fn get_error_hint(&self, transaction_hash: H256) -> Result<Option<String>> {
+        if let Some(Invoice::Failure(error_string)) = self.client.parcel_invoice(&transaction_hash.into()) {
+            Ok(Some(error_string))
+        } else {
+            Ok(None)
+        }
+    }
+
     fn get_regular_key(&self, address: PlatformAddress, block_number: Option<u64>) -> Result<Option<Public>> {
         let block_id = block_number.map(BlockId::Number).unwrap_or(BlockId::Latest);
         let address = address.try_address().map_err(errors::core)?;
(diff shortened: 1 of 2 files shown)


core/src/client/mod.rs
```diff
@@ -245,7 +245,6 @@ pub trait BlockChainClient:
     /// Get block status by block header hash.
     fn block_status(&self, id: &BlockId) -> BlockStatus;
 
-
```
sgkim126: Unrelated change

HoOngEe: Now that change is removed

rpc/src/v1/impls/chain.rs
```diff
@@ -189,6 +189,17 @@ where
         Ok(self.client.balance(address, block_id.into()).map(Into::into))
     }
 
+    fn get_error_hint(&self, transaction_hash: H256) -> Result<Option<String>> {
+        let invoice_op = self.client.parcel_invoice(&transaction_hash.into());
```
sgkim126: This can be simplified as below.
```rust
        if let Some(Invoice::Failure(error_string)) = self.client.parcel_invoice(&transaction_hash.into()) {             
            Ok(Some(error_string))                                                                                       
        } else {                                                                                                         
            Ok(None)                                                                                                     
        }                                                                                                                
```

HoOngEe: Simplified.

PR #1276, opened 2019-02-15
Do not change to the `Propose` step when a proposal is imported

majecty:


sgkim126:
@majecty Would you explain it more?

majecty:
I found a few DoubleVote logs in Corgi.

1. A node sent a prevote message with (block_hash = null) in height `h` and view `v`
2. A node sent a prevote message with (block_hash = Some(~)) in height `h` and view `v`

In the code, TendermintStatus is changed to the Propose step when a proposal is imported. However, it is not intended behavior. If a node is in (`Prevote`/`Precommit`/`Commit`) state, it should not move to the `Propose` state when a proposal is imported. And If a node is in `Propose` state, it should move to the `Prevote` state.

core/src/consensus/tendermint/mod.rs
```diff
@@ -591,8 +591,8 @@ impl TendermintInner {
         let current_height = self.height;
         if current_height == height && self.view == proposal_view {
             self.proposal = Some(proposal.hash());
-            let prev_step = mem::replace(&mut self.step, TendermintState::Propose);
```
majecty: IMO, this line was added to move `self.step` to a new variable, which is used in match expression, and fills the `self.step` with an empty or default value. However, the `TendermintState::Propose` is not a good empty or default value. Also, the code omitted to return prev_step.

So I changed it by cloning the step field to use in the match expression.

PR #1286, opened 2019-02-19
Modify the calculation method of view_proposer

### LINKED ISSUE #1279: Make the validators create an equal number of blocks even though some validators are inactive.
Currently, the index of the proposer is `(view + height) mod validatorCount`.
When a node is inactive, the inactive node's next proposer creates two blocks sequentially.

If we use the index of the next proposer by `(prev block's proposer's index + 1 + view) mod validatorCount`, all validators will create the same number of blocks.

HoOngEe:
To give each validator an equal chance of proposing, tendermint
should check new propser if its index is the same as
 `previous block's proposer index` + `view` + 1. Previous checking
was problematic when a node is inactive.

It closes #1279

### DIFF
--- a/core/src/client/test_client.rs
+++ b/core/src/client/test_client.rs
@@ -188,37 +188,44 @@ impl TestBlockChainClient {
     pub fn add_blocks(&self, count: usize, transaction_length: usize) {
         let len = self.numbers.read().len();
         for n in len..(len + count) {
-            let mut header = BlockHeader::new();
-            header.set_score(From::from(n));
-            header.set_parent_hash(*self.last_hash.read());
-            header.set_number(n as BlockNumber);
-            header.set_extra_data(self.extra_data.clone());
-            let mut transactions = Vec::new();
-            for _ in 0..transaction_length {
-                let keypair = Random.generate().unwrap();
-                // Update seqs value
-                self.seqs.write().insert(keypair.address(), 0);
-                let tx = Transaction {
-                    seq: 0,
-                    fee: 10,
-                    network_id: NetworkId::default(),
-                    action: Action::Pay {
-                        receiver: Address::random(),
-                        quantity: 0,
-                    },
-                };
-                let signed = SignedTransaction::new_with_sign(tx, keypair.private());
-                transactions.push(signed);
-            }
-            header.set_transactions_root(skewed_merkle_root(
-                *self.last_transactions_root.read(),
-                transactions.iter().map(Encodable::rlp_bytes),
-            ));
-            let mut rlp = RlpStream::new_list(2);
-            rlp.append(&header);
-            rlp.append_list(&transactions);
-            self.import_block(rlp.as_raw().to_vec()).unwrap();
+            self.add_block_with_author(None, n, transaction_length);
         }
+    }
+    /// Add a block to test client with designated author.
+    pub fn add_block_with_author(&self, author: Option<Address>, n: usize, transaction_length: usize) {
+        let mut header = BlockHeader::new();
+        header.set_score(From::from(n));
+        header.set_parent_hash(*self.last_hash.read());
+        header.set_number(n as BlockNumber);
+        header.set_extra_data(self.extra_data.clone());
+        if let Some(addr) = author {
+            header.set_author(addr);
+        }
+        let mut transactions = Vec::with_capacity(transaction_length);
+        for _ in 0..transaction_length {
+            let keypair = Random.generate().unwrap();
+            // Update seqs value
+            self.seqs.write().insert(keypair.address(), 0);
+            let tx = Transaction {
+                seq: 0,
+                fee: 10,
+                network_id: NetworkId::default(),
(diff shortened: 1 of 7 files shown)


core/src/consensus/tendermint/mod.rs
```diff
@@ -203,8 +215,8 @@ impl TendermintInner {
     }
 
     /// Find the designated for the given view.
-    fn view_proposer(&self, bh: &H256, height: Height, view: View) -> Address {
-        let proposer_nonce = height + view;
+    fn view_proposer(&self, bh: &H256, _height: Height, view: View) -> Address {
+        let proposer_nonce = self.prev_block_proposer_idx() + 1 + view;
```
majecty: The `view_proposer` receives the height, view, and prev block's hash, and returns the propser's address.
We should use the give prev block's hash instead of the (current height -1)th block's hash.

Please remind that this equation `received height == tendermint.height` is not always true.

HoOngEe: Fixed.

core/src/consensus/tendermint/mod.rs
```diff
@@ -184,23 +184,30 @@ impl TendermintInner {
...
+            .hash()
     }
 
     /// Get the proposer of previous block to check new proposer is valid.
-    fn prev_block_proposer_idx(&self, height: Height) -> usize {
-        let prev_proposer = self.prev_block_of_height(height).author();
-        self.validators
-            .get_index_by_address(&self.prev_block_hash(), &prev_proposer)
-            .expect("The proposer must be in the validator set")
+    fn prev_block_proposer_idx(&self, height: Height) -> Result<usize, EngineError> {
```
sgkim126: Why do you make it return `Result` instead of `Option`?
The only error that it returns is PrevBlockNotExist. I think EngineError is too broad to handle it.

HoOngEe: I changed the function return type to option and handled error in its caller function.

core/src/client/test_client.rs
```diff
@@ -188,37 +188,44 @@ impl TestBlockChainClient {
...
+    pub fn add_block_with_author(&self, author: Option<Address>, n: usize, transaction_length: usize) {
+        let mut header = BlockHeader::new();
+        header.set_score(From::from(n));
+        header.set_parent_hash(*self.last_hash.read());
+        header.set_number(n as BlockNumber);
+        header.set_extra_data(self.extra_data.clone());
+        if let Some(addr) = author {
+            header.set_author(addr);
         }
+        let mut transactions = Vec::new();
```
sgkim126:
```rust
let mut transactions = Vec::with_capacity(transaction_length);
```

HoOngEe: Fixed

sgkim126:
@HoOngEe Please resolve the conflict.

core/src/consensus/tendermint/mod.rs
```diff
@@ -184,23 +184,27 @@ impl TendermintInner {
...
     }
 
     /// Get the proposer of previous block to check new proposer is valid.
-    fn prev_block_proposer_idx(&self, height: Height) -> usize {
-        let prev_proposer = self.prev_block_of_height(height).author();
-        self.validators
-            .get_index_by_address(&self.prev_block_hash(), &prev_proposer)
-            .expect("The proposer must be in the validator set")
+    fn prev_block_proposer_idx(&self, height: Height) -> Option<usize> {
+        self.prev_block_header_of_height(height).and_then(|prev_header| {
```
majecty: Use `map` instead of the `and_then`.
We could remove the unnecessary `Some()`

core/src/consensus/tendermint/mod.rs
```diff
@@ -215,10 +219,12 @@ impl TendermintInner {
     }
 
     /// Find the designated for the given view.
-    fn view_proposer(&self, bh: &H256, height: Height, view: View) -> Address {
-        let proposer_nonce = self.prev_block_proposer_idx(height) + 1 + view;
-        ctrace!(ENGINE, "Proposer nonce: {}", proposer_nonce);
-        self.validators.get_address(bh, proposer_nonce)
+    fn view_proposer(&self, bh: &H256, height: Height, view: View) -> Option<Address> {
+        self.prev_block_proposer_idx(height).and_then(|prev_proposer_idx| {
```
majecty: Use `map` instead of `and_then`

sgkim126:
@HoOngEe please fix the conflicts.

core/src/consensus/tendermint/mod.rs
HoOngEe: Fixed

core/src/consensus/tendermint/mod.rs
HoOngEe: Fixed

sgkim126:
@HoOngEe test build failed.

PR #1295, opened 2019-02-21
Fix a bug that prevents creating a connection to a remote node

sgkim126:
It's because I queried the peer address right after I request a connection.
There can be no peer because creating a connection also non-blocking.


sgkim126:
@majecty It fixes the network problem.

PR #1296, opened 2019-02-21
Run sync tests on OSX

### LINKED ISSUE #1290: Creating a connection sometimes fails on OSX
`TcpStream::connect()` sometimes fails with "Socket is not connected (os error 57)".
It seems that it only happens on OSX.
Some e2e tests are skipped on Travis because of it. We should run the tests after fixing it.

sgkim126:
It fixes #1290


sgkim126:
https://github.com/CodeChain-io/codechain/pull/1295 would affect this issue.

PR #1298, opened 2019-02-21
Update ring to 0.14.6

HoOngEe:
Updated dependency "ring"


crypto/src/error.rs
```diff
@@ -28,6 +28,10 @@ quick_error! {
             cause(e)
             from()
         }
+        zeroIterations {
```
sgkim126: `ZeroIterations`

HoOngEe: Fixed

crypto/src/error.rs
```diff
@@ -28,6 +28,10 @@ quick_error! {
             cause(e)
             from()
         }
+        ZeroIterations {
+            description("Iterations' value should not be zero")
+            from()
```
sgkim126: I think this `from()` does nothing. Would you check?

HoOngEe: The code works without from(), I removed it.

PR #1301, opened 2019-02-22
Support cidr

HoOngEe:
Change FilterEntry's address type from IpAddr to IpCidr.
This change requires to change the form of the whitelist addresses used in making configuartions.

Previously to allow the addresses ranging from 141.223.175.0 to 141.223.175.255 using whitelist, we had to add "141.223.175.0" to the whitelist.
After the change,  the explicit suffix "141.223.175.0/24" is needed. An ipAddr input would lead filter to block that single IP Address. RPCs related to the network whitelist and Blacklist should also be changed.


rpc/src/v1/traits/net.rs
```diff
@@ -48,16 +49,16 @@ build_rpc_trait! {
         fn get_established_peers(&self) -> Result<Vec<SocketAddr>>;
 
         #[rpc(name = "net_addToWhitelist")]
-        fn add_to_whitelist(&self, IpAddr, Option<String>) -> Result<()>;
+        fn add_to_whitelist(&self, IpCidr, Option<String>) -> Result<()>;
```
sgkim126: https://github.com/CodeChain-io/codechain/blob/master/spec/JSON-RPC.md
Would you update the spec?

HoOngEe: I created another pr#1321

PR #1303, opened 2019-02-25
Forbid non-active user be registered as special users

sgkim126:
The active account means the account that has ever had the balance. In
other words, it's the account that has non-zero balance or seq. This
term is introduced in this commit but the concept is already used. The
public key of the account that has non-zero balance or seq cannot be
used as a regular key.

This commit forbids the non-active account becoming the shard users,
shard owners, the administrator of the asset or the approver. If it's
allowed the regular account to be special users, criteria for which of
the master account and the regular account becomes a special user is not
intuitive. It's a restriction to make the logic simple.

### DIFF
--- a/state/src/impls/test_helper.rs
+++ b/state/src/impls/test_helper.rs
@@ -481,6 +481,11 @@ macro_rules! set_top_level_state {
 
         set_top_level_state!($state, [$($x),*]);
     };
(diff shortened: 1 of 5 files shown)


sgkim126:
@Kais-DkM Because of this restriction, you should design the console sends 1 quark when you set a none used address as shard user, shard owner, asset administrator, or asset approver.

PR #1320, opened 2019-02-27
Modify signer index calculation method in ConsensusMessage

HoOngEe:
Though we changed the proposer calculation method, this change
did not applied to the `new_proposal` method in `ConsensusMessage`.
Now the change is applied.


core/src/consensus/tendermint/mod.rs
```diff
@@ -1668,14 +1673,23 @@ impl TendermintExtension {
...
 
             let num_validators = tendermint.validators.count(&parent_hash);
-            let message = match ConsensusMessage::new_proposal(signature, num_validators, &header_view) {
-                Ok(message) => message,
-                Err(err) => {
-                    cdebug!(ENGINE, "Invalid proposal received: {:?}", err);
+            let prev_proposer_idx = match tendermint.prev_block_proposer_idx(number as Height) {
+                Some(idx) => idx,
+                None => {
+                    cdebug!(ENGINE, "Prev block proposer does not exist for height {}", number);
```
majecty: Please use `cwarn!`

core/src/consensus/tendermint/mod.rs
```diff
@@ -1668,14 +1673,23 @@ impl TendermintExtension {
...
                     return
                 }
             };
 
+            let message =
+                match ConsensusMessage::new_proposal(signature, num_validators, &header_view, prev_proposer_idx) {
+                    Ok(message) => message,
+                    Err(err) => {
+                        cdebug!(ENGINE, "Invalid proposal received: {:?}", err);
```
majecty: Please use cwarn!

HoOngEe: Fixed

core/src/consensus/tendermint/mod.rs
HoOngEe: Fixed

PR #1323, opened 2019-02-27
Reduce tendermint messages

majecty:


core/src/consensus/tendermint/mod.rs
```diff
@@ -924,6 +933,11 @@ impl TendermintInner {
             return
         }
 
+        if token == ENGINE_TIMEOUT_BROADCAST_STEP_STATE && self.votes_received_changed {
+            self.votes_received_changed = false;
+            self.broadcast_state(&self.vote_step(), self.proposal, self.last_lock, self.votes_received);
```
sgkim126: You should return here.

majecty: I added it. Thanks.

Issue #1297, opened 2019-02-21
Stabilize tendermint e2e tests


sgkim126:
https://github.com/CodeChain-io/codechain/blob/5afcd730f25162a61f9112ac6f9394869b4eef90/test/src/e2e.long/tendermint.test.ts#L39
Currently, Travis doesn't run tendermint tests because of a timeout.
The timeout also happens on the local machine too. It should be stabilized.
I suspected that it failed because of the network issue. But it still fails after changing the network protocol.
It seems that there are other issues.

sgkim126:
It seems that increasing timeout fixes it. I'll do more tests.

PR #1326, opened 2019-02-27
Fix tendermint e2e tests

sgkim126:
It closes https://github.com/CodeChain-io/codechain/issues/1297


sgkim126:
@majecty I changed the scripts a little. Would you check it works as your intention?

PR #1340, opened 2019-03-04
Compress the body response in the block sync

majecty:


sync/src/block/message/response.rs
```diff
@@ -32,10 +34,23 @@ impl Encodable for ResponseMessage {
...
+                let mut uncompressed = {
+                    let mut inner_list = RlpStream::new_list(bodies.len());
+                    bodies.iter().for_each(|body| {
+                        inner_list.append_list(body);
+                    });
+                    inner_list.out()
+                };
+
+                let compressed = {
+                    // TODO: Cache the Encoder object
```
sgkim126: Do you have a reason to cache it?

sync/src/block/message/response.rs
```diff
@@ -67,8 +82,28 @@ impl ResponseMessage {
...
+                let uncompressed = {
+                    // TODO: Cache the Decoder object
+                    let mut snappy_decoder = snap::Decoder::new();
+                    snappy_decoder.decompress_vec(&compressed).map_err(|err| {
+                        cwarn!(SYNC, "Decompression failed while decoding a body response: {}", err);
+                        DecoderError::Custom("Invalid compression format")
+                    })?
+                };
+
+                let bodies_rlp = UntrustedRlp::new(&uncompressed);
```
sgkim126: How about naming it `uncompressed_rlp`?

sync/src/block/message/response.rs
majecty: The document recommended it.
https://docs.rs/snap/0.2.5/snap/struct.Encoder.html

sync/src/block/message/response.rs
majecty: Good.

sync/src/block/message/response.rs
majecty: The Encoder uses a big size buffer in heap.
Caching the Encoder object will reduce the buffer allocation/deallocation.

sgkim126: Actually, that's what I worried.
If you cache the encoder, the memory will not be deallocated. It can make the extension hold very large unused memory.

PR #1346, opened 2019-03-05
Fix the sync bug that occurs in Corgi

majecty:


core/src/consensus/tendermint/mod.rs
```diff
@@ -226,11 +226,11 @@ impl TendermintInner {
     }
 
     /// Find the designated for the given view.
-    fn view_proposer(&self, bh: &H256, height: Height, view: View) -> Option<Address> {
-        self.prev_block_proposer_idx(height).map(|prev_proposer_idx| {
+    fn view_proposer(&self, prev_block_hash: &H256, _height: Height, view: View) -> Option<Address> {
```
sgkim126: I think you can remove `height`.

test/src/e2e.long/tendermint.test.ts
```diff
@@ -255,9 +255,9 @@ describe("Tendermint ", function() {
         await promiseExpect.shouldFulfill(
             "wait blocknumber",
             Promise.all([
-                nodes[0].waitBlockNumber(startHeight + 1),
```
sgkim126: What is the purpose of this test?
Does it test the block sync extension works?
If then, I think it should be changed to the below scenario.
1. Connect three nodes.
2. Generate blocks.
3. Connect one more node.
4. Check whether the last node follows the synchronization.

majecty: I want to check two things.

1. Whether the block extension is working or not working.
2. Whether the node which received blocks using block extension could participate or could not participate in the Tendermint consensus.

core/src/consensus/tendermint/mod.rs
majecty: Thx!

PR #1349, opened 2019-03-06
Make TimerApi::set_handler receive Weak instead of &Arc

sgkim126:
A set_handler method downgrades the handler to Weak, but it doesn't use the handler.


sgkim126:
@foriequal0 ping

Issue #1327, opened 2019-02-27
Create a dedicated thread per network extension


sgkim126:
Currently, the network service creates 4 threads and all extensions share these threads. And the network message encryption/decryption and signature checking are also done in these threads.
This design has three problems.
1. If some extension is CPU heavy and holds the thread for a long time, other extensions cannot receive the message at all.
2. Events in the same extension can occur simultaneously. It makes debugging the extension hard.
3. The extension should be `Send` + `Sync`. It makes implementing the extension hard.

So I want to create a thread per extension and use a channel to pass events to an extension.

sgkim126:
https://github.com/CodeChain-io/codechain/pull/1352 fixed it.

PR #1352, opened 2019-03-07
Create a dedicated thread per network extension

sgkim126:
It fixed https://github.com/CodeChain-io/codechain/issues/1327

### DIFF
--- a/network/src/client.rs
+++ b/network/src/client.rs
@@ -18,7 +18,7 @@ use std::collections::HashMap;
 use std::sync::{Arc, Weak};
 
 use cio::IoChannel;
-use ctimer::{TimerApi, TimerLoop, TimerToken};
+use ctimer::{TimeoutHandler, TimerApi, TimerLoop, TimerToken};
 use parking_lot::RwLock;
 use time::Duration;
 
@@ -81,6 +81,16 @@ impl Api for ClientApi {
     }
 }
 
+impl TimeoutHandler for ClientApi {
+    fn on_timeout(&self, token: TimerToken) {
+        let extension_guard = self.extension.read();
+        let some_extension = extension_guard.as_ref().expect("Extension should be initialized");
+        if let Some(extension) = some_extension.upgrade() {
+            extension.on_timeout(token);
+        }
+    }
+}
+
 pub struct Client {
     extensions: RwLock<HashMap<&'static str, Arc<NetworkExtension>>>,
     p2p_channel: IoChannel<P2pMessage>,
@@ -106,7 +116,7 @@ impl Client {
         let name = extension.name();
         *api.extension.write() = Some(Arc::downgrade(&extension) as Weak<NetworkExtension>);
         api.timer.set_name(name);
-        api.timer.set_handler(Arc::downgrade(&extension));
+        api.timer.set_handler(Arc::downgrade(&api));
         extension.on_initialize();
         let trait_extension = Arc::clone(&extension) as Arc<NetworkExtension>;
         if extensions.insert(name, trait_extension).is_some() {
@@ -162,7 +172,7 @@ mod tests {
     use std::vec::Vec;
 
     use cio::IoService;
-    use ctimer::{TimeoutHandler, TimerLoop};
+    use ctimer::TimerLoop;
     use parking_lot::Mutex;
     use time::Duration;
 
@@ -246,9 +256,7 @@ mod tests {
             let mut callbacks = self.callbacks.lock();
             callbacks.push(Callback::Message);
         }
-    }
 
-    impl TimeoutHandler for TestExtension {
         fn on_timeout(&self, _timer_id: usize) {
             let mut callbacks = self.callbacks.lock();
             callbacks.push(Callback::Timeout);
--- a/network/src/client.rs
+++ b/network/src/client.rs
@@ -15,51 +15,50 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 use std::collections::HashMap;
-use std::sync::{Arc, Weak};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::thread::{spawn, JoinHandle};
 
 use cio::IoChannel;
+use crossbeam_channel as crossbeam;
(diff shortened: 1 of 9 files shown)


network/src/extension.rs
```diff
@@ -65,4 +65,6 @@ pub trait Extension: TimeoutHandler + Send + Sync {
     fn on_node_removed(&self, _node: &NodeId) {}
 
     fn on_message(&self, _node: &NodeId, _message: &[u8]) {}
+
+    fn on_timeout(&self, _token: TimerToken) {}
```
foriequal0: I think we could remove default impl for `TimeoutHandler`

network/src/client.rs
```diff
@@ -161,8 +129,21 @@ impl Client {
         extensions.iter().map(|(name, extension)| (name.to_string(), extension.versions().to_vec())).collect()
     }
 
-    define_method!(on_node_added; id, &NodeId; version, u64);
-    define_broadcast_method!(on_node_removed; id, &NodeId);
+    pub fn on_node_removed(&self, id: &NodeId) {
+        let extensions = self.extensions.read();
+        for (_, ref extension) in extensions.iter() {
```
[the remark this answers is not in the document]
sgkim126: Thank you. I'll apply the review on #1353

network/src/client.rs
```diff
@@ -161,8 +129,21 @@ impl Client {
...
+            extension.on_node_removed(id);
+        }
+    }
+
+    pub fn on_node_added(&self, name: &str, id: &NodeId, version: u64) {
+        let extensions = self.extensions.read();
+        if let Some(ref extension) = extensions.get(name) {
```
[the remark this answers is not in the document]
sgkim126: ditto

network/src/extension.rs
sgkim126: ditto
https://github.com/CodeChain-io/codechain/pull/1353/commits/c0aa454844ec2f72f75d367689b467d5fe1aa842

network/src/client.rs
```diff
@@ -104,24 +124,95 @@ impl Client {
...
+                rx,
+            )
         };
         let extension = Arc::new(factory(Arc::clone(&api) as Arc<Api>));
         let name = extension.name();
-        *api.extension.write() = Some(Arc::downgrade(&extension) as Weak<NetworkExtension>);
+
+        let cloned_extension = Arc::clone(&extension);
+
+        let extension_initialized: Arc<(Mutex<bool>, Condvar)> = Default::default();
```
foriequal0: I think it would be better rather than using tuple.
1. give it a proper type.
```rust
struct Barrier {
   flag: Mutex<bool>,
   condvar: Condvar
}
impl Barrier {
  fn wait(&self) { ... }
  fn trigger(&self) { ... }
}

let init_barrier = Arc::new(InitCondvar {
   flag: Mutex::new(false),
   condvar: Condvar::new(),
});
```

2. ExtensionMessage::Initialize.
```rust
        let join = Some(spawn(move || {
            if let Ok(ExtensionMessgae::Initialize) = rx.recv() {
                cloned_extension.on_initialize();
            } else {
                 unreachable!("Must be initialized first");
            }
            loop {
                crossbeam::select! {
                    recv(rx) -> msg => {
                        match msg {
...
```

network/src/client.rs
```diff
@@ -136,36 +227,75 @@ impl Client {
...
+        extensions
+            .iter()
+            .map(
+                |(
+                    name,
+                    Extension {
+                        versions,
+                        ..
+                    },
+                )| (name.to_string(), versions.clone()),
```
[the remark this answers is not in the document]
sgkim126: Nice idea. I'll fix it.

network/src/client.rs
sgkim126: 1. I don't want to create a new structure for the local only and used once type.
If a tuple repeats an item of the same type, it may be better to give it a clear name even though it's used on local only. But they have different type and the type shows the purpose explicitly in this case.

sgkim126: And FYI, `Barrier` is not a good name for this kind of structure.
`Barrier` is a well-known synchronization primitive for thread coordination.
It makes threads wait until all participating threads reach some point.

https://en.cppreference.com/w/cpp/experimental/barrier
https://doc.rust-lang.org/std/sync/struct.Barrier.html
https://linux.die.net/man/3/pthread_barrier_wait

sgkim126: I changed it to use a channel.
The crossbeam channel guarantees that the receiver receives sent messages after the sender disconnects the channel.

network/src/client.rs
```diff
@@ -136,36 +222,64 @@ impl Client {
...
+        extensions.iter().map(|(name, extension)| (name.to_string(), extension.versions.clone())).collect()
     }
 
     pub fn on_node_removed(&self, id: &NodeId) {
         let extensions = self.extensions.read();
-        for (_, extension) in extensions.iter() {
-            extension.on_node_removed(id);
+        for (
+            name,
+            Extension {
```
foriequal0: `extension.sender` would be more concise.

network/src/client.rs
```diff
@@ -136,36 +222,64 @@ impl Client {
...
+                cwarn!(NETAPI, "{} cannot remove {}: {:?}", name, id, err);
+            }
         }
     }
 
     pub fn on_node_added(&self, name: &str, id: &NodeId, version: u64) {
         let extensions = self.extensions.read();
-        if let Some(extension) = extensions.get(name) {
-            extension.on_node_added(id, version);
+        if let Some(Extension {
```
foriequal0: `extension.sender` would be more concise(2)

network/src/client.rs
```diff
@@ -136,36 +222,64 @@ impl Client {
...
+            }
         } else {
             cdebug!(NETAPI, "{} doesn't exist.", name);
         }
     }
 
     pub fn on_message(&self, name: &str, id: &NodeId, data: &[u8]) {
         let extensions = self.extensions.read();
-        if let Some(extension) = extensions.get(name) {
+        if let Some(Extension {
```
foriequal0: `extension.sender` would be more concise(3)

sgkim126:
@foriequal0 ping

PR #1354, opened 2019-03-07
Fix Tendermint liveness problem

majecty:


core/src/consensus/tendermint/mod.rs
```diff
@@ -771,8 +771,9 @@ impl TendermintInner {
         }
         let prev_proposer_idx = self.block_proposer_idx(*header.parent_hash()).expect("Prev block must exists");
 
-        let vote_step =
-            VoteStep::new(header.number() as Height, consensus_view(&header).expect("I am proposer"), Step::Propose);
+        assert_eq!(self.view, consensus_view(&header).expect("I am proposer"));
```
sgkim126: Let's make it `debug_assert_eq`

core/src/consensus/tendermint/mod.rs
```diff
@@ -1152,19 +1153,13 @@ impl TendermintInner {
...
-        if self.is_signer_proposer(&parent_hash) {
-            let vote_info = message_info_rlp(vote_step, Some(hash));
-            let signature = self.sign(blake256(&vote_info)).expect("I am proposer");
-            self.extension().broadcast_proposal_block(signature, block.into_inner());
-        } else if let Some(signature) = self.votes.round_signature(&vote_step, &hash) {
-            self.extension().broadcast_proposal_block(signature, block.into_inner());
-        } else {
-            cwarn!(ENGINE, "There is a proposal but does not have signature {:?}", vote_step);
-        }
+        assert_eq!(true, self.is_signer_proposer(&parent_hash));
```
```rust
assert!(self.is_signer_proposer(&parent_hash));
```

core/src/consensus/tendermint/mod.rs
```diff
@@ -474,6 +482,33 @@ impl TendermintInner {
         }
     }
 
+    fn locked_proposal_block(&self, locked_view: View) -> Result<encoded::Block, String> {
+        let vote_step = VoteStep::new(self.height, locked_view, Step::Propose);
+        let locked_proposal_hash = self.votes.get_block_hashes(&vote_step).first().cloned();
+        let locked_proposal_hash = match locked_proposal_hash {
```
sgkim126: You can express it as below.
```rust
let locked_proposal_hash = locked_proposal_hash.ok_or_else(|| {
    self.request_proposal(self.height, locked_view);
    format!(
        "Have a lock on {}-{}, but do not received a locked proposal",
        self.height, locked_view
    )
})?;
```

core/src/consensus/tendermint/mod.rs
```diff
@@ -474,6 +482,33 @@ impl TendermintInner {
...
+                self.request_proposal(self.height, locked_view);
+                return Err(format!(
+                    "Have a lock on {}-{}, but do not received a locked proposal",
+                    self.height, locked_view
+                ))
+            }
+            Some(hash) => hash,
+        };
+
+        let locked_proposal_block = match self.client().block(&BlockId::Hash(locked_proposal_hash)) {
```
```rust
let locked_proposal_block = match self.client().block(&BlockId::Hash(locked_proposal_hash)).ok_or_else(|| {
    format!("Have a lock on {}-{}, and received the locked proposal, but the proposal is not imported yet.", self.height, locked_view)
})?;
```

PR #1359, opened 2019-03-08
Use an event channel to communicate with the block sync extension

sgkim126:


sgkim126:
@majecty The last two commits(14f7cf320db1d616d14a41351c66299a2b476252 and 22892b6d61f82c3e875002bcb77ef4a9574cf991) makes network extensions be used only in the extension thread.
But it makes the tendermint e2e tests fail. I think it's because there is a deadlock. Would you check it?

sgkim126:
@majecty 
I'll upload another PR to make CodeChain use an event channel for Tendermint extension. It needs some more refactoring because the current tendermint design is not suited to do it.

sync/src/block/extension.rs
```diff
@@ -690,8 +708,48 @@ impl Extension {
...
+pub struct BlockSyncSender(EventSender<Event>);
+
+impl From<EventSender<Event>> for BlockSyncSender {
+    fn from(sender: EventSender<Event>) -> Self {
+        BlockSyncSender(sender)
+    }
+}
+
+impl ChainNotify for BlockSyncSender {
+    fn new_headers(
```
sgkim126: @Kais-DkM Should ChainNotify be synchronous?
Is it okay not to wait for the result of the event?

Kais-DkM: `new_headers` and `new_blocks` creates all the lock they need.
It should be safe to call them asynchronously.

PR #1372, opened 2019-03-11
Name the network extension threads

sgkim126:


network/src/client.rs
```diff
@@ -146,43 +146,48 @@ impl Client {
...
-            init_receiver.recv().expect("The main thread must send one message");
-            cloned_extension.on_initialize();
-            loop {
-                crossbeam::select! {
-                    recv(rx) -> msg => {
-                        match msg {
-                            Ok(ExtensionMessage::NodeAdded(id, version)) => {
-                                cloned_extension.on_node_added(&id, version);
+        let join = Some(
+            Builder::new()
```
sgkim126: It changes `spawn(` to `Builder::new().name(format!("{}.ext", name)).spawn(` and add `unwrap()` because `spawn()` is `Bulder::new().spawn().unwrap()`.
Other changes are because of the formatter.

PR #1373, opened 2019-03-11
Cleanup tendermint module

sgkim126:


sgkim126:
It seems that the failed e2e tests are the same bug with https://github.com/CodeChain-io/codechain/pull/1375.

PR #1377, opened 2019-03-11
Crash when block/extension's data is corrupted

majecty:


sgkim126:
How about using `debug_assert` instead of `assert`?

PR #1380, opened 2019-03-12
Refactor the Tendermint module to run it on a single thread

sgkim126:


core/src/consensus/tendermint/mod.rs
```diff
@@ -1197,29 +1471,50 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
...
+            .unwrap();
+        receiver.recv().unwrap()
     }
 
     /// Called when the node is the leader and a proposal block is generated from the miner.
     /// This writes the proposal information and go to the prevote step.
     fn proposal_generated(&self, sealed_block: &SealedBlock) {
-        let mut guard = self.inner.lock();
-        guard.proposal_generated(sealed_block);
+        // TODO
```
majecty: What is the purpose of the TODO?

core/src/consensus/tendermint/mod.rs
```diff
@@ -1273,74 +1568,101 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
...
+            })
+            .unwrap();
+        receiver.recv().unwrap()
     }
 
     fn set_signer(&self, ap: Arc<AccountProvider>, address: Address) {
-        let mut guard = self.inner.lock();
-        guard.set_signer(ap, address)
+        self.has_signer.store(true, AtomicOrdering::SeqCst);
+        // TODO
```
majecty: What is the purpose of the TODO?

core/src/consensus/tendermint/mod.rs
sgkim126: Sorry. It should be removed. I'll update the commit.

core/src/consensus/tendermint/mod.rs
sgkim126: ditto

PR #1384, opened 2019-03-13
Add rust-toolchain to set the default rust version

sgkim126:


sgkim126:
@majecty @HoOngEe @foriequal0 I added `rust-toolchain` file. From now on you should change this file and travis.yml when you want to upgrade the version of the rust.

sgkim126:
Please run `rustup update 1.32.0` when Cargo doesn't work after this patch.

PR #1385, opened 2019-03-13
Remove Send/Sync marker from the network extension and api.

sgkim126:


sgkim126:
It removes unnecessary locks and moves on_initialize callback to a constructor.
@majecty Would you check the tendermint extension?
@Kais-DkM Would you check the sync extension?

Kais-DkM:
There are some other components that will try to call functions inside network extensions.
(ex. ChainNotify, TimerHandler, etc.)
Aren't they running in a different thread from network workers?
I think locks shouldn't be removed if that's the case.

Kais-DkM:
Also, if you're going to apply `tokio` later, I'd recommend removing only `Sync`, not `Send` **and** `Sync`.
Default runtime of `tokio` uses thread stealing queue, and it requires underlying workers to be `Send`.

sgkim126:
> There are some other components that will try to call functions inside network extensions.
> (ex. ChainNotify, TimerHandler, etc.)
> Aren't they running in a different thread from network workers?
> I think locks shouldn't be removed if that's the case.

There is no way to call the function of network extensiosn from the other threads.
https://github.com/CodeChain-io/codechain/pull/1385/commits/5cec2721a0f992992e81abfa020a28697c84dc5c

sgkim126:
> Also, if you're going to apply `tokio` later, I'd recommend removing only `Sync`, not `Sync` **and** `Send`.
> Default runtime of `tokio` uses thread stealing queue, and it requires underlying workers to be `Send`.

The network threads, which have a socket and receive/send messages, don't call the extension directly. It uses a channel to invoke an event.

Kais-DkM:
> There is no way to call the function of network extensiosn from the other threads.
> [5cec272](https://github.com/CodeChain-io/codechain/commit/5cec2721a0f992992e81abfa020a28697c84dc5c)

Oh, are the event callbacks called only from network thread? It seems to be safe in that case.

Kais-DkM:
> The network threads, which have a socket and receive/send messages, don't call the extension directly. It uses a channel to invoke an event.

I think you'll be using [tokio::sync::mpsc](https://docs.rs/tokio/0.1.17/tokio/sync/mpsc/index.html) for sending events when using `tokio`.
`Receiver` of `mpsc` are `Stream`s, so event listeners would be residing in the tokio runtime in most cases.
I think they should be `Send` in that case.

sgkim126:
> > The network threads, which have a socket and receive/send messages, don't call the extension directly. It uses a channel to invoke an event.
> 
> I think you'll be using [tokio::sync::mpsc](https://docs.rs/tokio/0.1.17/tokio/sync/mpsc/index.html) for sending events when using `tokio`.
> `Receiver` of `mpsc` are `Stream`s, so event listeners would be residing in the tokio runtime in most cases.
> I think they should be `Send` in that case.

What should be `Send`?
The message should be `Send`. But it doesn't remove `Send` from the message.

It removes `Send` from the extension and the api.
Both of them do not need to be sent to other threads. The extension is a consumer of the message from the network threads and the api is a producer of the message to the network threads.

Please don't confuse the message and the extension and the api.

Kais-DkM:
Let me explain my concerns a little bit more.
If we're going to use `tokio`, they would be sent through `mpsc`, and read with `Receiver` interface.
Maybe the code for that will be something like this:
```rust
pub struct Extension {
    receiver: Receiver,
    ...
}

impl Stream for Extension {
    type Item = Event;
    type Error = ();

    fn poll(&mut self) -> Poll<Option<()>, ()> {
        let event = try_ready!(self.receiver.poll().map_err(|_| ()));

        do_something(event);
        Ok(Async::Ready(Some(())))
    }
}
```

and traits required by `spawn` of default runtime is
```rust
Future<Item = (), Error = ()> + 'static + Send, 
```

According to `Send` implementations stated in [here](https://docs.rs/tokio/0.1.17/tokio/prelude/future/struct.Then.html#impl-Send), every future involved seems to be required to implement `Send`.
I think `Extension`, which will be implementing `Stream` will need `Send` trait if it is going to listen to messages sent from tokio runtime, since it needs to be spawned by the same runtime.

Please let me know there are any errors in this logic.

sgkim126:
No. Extension will not implement a `Stream`.

Kais-DkM:
Then how is it going to read messages from `mpsc`?

sgkim126:
There is no need to make the consumer of the `Receiver` derive `Stream`.
`Receiver` already derives `Stream` and the extension only need to consume it.

Kais-DkM:
And to consume it, it needs to be spawned and participate in event loop of runtime.
Or do you mean that extensions will be consuming them outside of the tokio system?

sgkim126:
> And to consume it, it needs to be spawned and participate in event loop of runtime.
> Or do you mean that extensions will be consuming them outside of the tokio system?

Doesn't matter. The executor of Tokio is a trait. There is no need to use the thread pool. You can use a single-threaded executor.

Kais-DkM:
Oh, you're going to use single-threaded executor.
I thought we'd be using thread stealing queue based executor.
It makes sense then.
But please review once more if you really don't need a multi-threaded executor in this case.

sgkim126:
I'm not going to use a thread pool for extensions because of fairness.
All extension does a different thing. If you groups these kinds of workers in the same thread pool, the fairness always become a problem.
A future is not a useful abstraction to solve it, so not using a thread pool is the best solution in most case.

PR #1391, opened 2019-03-18
Compress the payload of the transaction sync

majecty:


majecty:
@sgkim126 I fixed the integration test.

PR #1393, opened 2019-03-18
Ignore known txs in the Transaction Sync

majecty:


sync/src/transaction/extension.rs
```diff
@@ -30,14 +30,14 @@ use super::message::Message;
 
 const BROADCAST_TIMER_TOKEN: TimerToken = 0;
 const BROADCAST_TIMER_INTERVAL: i64 = 1000;
-const MAX_HISTORY_SIZE: usize = 100;
+const MAX_HISTORY_SIZE: usize = 100_000;
```
sgkim126: Make the other commit for it.

majecty: I divided the commit.

PR #1405, opened 2019-03-21
Log JSONRPC requests

majecty:


codechain/rpc_apis.rs
```diff
@@ -55,7 +55,7 @@ impl ApiDependencies {
     }
 }
 
-pub fn setup_rpc(mut handler: MetaIoHandler<()>) -> MetaIoHandler<()> {
+pub fn setup_rpc(mut handler: MetaIoHandler<(), impl Middleware<()>>) -> MetaIoHandler<(), impl Middleware<()>> {
```
sgkim126: It means that the input and output can be different types which implement MiddleWare.
```rust
pub fn setup_rpc<M: MiddleWare<()>>(mut handler: MetaIoHandler<(), M>) -> MetaIoHandler<(), M> {
```

majecty: I fixed it.

PR #1417, opened 2019-03-28
Modified the chain to include only successful transactions

sgkim126:


test/src/e2e.long/timelock.test.ts
```diff
@@ -114,9 +115,16 @@ describe("Timelock", function() {
             shardId: asset.shardId,
             recipient: await node.createP2PKHAddress()
         });
-        const results1 = await node.sendAssetTransaction(failedTx);
-        expect(results1!.length).to.equal(1);
-        expect(results1![0]).to.be.false;
+        const blockNumber = await node.sdk.rpc.chain.getBestBlockNumber();
```
sgkim126: The current mempool stores the failed reason on block commit, and it doesn't commit the block has no transactions.
Because of it, you cannot get the error hint if you sent the one failed transaction.
To resolve this problem, I made the node stop sealing and sending one more transaction which succeeds.
You don't worry about this because it will not happen in the real network.
But I'll enhance the mempool to store the failed reason even it doesn't commit the block.

PR #1423, opened 2019-03-29
Add spec for the minimum fee

sgkim126:


spec/Transaction.md
```diff
@@ -2,6 +2,8 @@
 
 Only a platform account can generate a transaction.
 The transaction generator pays the transaction fees.
+The transactions that cost less than the [minimum fee](Minimum-Fee.md) are rejected.
```
ScarletBlue: Transactions that cost less than the [minimum fee](Minimum-Fee.md) are rejected.

spec/Transaction.md
```diff
@@ -2,6 +2,8 @@
...
+The minumum fee is different for each transaction types.
```
ScarletBlue: The minumum fee is different for each transaction type.

PR #1424, opened 2019-03-29
Add an option to decide whether Miner allows CreateShard transaction

sgkim126:


sgkim126:
@majecty We decided to disable CreateShard until we decided the details.

PR #1435, opened 2019-04-03
Upgrade sdk used by e2e tests

sgkim126:


sgkim126:
It will be landed with https://github.com/CodeChain-io/codechain/pull/1442

PR #1440, opened 2019-04-03
Add an RPC to get the recent network usage

sgkim126:


spec/JSON-RPC.md
```diff
@@ -2245,6 +2246,38 @@ No parameters
 
 [Back to **List of methods**](#list-of-methods)
 
+## net_recentNetworkUsage
+Gets the recent network usage.
```
ScarletBlue: Gets the recent network usage and returns an object.

spec/JSON-RPC.md
```diff
@@ -2245,6 +2246,38 @@ No parameters
...
+It returns an object.
+The key of the object is a meaningful separation, but what is used as a key depends on the implementation.
```
ScarletBlue: I think this part sounds a bit awkward. Let's try fixing it.

spec/JSON-RPC.md
```diff
@@ -2245,6 +2246,38 @@ No parameters
...
+The value of the object is the size of bytes that the node sent in the recent period.
+The exact timespan of the recent is also an implantation dependent.
```
ScarletBlue: What is implantation dependent?

sgkim126: Typo. implementation dependent

spec/JSON-RPC.md
sgkim126: I'll change the next sentence to 'The return type is an object.'

spec/JSON-RPC.md
sgkim126: I changed it to `The key of the object is a string, but what the keys are depend on the implementation.`

sgkim126:
I split the commit into two commits. I'll merge the implementation first for Corgi network and update the spec later.

Issue #1431, opened 2019-04-01
Implement an RPC to get the minimum cost of the transaction


sgkim126:
# chain_getMinTransactionFee
Gets the min cost of the transaction.
It returns null if the first parameter is an invalid transaction type or the given block number is larger than the current best block.

### Params 
 1. transaction type - `string`
 2. block number - `number` | `null`

### Returns
`number` | `null`

Errors: `Invalid Params`

### Request Example
```
  curl \
    -H 'Content-Type: application/json' \
    -d '{"jsonrpc": "2.0", "method": "chain_getMinTransactionFee", "params": ["pay", 3], "id": 7}' \
    localhost:8080
```

### Response Example
```
{
  "jsonrpc":"2.0",
  "result":100,
  "id":7
}
```

sgkim126:
https://github.com/CodeChain-io/codechain/pull/1451 closed it

PR #1451, opened 2019-04-05
Implement chain_getMinTransactionFee rpc

HoOngEe:
Implemented `chain_getMinTransactionFee`.
It closes https://github.com/CodeChain-io/codechain/issues/1431


HoOngEe:
Now the string pattern matched in chain rpc function. I Implemented like this because the only usage of the type indicating string is unique especially in this rpc. But I think it's more appropriate for the matching between these two information to be immersed in `EngineInfo` trait method.
Any Opinion?

rpc/src/v1/impls/chain.rs
```diff
@@ -239,6 +239,29 @@ where
         Ok(self.client.block(&BlockId::Hash(block_hash)).map(|block| block.transactions_count()))
     }
 
+    fn get_min_transaction_fee(&self, action_type: String) -> Result<Option<u64>> {
+        let common_parameters = self.client.common_params();
+        Ok(match action_type.as_str() {
+            "MintAsset" => Some(common_parameters.min_asset_mint_cost),
```
sgkim126: I think it's better to use camelCase.
Because the action type of RPC uses camelCase.

sgkim126:
I think a string match in the RPC method is a fine option because it's only used here.

rpc/src/v1/impls/chain.rs
HoOngEe: I changed them to `camelCase`s

rpc/src/v1/traits/chain.rs
```diff
@@ -124,6 +124,10 @@ build_rpc_trait! {
          # [rpc(name = "chain_getBlockTransactionCountByHash")]
         fn get_block_transaction_count_by_hash(&self, H256) -> Result<Option<usize>>;
 
+        ///Gets the minimum transaction fee of the given name.
+         # [rpc(name = "chain_getMinTransactionFee")]
+        fn get_min_transaction_fee(&self, String) -> Result<Option<u64>>;
```
sgkim126: You need a block number.

HoOngEe: I added the block_number parameter.

PR #1453, opened 2019-04-08
Add a command that prints the commit hash

majecty:
Previous result of the `codechain -v`:

  `codechain 0.1.0`

New result of the `codechain -v`:

  `codechain 1.1.1 - f428ebcdc4cc29b69401a9b8c52ee25b03e4f6c2`

### DIFF
--- a/codechain/subcommand/mod.rs
+++ b/codechain/subcommand/mod.rs
@@ -27,6 +27,10 @@ pub fn run_subcommand(matches: &ArgMatches) -> Result<(), String> {
     match subcommand.name.as_str() {
         "account" => run_account_command(&subcommand.matches),
         "convert" => run_convert_command(&subcommand.matches),
+        "commit-hash" => {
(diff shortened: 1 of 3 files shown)


majecty:
I'm considering which name to use, "commit-hash" or "commitHash."

codechain/subcommand/mod.rs
```diff
@@ -27,6 +27,10 @@ pub fn run_subcommand(matches: &ArgMatches) -> Result<(), String> {
     match subcommand.name.as_str() {
         "account" => run_account_command(&subcommand.matches),
         "convert" => run_convert_command(&subcommand.matches),
+        "commitHash" => {
```
sgkim126: How about `commit-hash`?
I'm familiar with it rather than the pascal case for cli options.

majecty: I changed it.

PR #1457, opened 2019-04-08
Remove "husky" and "saluki"

majecty:


majecty:
Let's remove Husky and Saluki config files if it needed.

PR #1464, opened 2019-04-10
Add tests for multisig

sgkim126:


test/src/helper/spawn.ts
```diff
@@ -235,7 +235,7 @@ export default class CodeChain {
                         }) exited with code ${code}, ${signal}`
                     );
                 } else if (!this.isTestFailed) {
-                    unlinkSync(this.logPath);
+                    // unlinkSync(this.logPath);
```
joojis: What is this change for?

sgkim126: Fixed

PR #1469, opened 2019-04-11
Ignore requests of block propagation if there are too many requests

majecty:


Kais-DkM:
One way of distinguishing them would be filtering them by block height. If the peer is requesting block far away from the current block, it would be in initial block download stage with high possibility.

But I think it would be okay to use simple counting as long as our numbers fit in reasonable bandwidth. Maybe a better way would be using the number of headers/blocks inside message rather than the number of messages received.

majecty:
It's too old and not urgent.
Let's try it again when it became needed or when we can distinguish normal requests and malicious requests.

PR #1475, opened 2019-04-12
Improve e2e test about partial signature and multisig

HoOngEe:


test/src/e2e/multisig.test.ts
```diff
@@ -39,14 +42,47 @@ import {
...
+    const seq = await node.sdk.rpc.chain.getSeq(faucetAddress);
+    const pay = await node.sendPayTx({ seq, quantity: 1 });
+    const targetTxHash = await node.sendAssetTransaction(targetTx, {
+        seq: seq + 1
+    });
+
+    await node.sdk.rpc.devel.startSealing();
+    await node.waitBlockNumber(blockNumber + 1);
+
+    return {
```
sgkim126: How about checking whether the dummy transaction included in this function instead of returning it?

test/src/e2e/multisig.test.ts
```diff
@@ -113,8 +149,8 @@ describe("Multisig", function() {
                 );
 
             const hash = await node.sendAssetTransaction(transfer);
-            expect(await node.sdk.rpc.chain.containsTransaction(hash)).be.true;
-            expect(await node.sdk.rpc.chain.getTransaction(hash)).not.null;
+            expect(await chain.containsTransaction(hash)).be.true;
```
sgkim126: Why do we need these changes?

HoOngEe: I think these changes makes the code looks better. Is this sort of refactoring not desirable?

test/src/e2e/multisig.test.ts
HoOngEe: Ok. That seems better.

test/src/e2e/multisig.test.ts
sgkim126: Adding a method to `node` would be good.
But I think it doesn't enhance the readability because the reader must check what `chain` means.

HoOngEe: Ok. This change has been reverted.

sgkim126:
@HoOngEe Please run `yarn fmt`

HoOngEe:
Oh, I'll try to be more careful.

test/src/e2e/partialSignature.test.ts
```diff
@@ -18,16 +18,51 @@ import * as chai from "chai";
...
+    await node.waitBlockNumber(blockNumber + 1);
+
+    expect(await node.sdk.rpc.chain.containsTransaction(signedDummyTx.hash()))
+        .be.true;
+    return targetTxHash;
+}
 
 describe("Partial signature", function() {
     let node: CodeChain;
+    let key: Key;
```
sgkim126: The same comment as https://github.com/CodeChain-io/codechain/pull/1475#discussion_r275603525

test/src/e2e/multisig.test.ts
```diff
@@ -39,12 +42,39 @@ import {
...
+        seq: seq + 1
+    });
+
+    await node.sdk.rpc.devel.startSealing();
+    await node.waitBlockNumber(blockNumber + 1);
+
+    expect(await node.sdk.rpc.chain.containsTransaction(signedDummyTx.hash()))
+        .be.true;
+
+    return targetTxHash;
```
sgkim126: Is the caller should know the transaction hash?
If all callers use the hash only to check whether the transaction failed, I think it can be included in this function.

test/src/e2e/partialSignature.test.ts
```diff
@@ -346,6 +385,124 @@ describe("Partial signature", function() {
         });
     });
 
+    describe.only("dynamic inputs and outputs", function() {
```
sgkim126: ?

HoOngEe: I removed the "only" tag.

test/src/e2e/multisig.test.ts
HoOngEe: At first I was considering including the check routine for the transaction failure, but soon after I thought it is natural for all test functions included in "it" marked test cases to have explicit "expect" in their bodies.
I thought it looks consistent that all test cases marked with "it" end with the "expect" clauses.
Does the become better after including the check routines?

sgkim126: I think it's better to rename this function to `expectTransactionFail` and to check the failure internally.

HoOngEe: I renamed and checked the failure inside the function.

test/src/e2e/partialSignature.test.ts
sgkim126: ping?

HoOngEe: I've reverted this change.

test/src/e2e/partialSignature.test.ts
```diff
@@ -341,8 +331,7 @@ describe("Partial signature", function() {
                 });
                 const hash = await node.sendAssetTransaction(tx);
                 expect(await node.sdk.rpc.chain.getTransaction(hash)).not.null;
-                expect(await node.sdk.rpc.chain.containsTransaction(hash)).be
-                    .true;
+                expect(await node.sdk.rpc.chain.containsTransaction(hash)).be.true;
```
sgkim126: You shouldn't have changed it.
You reverted this change in https://github.com/CodeChain-io/codechain/pull/1475/commits/455792e7359dae672dc7d942b962700517adb70c

test/src/e2e/partialSignature.test.ts
```diff
@@ -166,6 +166,44 @@ describe("Partial signature", function() {
...
+                assetType,
+                shardId: 0,
+                quantity: 1000,
+                recipient: address1
+            });
+        await node.sdk.key.signTransactionBurn(tx, 0);
+        tx.addInputs(assets[0]);
+        await node.sdk.key.signTransactionInput(tx, 0);
+
+        expectTransactionFail(node ,tx);
```
sgkim126: Please run `yarm fmt`.

test/src/e2e/partialSignature.test.ts
HoOngEe: I ran `yarn fmt` in every commit not to make such changes.

sgkim126: It seems that it's not fixed yet.

HoOngEe: Now there is no such changes reverted in the middle of the commit in this pr. Sorry for your rechecking.

PR #1476, opened 2019-04-12
Allow struct/array to overflow

sgkim126:
How about enabling [`overflow_delimited_expr`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#overflow_delimited_expr) option?


majecty:
It looks better than before.

PR #1478, opened 2019-04-15
Update pivot to the best_header if the best_header is already imported

majecty:


sync/src/block/downloader/header.rs
```diff
@@ -74,9 +74,16 @@ impl HeaderDownloader {
     }
 
     pub fn update(&mut self, total_score: U256, best_hash: H256) -> bool {
-        if self.total_score < total_score {
+        if self.total_score <= total_score {
             self.total_score = total_score;
             self.best_hash = best_hash;
+
+            if self.client.block_header(&BlockId::Hash(best_hash)).is_some() {
```
majecty: I suspect that this statement caused the deadlock.

majecty: I found the below log when applied this commit to Corgi.
"cannot timeout 0: "SendError"

sgkim126: AFAIK, `SendError` is occurred only when the receiver of the channel had been closed because we used an unbounded channel. Please check whether the extension thread had been crashed.

majecty: Thanks to seulgi, I found that the error was caused by a log statement which is deleted in this PR.

PR #1479, opened 2019-04-15
Enhance tests about payment transaction

sgkim126:


test/src/e2e/pay.test.ts
```diff
@@ -16,6 +16,7 @@
 
 import { expect } from "chai";
 import "mocha";
+import { aliceAddress, faucetAddress } from "../helper/constants";
```
joojis: It seems `aliceAddress` is not used. Otherwise, LGTM.

sgkim126: Thank you. I fixed it.

PR #1480, opened 2019-04-15
Send header request to a random peer at every timeout

majecty:
Currently, nodes in a network request the same headers to multiple
nodes. To reduce network usage, send a header request to a random node at
every timeout.

### DIFF
--- a/sync/src/block/extension.rs
+++ b/sync/src/block/extension.rs
@@ -290,12 +290,15 @@ impl NetworkExtension<Event> for Extension {
                 let mut peer_ids: Vec<_> = self.header_downloaders.keys().cloned().collect();
                 peer_ids.shuffle(&mut thread_rng());
 
-                for id in peer_ids {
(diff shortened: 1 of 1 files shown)


sync/src/block/extension.rs
```diff
@@ -290,12 +290,15 @@ impl NetworkExtension<Event> for Extension {
                 let mut peer_ids: Vec<_> = self.header_downloaders.keys().cloned().collect();
                 peer_ids.shuffle(&mut thread_rng());
 
-                for id in peer_ids {
+                for id in peer_ids.clone() {
```
sgkim126: How about it?
```rust
for id in &peer_ids {
    let request = self.header_downloaders.get_mut(id).and_then(HeaderDownloader::create_request);
    if let Some(request) = request {
        self.send_header_request(id, request);
        break
    }
}
```

majecty: Good!

majecty:
@sgkim126 I fixed it.

PR #1481, opened 2019-04-15
Make `new_headers` in `chain_notify` called on all headers in chain

majecty:


core/src/client/importer.rs
```diff
@@ -391,7 +391,9 @@ impl Importer {
                 .block_header(&BlockId::Hash(*header.parent_hash()))
                 .expect("Parent of importing header must exist")
                 .decode();
-            if self.check_header(&header, &parent_header) {
+            if client.block_header(&BlockId::Hash(hash)).is_some() {
```
sgkim126: I am worried that this part will be a performance bottleneck because block_header method reads data from disk.

core/src/client/importer.rs
```diff
@@ -365,14 +365,19 @@ impl Importer {
     /// This is triggered by a message coming from a header queue when the header is ready for insertion
     pub fn import_verified_headers(&self, client: &Client) -> usize {
         let max_headers_to_import = 256;
+        let lock = self.import_lock.lock();
+        let headers = self.header_queue.drain(max_headers_to_import);
+        self.import_headers(headers, client, &lock)
+    }
 
-        let _lock = self.import_lock.lock();
+    fn import_headers(&self, headers: Vec<Header>, client: &Client, _importer_lock: &MutexGuard<()>) -> usize {
```
sgkim126: I found no reason to move headers.
How about changing it to take `&[Header]`?

sgkim126:
```rust
pub fn import_headers<'a>(&'a self, headers: impl IntoIterator<Item = &'a Header>, client: &Client, _importer_lock: &MutexGuard<()>) -> usize {
```
would be better.
It can remove the cloning of header in https://github.com/CodeChain-io/codechain/pull/1481/commits/a17aec5dd566b6289a4d5b397c848d39465337fc

core/src/client/client.rs
```diff
@@ -786,12 +786,14 @@ impl ImportSealedBlock for Client {
         let start = Instant::now();
         let route = {
             // scope for self.import_lock
-            let _import_lock = self.importer.import_lock.lock();
+            let import_lock = self.importer.import_lock.lock();
 
             let number = block.header().number();
             let block_data = block.rlp_bytes();
             let header = block.header().clone();
```
```rust
let header = block.header();

self.importer.import_headers(vec![header], self, &import_lock);

let route = self.importer.commit_block(block, header, &block_data, self)
```

core/src/client/importer.rs
```diff
@@ -96,11 +96,15 @@ impl Importer {
...
             let mut import_results = Vec::with_capacity(max_blocks_to_import);
 
-            let _import_lock = self.import_lock.lock();
+            let import_lock = self.import_lock.lock();
             let blocks = self.block_queue.drain(max_blocks_to_import);
             if blocks.is_empty() {
                 return 0
             }
+
+            let headers = blocks.iter().map(|block| block.header.clone()).collect();
```
sgkim126:
```rust
{
    let headers = blocks.iter().map(|block| &block.header);
    self.import_headers(headers, client, &import_lock);
}
```

core/src/client/importer.rs
Kais-DkM: Most of the headers accessed by this part will be close to the best block, and they will have been cached with high possibility.
I think we can revisit this if we see a real bottleneck from here.

sgkim126: Okay.

core/src/client/importer.rs
```diff
@@ -365,16 +369,21 @@ impl Importer {
     /// This is triggered by a message coming from a header queue when the header is ready for insertion
     pub fn import_verified_headers(&self, client: &Client) -> usize {
         let max_headers_to_import = 256;
+        let lock = self.import_lock.lock();
+        let headers = self.header_queue.drain(max_headers_to_import);
+        self.import_headers(headers, client, &lock)
+    }
 
-        let _lock = self.import_lock.lock();
+    pub fn import_headers(&self, headers: Vec<Header>, client: &Client, _importer_lock: &MutexGuard<()>) -> usize {
```
Kais-DkM: Do we need to pass a lock to this function?

sgkim126: To make the caller not forget to hold the `import_lock`?

Kais-DkM: How about using `ReentrantMutex` then?

Kais-DkM:
AFAIK, our implementation didn't need this feature because there were no cases where the blocks are imported before their headers are imported.
Does the tendermint extension behave differently?

core/src/client/importer.rs
sgkim126: I want to avoid to use the recursive lock if possible.
http://www.zaval.org/resources/library/butenhof1.html

Kais-DkM: Sounds reasonable.

majecty:
@Kais-DkM  When we use the PoW consensus, a block's miner does not get the header that the miner created in the `new_headers` in `chain_notify` function.

majecty:
@sgkim126 I fixed them.

PR #1494, opened 2019-04-24
Fix .dockerignore file to get commit hash successfully from docker image clients.

### LINKED ISSUE #1443: commitHash RPC doesn't work on container images.
Images in https://hub.docker.com/r/kodebox/codechain/tags don't return `commitHash` RPC properly.

```
curl     -H 'Content-Type: application/json'     -d '{"jsonrpc": "2.0", "method": "commitHash", "params": [], "id": null}'     localhost:8080
{"jsonrpc":"2.0","result":"","id":null}
```
It should return a git hash but an empty string is returned.

I thought that it's the problem of `vergen`(https://github.com/rustyhorde/vergen) because it uses system installed git binary not the git2 library. https://github.com/rustyhorde/vergen/issues/15
But Dockerfile installs git. https://github.com/CodeChain-io/codechain/blob/c9ec8b3056f1bf5a98473ce1ac50ff3b783f250b/docker/ubuntu/Dockerfile#L10

And I suspected that `strip` removed the docker hash. https://github.com/CodeChain-io/codechain/blob/c9ec8b3056f1bf5a98473ce1ac50ff3b783f250b/docker/ubuntu/Dockerfile#L37
But when I run the binary which is stripped manually, `commitHash` RPC works fine.

HoOngEe:
It closes #1443

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,3 @@
-/.git/
 /.idea/
 /target/
 /keys/
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,5 @@
 /keys/
 /db/
 /docker/
+/keystore.db
+/test/


sgkim126:
I think it's better to ignore the test directory also.
It would not be used to build the docker image.

HoOngEe:
Ok, the test directory is also ignored.

sgkim126:
LGTM

PR #1495, opened 2019-04-25
Remove the cli optional argument "no-miner"

### LINKED ISSUE #1492: Change default options to no-miner
Most of the nodes would not be the miner.
Let's change the default option.

HoOngEe:
Now the default value of mining.disable is true and could be set to false not through the no-miner option but through not giving both author and engine signer options.
It closes #1492 

As the `no-miner` option disappears, previous `no-miner` behavior is mimicked implicitly. If one wants to execute the codechain with PoW type engine but doesn't specify the author then it implicitly indicates `no-miner`. This is same for PBFT or PoA type engine if one doesn't specify the engine signer.
In addition, the error message has changed. If one wants PoW type engine but specified engine singer, it can be speculated that he/she wants mining. Then the proper guide would be given to make him/her correctly designate the author. This is same for the PBFT or PoA type.
Now the solo type engine but `no-miner` is impossible. It is weird in some sense.

### DIFF
--- a/codechain/codechain.yml
+++ b/codechain/codechain.yml
@@ -155,14 +155,10 @@ args:
         long: author
         help: Specify the block author (aka "coinbase") address for sending block rewards from sealed blocks.
         takes_value: true
-        conflicts_with:
-            - no-miner
     - engine-signer:
         long: engine-signer
         help: Specify the address which should be used to sign consensus messages and issue blocks.
         takes_value: true
-        conflicts_with:
-            -  no-miner
     - password-path:
         long: password-path
         help: Specify the password file path.
--- a/codechain/config/mod.rs
+++ b/codechain/config/mod.rs
@@ -215,7 +215,6 @@ pub struct Operating {
 #[derive(Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct Mining {
-    pub disable: Option<bool>,
     pub author: Option<PlatformAddress>,
     pub engine_signer: Option<PlatformAddress>,
     pub mem_pool_size: Option<usize>,
@@ -374,9 +373,6 @@ impl Operating {
 
 impl Mining {
     pub fn merge(&mut self, other: &Mining) {
-        if other.disable.is_some() {
-            self.disable = other.disable;
-        }
         if other.author.is_some() {
             self.author = other.author;
         }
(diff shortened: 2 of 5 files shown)


codechain/config/presets/config.dev.toml
```diff
@@ -5,7 +5,7 @@ keys_path = "keys"
 chain = "solo"
 
 [mining]
-disable = false
+disable = true
```
sgkim126: I think it's better to remove from the presets also.

codechain/config/mod.rs
```diff
@@ -395,10 +395,9 @@ impl Mining {
     }
 
     pub fn overwrite_with(&mut self, matches: &clap::ArgMatches) -> Result<(), String> {
-        if matches.is_present("no-miner") {
-            self.disable = Some(true);
+        if matches.is_present("author") || matches.is_present("engine-signer") {
```
sgkim126: You should check self.author and self.engine_signer also.
I think it's better to check at the last of the method.

codechain/config/presets/config.dev.toml
HoOngEe: Do you mean to remove the `disable` preset default and set it after checking cli arguments?

sgkim126: Yes. I do.

HoOngEe: Ok, I understand.

codechain/config/mod.rs
HoOngEe: Ok. I moved the checking to the last part and made to check self.author and self.engine_signer

codechain/config/mod.rs
```diff
@@ -439,6 +435,11 @@ impl Mining {
         if let Some(work_queue_size) = matches.value_of("work-queue-size") {
             self.work_queue_size = Some(work_queue_size.parse().map_err(|_| "Invalid size")?);
         }
+        if self.author.is_some() || self.engine_signer.is_some() {
```
sgkim126: `disable` field of `config::Mining` is weird because we cannot set `disable` via a config file.
Let's remove this field and make add a method to check it.

HoOngEe: Ok, that seems great

codechain/codechain.yml
```diff
@@ -133,21 +133,14 @@ args:
         long: enable-devel-api
         help: Enable the RPC's devel APIs
         takes_value: false
-    - no-miner:
```
sgkim126: For the backward compatibility, please keep this CLI and show the warning message that this option is deprecated.

sgkim126: Let's remove it on 2.0 release

HoOngEe: I kept the `no-miner` CLI. And wanted to add a message using `cwarn!` macro but the CLI options are processed before building `logger`, I just added a warning message using `println!`

sgkim126:
@HoOngEe,
Because we don't have a plan to release a new backward-compatible binary, I think it should be merged after 2.0 and remove no-miner option at 3.0.
Please wait for the next release.

PR #1496, opened 2019-04-26
Remove Reactor

sgkim126:
It was designed to create a thread for the future, but we don't use it.
I think it's better to re-design the Reactor because the future will be
stabilized in the standard library soon or use the tokio directly.


sgkim126:
@kseo Please review it.

sgkim126:
@kseo Please review it.

PR #1499, opened 2019-04-26
Add base-path option to the codechin cli arguments

### LINKED ISSUE #236: Add option to specify base directory
Currently, config files are loaded from relative path from working directory.
We need be able to specify base directory to find them.

HoOngEe:
Now It's possible to specify `base-path` on which the keys and db directory will be created when running codechain. But there already are options `db-path` and `keys-path`.
The rule relating to `db-path`, `keys-path` and `base-path` is like following:

`db-path` and `keys-path` have higher priority than `base-path` so if
both `db-path` and `keys-path` are set, `base-path` becomes meaningless.

it closes #236


codechain/run_node.rs
```diff
@@ -253,10 +255,9 @@ pub fn run_node(matches: &ArgMatches) -> Result<(), String> {
     clogger::init(&LoggerConfig::new(instance_id)).expect("Logger must be successfully initialized");
 
     let pf = load_password_file(&config.operating.password_path)?;
-    let keys_path = match config.operating.keys_path {
-        Some(ref keys_path) => keys_path,
-        None => DEFAULT_KEYS_PATH,
-    };
+    let base_path = config.operating.base_path.as_ref().unwrap().clone();
+    let combined_keys_path = base_path + "/" + DEFAULT_KEYS_PATH;
+    let keys_path = config.operating.keys_path.as_ref().unwrap_or(&combined_keys_path);
```
sgkim126:
```rust
let keys_path = config.operating.keys_path.map(String::as_str).unwrap_or_else(|| base_path + "/" + DEFAULT_KEYS_PATH);
```

codechain/run_node.rs
```diff
@@ -203,7 +203,9 @@ fn unlock_accounts(ap: &AccountProvider, pf: &PasswordFile) -> Result<(), String
 }
 
 pub fn open_db(cfg: &config::Operating, client_config: &ClientConfig) -> Result<Arc<KeyValueDB>, String> {
-    let db_path = cfg.db_path.as_ref().map(String::as_str).unwrap();
+    let base_path = cfg.base_path.as_ref().unwrap().clone();
+    let combined_db_path = base_path + "/" + DEFAULT_DB_PATH;
+    let db_path = String::as_str(cfg.db_path.as_ref().unwrap_or(&combined_db_path));
```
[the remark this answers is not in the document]
HoOngEe: To match type with the result after the `map(String::as_str)` the code should be like `... .unwrap_or_else(|| String::as_str(&(base_path + "/" + DEFAULT_DB_PATH)))`. But in that case we uses the reference of temporary value, so the "temporary value does not live long enough" problem arises. That's the reason why I declared a new variable `combined_db_path`. Originally I was fell into the dilemma : To construct a value in one expression using `base_path + "/" + DEFAULT_DB_PATH` it must not be used as a reference. Simultaneously `cfg` is a borrowed content so we cannot extract values of it. We should use `as_ref()`. Then as long as I know, there is no way to construct one value using both in one sentence combined with `unwrap_or()` or `unwrap_or_default()`.
Is there any good way?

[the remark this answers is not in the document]
sgkim126: Sorry. The above code was wrong.
Use
```rust
let db_path = cfg.db_path.as_ref().map(String::clone).unwrap_or_else(|| base_path + "/" + DEFAULT_DB_PATH);
let client_path = Path::new(&db_path);
```
IMO, we don't need to try to take the reference at the cost of creating a new string.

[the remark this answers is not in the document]
HoOngEe: Changed

codechain/run_node.rs
HoOngEe: Changed

codechain/config/presets/config.prod.toml
```diff
@@ -1,9 +1,9 @@
 [codechain]
 quiet = false
-db_path = "db"
-keys_path = "keys"
+base_path = "."
 chain = "mainnet"
 
+
```
sgkim126: You don't need this change.

sgkim126:
Hi @majecty @foriequal0,

I have a question about the behavior of a relative path.
What do you expect when you specify base-dir and specify db-path as a relative path?
A relative path from the current working directory? or from the base dir?

majecty:
I'm not familiar with the setting two paths in CLI.
I found that in the Parity's config, the DB path is relative to the base path.
https://github.com/paritytech/parity-ethereum/blob/7c335e87642b848f61ab3ee7bbff5964b53d6f11/util/dir/src/lib.rs#L179

codechain/config/presets/config.prod.toml
HoOngEe: Without this change, the code in the line https://github.com/CodeChain-io/codechain/pull/1499/files/3c283f526dfb43ca1261fa5a77f090e2626b41c7#diff-22a0b3dcef5d27fab89655ebdb330967R257 could be a cause of panic.
Do I need to remove this change and use `unwrap_or` for the line mentioned above?

sgkim126: Why does it panic when you remove an empty line?

HoOngEe: Oh I'm sorry I misunderstood that you mentioned about the `+base_path = "."` change.

PR #1512, opened 2019-05-03
Do not prevote to the block with irrelevant generation time

### LINKED ISSUE #1505: Do not pre-vote for blocks generated in the past or future
Because the block proposer can cheat the block generation time, the validators should not sign the very old block or the block generated in the future.

It's on the basic assumption that all nodes try to sync the time and the time error between the server is not that big. However, synchronizing the time of all nodes is impossible so the validator should be tolerant to some time error. The tolerance is how much the time error is allowed.

The degree of acceptance of future generated time error and the past generated time error must be treated differently. If the node keeps trying to synchronize the machine time, it is very rare to receive a proposal block generated in the future, and the error will be little. But time error of the block created at the past could be relatively large due to the network propagation time.

To handle this issue, you should make
* [x] the tendermint module doesn't send a pre-vote if the block was generated over 30 seconds ago or 5 seconds later from the machine time.
* [x] Make the above 30 seconds and 5 seconds configurable.
    * They should be the machine property, not the chain property.

HoOngEe:
If the proposal generation time is 30 seconds or more before from now or
if 5 seconds or more later from now, the block generation time is now considered irrelevant.

It closes #1505


core/src/consensus/tendermint/worker.rs
```diff
@@ -667,15 +668,17 @@ impl Worker {
...
                     let block_hash = match &self.last_two_thirds_majority {
-                        TwoThirdsMajority::Empty => None,
-                        TwoThirdsMajority::Unlock(_) => None,
-                        TwoThirdsMajority::Lock(locked_view, block_hash) => {
-                            if locked_view == &self.view {
-                                Some(*block_hash)
-                            } else {
-                                None
-                            }
+                        TwoThirdsMajority::Lock(locked_view, block_hash)
```
sgkim126: @majecty Would you check it?
I think tendermint should pre-commit the locked block regardless of the block generation time.

majecty: I think it's OK.
1. The locked proposal will receive pre-commits in next round. No problem for the liveness.
2. Other nodes don't know whether a node received enough pre-votes in the prevote timeout or not. It is not the violation of the rule in the other node's view.

majecty: On second thought, do not sending pre-vote message will be simpler.

HoOngEe: @sgkim126 Is it OK to send nil message in the prevote step rather than the precommit step?

sgkim126: I agree with @majecty.
Prevent sending prevote would be simpler.
Please change the commit and issue.

HoOngEe: I changed the design following the above discussion and changed also the pr message, commit message and issue message.

core/src/consensus/tendermint/worker.rs
```diff
@@ -667,15 +673,10 @@ impl Worker {
...
                     let block_hash = match &self.last_two_thirds_majority {
-                        TwoThirdsMajority::Empty => None,
-                        TwoThirdsMajority::Unlock(_) => None,
-                        TwoThirdsMajority::Lock(locked_view, block_hash) => {
-                            if locked_view == &self.view {
-                                Some(*block_hash)
-                            } else {
-                                None
-                            }
+                        TwoThirdsMajority::Lock(locked_view, block_hash) if locked_view == &self.view => {
```
sgkim126: It seems that it's the same with the previous code. Is it?

HoOngEe: Yes, It's the pre-commit step. Do I need to separate it to another commit or just revert it ?

sgkim126: Why do you change if it doesn't change the behavior?

HoOngEe: At first I was changing the pre-commit step and it helps to introduce checking block generation time and rejecting if irrelevant. After I had reverted it, I thought it still helps because it simplifies the routine. But it's meaningless in the sense that it doesn't change the behavior.

sgkim126: And it makes code fragile when changing because you used the underscore arm.
When someone adds a new enumerator, the compiler knows where to fix it in the original code, but it doesn't know in this code.

HoOngEe: Ah, I understand. I couldn't think about that.

HoOngEe:
I made a English review request to @ScarletBlue on the `codechain.yml` cli-option help message.

sgkim126:
@HoOngEe please resolve the conflicts

Issue #1514, opened 2019-05-07
Tendermint replaces the "beset header" with a lower-scoring header.


majecty:
Tendermint replaces the "best header" and the "best block" if all pre-commits are collected. However, it should not replace the "best header" if the original "best header" has a higher block number.

sgkim126:
Let's close it if there is no remaining issue.

PR #1518, opened 2019-05-08
Do not set the "best header" to a header that has a lower number than before

majecty:
This fixes https://github.com/CodeChain-io/codechain/issues/1514


sgkim126:
What should we do if the score of the committed header is greater than the current one even the number is lower?
I think it will not happen in PBFT but can happen in PoW.

majecty:
The function "update_best_as_committed" only be called in Tendermint consensus. The function should not be called If the consensus is PoW.

PR #1526, opened 2019-05-10
Reduce booting time by using is_empty method

### LINKED ISSUE #1513: Long booting time when there are many blocks.
Currently, it takes too much time to boot.
I'm suspecting the client initialization.

HoOngEe:
The previous version of codechain booting time is so slow when the db becomes heavy. The reason is that when codechain checks if db is empty, it used `db.keys().is_empty()`. The routine had collected all the keys in the db to a hashmap, so the majority of the booting time was occupied by the routine for iterating and collecting keys.
I sent the method `is_empty` in `JournalDB` up to `HashDB` and used the method in db empty checking.

It closes #1513


foriequal0:
Why do we have method that returns entire keys for the DB in the first place?

HoOngEe:
The original method `keys()` in `HashDB` is part of parity's implementation and I'm just guessing that we used `db.keys().is_emtpy()`  because there was no `db.is_empty()`.

PR #1530, opened 2019-05-14
Add spec for ChangeParameters

sgkim126:


spec/Staking.md
```diff
@@ -351,6 +351,40 @@ share of "0xAB..CD" = Math.floor(35 * (1000 / (500 + 1000))) = 23
 remaing share (of author) = 110 - (11 + 23) = 76
 ```
 
+## ChangeParameters
+This transaction will change the common parameters when more than half of the stakeholders agree.
+It does not change the other fields of the scheme file, because there are fields related to the genesis block.
+
+It does not provide a feature of voting.
+The vote initiator should collect the signature through the off-chain.
```
kseo: signature -> signatures

spec/Staking.md
```diff
@@ -351,6 +351,40 @@ share of "0xAB..CD" = Math.floor(35 * (1000 / (500 + 1000))) = 23
...
+key := usize
+value := usize | u64 | boolean | string
+```
+It is the list of the fields that the transaction changes.
+The stakeholder MUST NOT sign the transaction when the type of value is not a type that the key expected.
+
+The parameters that are not in the new_parameters are kept as the previous value.
+
+#### signatures
+`signatures` are the signature of stakeholders.
```
kseo: signature -> signatures

spec/Staking.md
```diff
@@ -351,6 +351,40 @@ share of "0xAB..CD" = Math.floor(35 * (1000 / (500 + 1000))) = 23
...
+The stakeholders should send the signature of `[ 0xFF, metadata_seq, new_parameters ]` to the vote initiator if they agree to the change.
+The transaction is valid only if more than half of the stakeholders are agreed.
```
kseo: are agreed -> agree

spec/Staking.md
```diff
@@ -351,6 +351,40 @@ share of "0xAB..CD" = Math.floor(35 * (1000 / (500 + 1000))) = 23
 remaing share (of author) = 110 - (11 + 23) = 76
 ```
 
+## ChangeParameters
+This transaction will change the common parameters when more than half of the stakeholders agree.
+It does not change the other fields of the scheme file, because there are fields related to the genesis block.
```
ScarletBlue: It does not change other fields of the scheme file because there are fields related to the genesis block.

spec/Staking.md
```diff
@@ -351,6 +351,40 @@ share of "0xAB..CD" = Math.floor(35 * (1000 / (500 + 1000))) = 23
...
+
+It does not provide a feature of voting.
```
ScarletBlue: It also does not provide a voting feature.

spec/Staking.md
```diff
@@ -351,6 +351,40 @@ share of "0xAB..CD" = Math.floor(35 * (1000 / (500 + 1000))) = 23
...
+The vote initiator should collect the signatures through the off-chain.
+
+This transaction increases the `seq` of `Metadata` and changes the `params` of `Metadata`.
+
+### Action
+`[ 0xFF, metadata_seq, new_parameters, ...signatures ]`
+
+#### metadata_seq
+The transaction fails if it's different from the `seq` of `Metadata`.
+It's introduced to prevent replay attacks.
```
ScarletBlue: The transaction fails if the metadata_seq is different from the `seq` of `Metadata` and is introduced to prevent replay attacks.

spec/Staking.md
```diff
@@ -351,6 +351,39 @@ share of "0xAB..CD" = Math.floor(35 * (1000 / (500 + 1000))) = 23
...
+### Action
+`[ 0xFF, metadata_seq, new_parameters, ...signatures ]`
+
+#### metadata_seq
+The transaction fails if the metadata_seq is different from the `seq` of `Metadata` and is introduced to prevent replay attacks.
+
+#### new_parameters
+```
+new_parameters := [ new_parameter(, new_parameter)* ]
+new_parameter := [ key, value ]
```
foriequal0: We use different conventions for the key of common parameters between rust(snake_case) and JSON (camelCase). What's the convention for the transaction?

spec/Staking.md
```diff
@@ -351,6 +351,39 @@ share of "0xAB..CD" = Math.floor(35 * (1000 / (500 + 1000))) = 23
...
+
+key := usize
+value := usize | u64 | boolean | string
+```
+It is the list of the fields that the transaction changes.
+The stakeholder MUST NOT sign the transaction when the type of value is not a type that the key expected.
```
ScarletBlue: Is it the key that expects a type?

sgkim126: yes

Issue #1506, opened 2019-04-30
Implement a transaction to change the parameters


sgkim126:
## Purpose of this transaction
Currently, CodeChain has a static scheme fixed from the genesis. In other words, CodeChain cannot change the operating rule, which is problematic when it comes to adding a new rule or changing the existing rule, such as the minimum fee.
So I want to add a new transaction to change the scheme.

## What this transaction does
This transaction will change only [the common parameters](https://github.com/CodeChain-io/codechain/blob/18e1e80af4503cfd4d06e9514e9f88370995abda/core/res/corgi.json#L54-L78) if enough stakeholders agree. The current stakes necessary to change the common parameters is more than 67% of the total stakes.

### Changing the fields other than the common parameters
It does not change the other fields of the scheme file, because there are fields related to the genesis block.
### Voting
It does not provide a feature of voting. Voting makes the rule complicated. Instead of voting on the chain, the voting initiator should collect the signature through the off-chain.

```
[ [CHANGE_PARAMETERS_ID, metadata_seq, new_parameters], signatures ]

CHANGE_PARAMETERS_ID := u16
metadata_seq := 0
new_parameters := string
signatures := array of string
```
### CHANGE_PARAMETERS_ID
```
0xFF
```
It's a constant number to distinguish from other transaction type.

### metadata_seq
This field is introduced to prevent a replay attack. The [metadata](https://github.com/CodeChain-io/codechain/blob/18e1e80af4503cfd4d06e9514e9f88370995abda/state/src/item/metadata.rs#L25) has a field to store `seq`. It's increased when the parameter is changed.

### new_parameters
```
new_parameters := [ new_parameter(, new_parameter)* ]
new_parameter := [ key ]
    | [ key, value ]

key := string
value := usize | u64 | boolean | string
```
The list that has the fields that this transaction will change.
The stakeholder MUST NOT sign the transaction when it the type of value is not a type that the key expected.
The parameters that are not in the `new_parameters` are kept as the previous value.

### signatures
The ChangeParameters transaction is valid only when the sum of the stakes that signed in the transaction is greater than the threshold. It's the signature of `[CHANGE_PARAMETERS_ID, metadata_seq, new_parameters]`.

### Metadata
[Metadata](https://github.com/CodeChain-io/codechain/blob/bdff3261b5f13edb81cc70cc37cfc15602f2c493/state/src/item/metadata.rs#L26-L29) state is affected.
One field is added to Metadata.

#### seq
[...]

PR #1535, opened 2019-05-15
Implement ChangeParams

sgkim126:
It closes https://github.com/CodeChain-io/codechain/issues/1506

### DIFF
--- a/core/src/consensus/stake/actions.rs
+++ b/core/src/consensus/stake/actions.rs
@@ -34,7 +34,7 @@ pub enum Action {
     },
     ChangeParams {
         metadata_seq: u64,
-        params: CommonParams,
+        params: Box<CommonParams>,
         signatures: Vec<Signature>,
     },
 }
@@ -62,7 +62,7 @@ impl Encodable for Action {
                 s.begin_list(3 + signatures.len())
                     .append(&ACTION_TAG_CHANGE_PARAMS)
                     .append(metadata_seq)
-                    .append(params);
+                    .append(&**params);
                 for signature in signatures {
                     s.append(signature);
                 }
@@ -110,7 +110,7 @@ impl Decodable for Action {
                     })
                 }
                 let metadata_seq = rlp.val_at(1)?;
-                let params = rlp.val_at(2)?;
+                let params = Box::new(rlp.val_at(2)?);
                 let signatures = (3..item_count).map(|i| rlp.val_at(i)).collect::<Result<_, _>>()?;
                 Ok(Action::ChangeParams {
                     metadata_seq,
@@ -133,7 +133,7 @@ mod tests {
     fn decode_fail_if_change_params_have_no_signatures() {
         let action = Action::ChangeParams {
             metadata_seq: 3,
-            params: CommonParams::default_for_test(),
+            params: CommonParams::default_for_test().into(),
             signatures: vec![],
         };
         assert_eq!(
@@ -149,7 +149,7 @@ mod tests {
     fn rlp_of_change_params() {
         rlp_encode_and_decode_test!(Action::ChangeParams {
             metadata_seq: 3,
-            params: CommonParams::default_for_test(),
+            params: CommonParams::default_for_test().into(),
             signatures: vec![Signature::random(), Signature::random()],
         });
     }
--- a/core/src/consensus/stake/actions.rs
+++ b/core/src/consensus/stake/actions.rs
@@ -14,13 +14,15 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use ckey::Address;
+use ckey::{Address, Signature};
+use ctypes::CommonParams;
 use rlp::{Decodable, DecoderError, Encodable, RlpStream, UntrustedRlp};
 
 const ACTION_TAG_TRANSFER_CCS: u8 = 1;
 const ACTION_TAG_DELEGATE_CCS: u8 = 2;
+const ACTION_TAG_CHANGE_PARAMS: u8 = 0xFF;
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum Action {
     TransferCCS {
         address: Address,
@@ -30,6 +32,11 @@ pub enum Action {
         address: Address,
         quantity: u64,
     },
+    ChangeParams {
+        metadata_seq: u64,
+        params: CommonParams,
+        signatures: Vec<Signature>,
+    },
 }
 
 impl Encodable for Action {
@@ -38,11 +45,28 @@ impl Encodable for Action {
             Action::TransferCCS {
                 address,
                 quantity,
-            } => s.begin_list(3).append(&ACTION_TAG_TRANSFER_CCS).append(address).append(quantity),
+            } => {
+                s.begin_list(3).append(&ACTION_TAG_TRANSFER_CCS).append(address).append(quantity);
+            }
             Action::DelegateCCS {
                 address,
                 quantity,
-            } => s.begin_list(3).append(&ACTION_TAG_DELEGATE_CCS).append(address).append(quantity),
+            } => {
+                s.begin_list(3).append(&ACTION_TAG_DELEGATE_CCS).append(address).append(quantity);
+            }
+            Action::ChangeParams {
+                metadata_seq,
+                params,
+                signatures,
+            } => {
+                s.begin_list(3 + signatures.len())
+                    .append(&ACTION_TAG_CHANGE_PARAMS)
+                    .append(metadata_seq)
+                    .append(params);
+                for signature in signatures {
+                    s.append(signature);
+                }
+            }
         };
     }
 }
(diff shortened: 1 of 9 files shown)


sgkim126:
@majecty @foriequal0 Would you review the current code first?
I'll remove the `do-not-merge` tag after adding tests

core/src/scheme/scheme.rs
```diff
@@ -163,6 +163,7 @@ impl Scheme {
         }
 
         {
+            println!("metadata init");
```
majecty: Should we use `println!`?

core/src/codechain_machine.rs
```diff
@@ -46,22 +47,51 @@ impl CodeChainMachine {
         CodeChainMachine {
             params: vec![Params {
```
majecty: Should we read the params from the Metadata DB?

core/src/consensus/blake_pow/mod.rs
```diff
@@ -164,7 +165,16 @@ impl ConsensusEngine<CodeChainMachine> for BlakePoW {
...
             + self.block_fee(Box::new(block.transactions().to_owned().into_iter().map(Into::into)));
-        self.machine.add_balance(block, &author, total_reward)
+        self.machine.add_balance(block, &author, total_reward)?;
+
+        let current_metadata = block.state().metadata()?.unwrap();
+        let current_seq = current_metadata.seq();
+        let block_number = block.header().number();
+        let parent_common_seq = self.machine.metadata_seq(Some(block_number - 1));
+        if parent_common_seq != current_seq {
+            self.machine.insert_common_params(block_number, current_seq, current_metadata.params().clone().unwrap());
```
majecty: What happened when a reorganization has occurred?
Should we remove the common params when the best chain is changed?

core/src/consensus/tendermint/engine.rs
```diff
@@ -173,6 +173,13 @@ impl ConsensusEngine<CodeChainMachine> for Tendermint {
         if total_fee != min_fee {
             self.machine.add_balance(block, &author, total_fee - min_fee)?
         }
+
+        let current_metadata = block.state().metadata()?.unwrap();
+        let current_seq = current_metadata.seq();
+        let parent_common_seq = self.machine.metadata_seq(Some(block_number - 1));
+        if parent_common_seq != current_seq {
+            self.machine.insert_common_params(block_number, current_seq, current_metadata.params().clone().unwrap());
+        }
```
majecty: What happen the block is not confirmed?
A proposal block always calls the `on_close_block`, but the proposal block may be not confirmed.

sgkim126: Yes, you are right.
Do we have an event that is occurred when the block is confirmed?

majecty: Yes, If we have the event, it will be perfect.

core/src/consensus/blake_pow/mod.rs
sgkim126: I modified `insert_common_params` to remove the changing of reorganized blocks.

core/src/consensus/tendermint/engine.rs
majecty: I found that `ChainNotify`'s `new_blocks` is called when the best block is changed.
The new best block's hash is contained in the `enacted` parameter.

core/src/codechain_machine.rs
```diff
@@ -22,22 +22,23 @@ use cstate::{StateError, TopState, TopStateView};
...
 
 use crate::block::{ExecutedBlock, IsBlock};
 use crate::client::{BlockInfo, TransactionInfo};
 use crate::error::Error;
 use crate::header::Header;
-use crate::scheme::CommonParams;
 use crate::transaction::{SignedTransaction, UnverifiedTransaction};
 
 struct Params {
+    seq: usize,
```
foriequal0: It seems that it is tied to the `seq` of the `metadata`. Should we change the name to `metadata_seq`?

core/src/consensus/stake/mod.rs
```diff
@@ -168,6 +204,39 @@ pub fn get_stakes(state: &TopLevelState) -> StateResult<HashMap<Address, u64>> {
     Ok(result)
 }
 
+fn change_params(
+    state: &mut TopLevelState,
+    metadata_seq: usize,
+    params: CommonParams,
+    signatures: &[Signature],
+) -> StateResult<()> {
+    state.update_params(metadata_seq, params.clone())?;
```
foriequal0: Why do we update before the signature check?

core/src/consensus/stake/mod.rs
```diff
@@ -168,6 +204,39 @@ pub fn get_stakes(state: &TopLevelState) -> StateResult<HashMap<Address, u64>> {
...
+    let stakes = get_stakes(state)?;
+    let signed_stakes = signatures.iter().try_fold(0, |sum, signature| {
+        let public = recover(signature, &encoded_action).unwrap_or_else(|err| {
+            unreachable!("The transaction with an invalid signature cannot pass the verification: {}", err);
+        });
+        let address = public_to_address(&public);
+        stakes.get(&address).map(|stake| sum + stake).ok_or_else(|| RuntimeError::SignatureOfInvalidAccount(address))
+    })?;
+    let total_stakes: u64 = stakes.values().sum();
+    if total_stakes / 2 > signed_stakes {
```
foriequal0: https://github.com/CodeChain-io/codechain/pull/1530/files#diff-ec496aca1d063d60bc2a7e03e1bcea3dR385
Half of the stakeholders? Half of the stakes?

sgkim126: They mean the same thing.
Mostly, if there are no mentions about it, one stake means one vote.
https://en.wikipedia.org/wiki/One_share,_one_vote

core/src/consensus/stake/mod.rs
sgkim126: Because checking signatures are the most expensive work.

foriequal0: `update_params` doesn't short-circuit so it doesn't change the average cost.

sgkim126: It does when the metadata_seq is different.
The code that returns `Ok(())` should be replaced to an error. Anyway, it does.

foriequal0: Okay then, can we mention the purpose of the unusual code the placement?

sgkim126: done

core/src/scheme/scheme.rs
sgkim126: removed

sgkim126:
I'm working on this issue.
The importer uses CommonParams to verify the transaction size limit or the transaction fee before importing because the CommonParams were immutable.
But this should be postponed after the transactions in the parent block executed because one of them can be ChangeParams.

sgkim126:
It depends on https://github.com/CodeChain-io/codechain/pull/1546. I'll remove the WIP tag after merging #1546.

core/src/consensus/tendermint/engine.rs
sgkim126: I don't need this event anymore because the states store it.

core/src/codechain_machine.rs
sgkim126: I removed this because I made the states have the params.

sgkim126:
@majecty @foriequal0 I added tests.
This PR depends on https://github.com/CodeChain-io/codechain/pull/1552 I'll remove `do-not-merge` after merging the PR.

sgkim126:
@majecty @foriequal0 Please review it again. All dependencies has been merged.

core/src/consensus/stake/actions.rs
```diff
@@ -30,6 +32,11 @@ pub enum Action {
         address: Address,
         quantity: u64,
     },
+    ChangeParams {
+        metadata_seq: u64,
+        params: CommonParams,
```
foriequal0: I think it contradicts the [spec](https://github.com/sgkim126/codechain/blob/master/spec/Staking.md#new_parameters). Spec says that the transaction contains fields that are going to be changed, however, this implementation seems to have whole `CommonParams` like a snapshot.

core/src/consensus/stake/actions.rs
```diff
@@ -77,7 +101,56 @@ impl Decodable for Action {
...
+                let item_count = rlp.item_count()?;
+                if item_count < 4 {
+                    return Err(DecoderError::RlpIncorrectListLen {
+                        expected: 4,
+                        got: item_count,
+                    })
+                }
+                let metadata_seq = rlp.val_at(1)?;
+                let params = rlp.val_at(2)?;
+                let signatures = (3..item_count).map(|i| rlp.val_at(i)).collect::<Result<_, _>>()?;
```
foriequal0: `.map(rlp.val_at)` without the closure doesn't work?

core/src/consensus/stake/actions.rs
sgkim126: Oh, I'd forgot to update the spec.
It can be dangerous to specify the field that the transaction wants to change. See the above scenario.
Assuming that the parameter X and Y contradicts in certain value and there are 2N+1 stakeholders and the parameters. If one malicious stakeholder collects the signature to change X of N stakeholders and the signature to change Y of others, only one malicious stakeholder can attack the chain.
So I changed ChangeParams have whole parameters but I forgot to update the spec. I'll update it.

sgkim126: https://github.com/CodeChain-io/codechain/pull/1560 I uploaded PR.

core/src/consensus/stake/actions.rs
sgkim126: It doesn't work.

sgkim126: Actually, I cannot understand what you said. How could you remove the closure here?
`map` takes a function that has 1 argument, and `val_at` is a method that takes 1 argument. It means `val_at` needs 2 parameters.

foriequal0: Okay, I understood. Thank you.

PR #1538, opened 2019-05-16
Fix the broken tests and refine tests code

sgkim126:


sgkim126:
The second commit is refactoring.
The first commit is more important than the second.
Because of my mistake, some e2e tests has been skipped.

test/src/e2e.long/reward2.test.ts
```diff
@@ -97,7 +97,13 @@ describe("reward2", function() {
 
         // nodeA creates 2 blocks
         {
-            await nodeA.pay(aliceAddress, 100); // +160 for alice in nodeA
+            await nodeA.pay(aliceAddress, 100); // +100 +50 +10*2/10 for alice in nodeA, +10*1/10 for bob
```
foriequal0: I think the test would have been easier if we used a scheme without `engine.solo.params.genesisStakes`. Can we split the scheme?

test/src/e2e.long/timelock.test.ts
```diff
@@ -109,14 +109,8 @@ describe("Timelock", function() {
         });
         const blockNumber = await node.sdk.rpc.chain.getBestBlockNumber();
         const seq = await node.sdk.rpc.chain.getSeq(faucetAddress);
```
foriequal0: `blockNumber`, `seq` is not used anymore.

test/src/e2e.long/reward2.test.ts
sgkim126: Splitting the scheme is okay, but even though we should test it too.
The e2e test should be a real scenario, not an easy-to-test scenario.

PR #1540, opened 2019-05-16
Refine interfaces

sgkim126:


core/src/consensus/mod.rs
```diff
@@ -380,9 +380,9 @@ pub trait CodeChainEngine: ConsensusEngine<CodeChainMachine> {
...
         header: &Header,
     ) -> Result<SignedTransaction, Error> {
-        self.machine().verify_transaction_unordered(tx, header)
+        CodeChainMachine::verify_transaction_unordered(tx, header)
     }
 }
 
 // convenience wrappers for existing functions.
-impl<T> CodeChainEngine for T where T: ConsensusEngine<CodeChainMachine> {}
+impl<T> CodeChainEngine for T where T: ConsensusEngine {}
```
foriequal0: Do we still need `CodeChainEngine`? It seems that all `ConsensusEngine` is `CodeChainEngine` due to blanket implementation.

sgkim126: They are different.
`ConsensusEngine` is a trait that each consensus should implement and `CodeChainEngine` is a trait that has default implementations of `verify_transaction_basic` and `verify_transaction_unordered`.

foriequal0: I still doubt that distinction is necessary. Default `impl`s can be moved to `CodeChainEngine`, or they can be implemented as a function without trait like `fn verify_transaction_{basic,unordered}(engine: &ConsensusEngine, ...)` to reduce abstractions. `verify_transaction_unordered` already just forwards to `CodeChainMachine::verify_transaction_unordered(tx, header)`, and `verify_transaction_basic` does almost the same but `find_action_handler_for` is intrrupt it. 
Anyway, I'll approve this PR.

PR #1545, opened 2019-05-17
Use the child of the best header to verify transactions

sgkim126:
Currently, it does not matter because the header is used to get the `CommonParams` that doesn't depend on the block number. But the next patch will make it changeable.


sgkim126:
@majecty Please review it again. I fixed e2e tests.

PR #1546, opened 2019-05-17
Split the verification that needs CommonParams and doesn't need it

sgkim126:
It's to ready the ChangeParams.
The current code assumes that the CommonParams are immutable.

### DIFF
--- a/rpc/src/v1/impls/chain.rs
+++ b/rpc/src/v1/impls/chain.rs
@@ -94,13 +94,17 @@ where
         shard_id: ShardId,
         block_number: Option<u64>,
     ) -> Result<Option<AssetScheme>> {
-        let network_id = self.client.common_params(block_number).network_id();
-        let block_id = block_number.map(BlockId::Number).unwrap_or(BlockId::Latest);
-        Ok(self
-            .client
-            .get_asset_scheme(asset_type, shard_id, block_id)
-            .map_err(errors::transaction_state)?
-            .map(|asset_scheme| AssetScheme::from_core(asset_scheme, network_id)))
+        if let Some(common_params) = self.client.common_params(block_number) {
+            let network_id = common_params.network_id();
+            let block_id = block_number.map(BlockId::Number).unwrap_or(BlockId::Latest);
+            Ok(self
+                .client
+                .get_asset_scheme(asset_type, shard_id, block_id)
+                .map_err(errors::transaction_state)?
+                .map(|asset_scheme| AssetScheme::from_core(asset_scheme, network_id)))
+        } else {
+            Ok(None)
+        }
     }
 
     fn get_text(&self, transaction_hash: H256, block_number: Option<u64>) -> Result<Option<Text>> {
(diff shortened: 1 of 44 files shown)


sgkim126:
It depends on https://github.com/CodeChain-io/codechain/pull/1545

rpc/src/v1/impls/chain.rs
```diff
@@ -113,7 +114,7 @@ where
             .client
             .get_text(transaction_hash, block_id)
             .map_err(errors::transaction_state)?
-            .map(|text| Text::from_core(text, self.client.common_params(block_number).unwrap().network_id())))
+            .map(|text| Text::from_core(text, self.client.common_params(block_id).unwrap().network_id())))
```
majecty: Should we use parent's block common params?

rpc/src/v1/impls/chain.rs
```diff
@@ -237,7 +238,7 @@ where
     }
 
     fn get_min_transaction_fee(&self, action_type: String, block_number: u64) -> Result<Option<u64>> {
-        if let Some(common_parameters) = self.client.common_params(Some(block_number)) {
+        if let Some(common_parameters) = self.client.common_params(block_number.into()) {
```
majecty: Should we use parent's block common params?
If the RPC's caller wants to fee to creates transactions in the next block, the current implementation is OK.
However, if the RPC's caller wants to check the block's min_fee to verify the block, we should use the previous block's common params.

rpc/src/v1/impls/chain.rs
```diff
@@ -219,16 +220,16 @@ where
     fn get_block_by_number(&self, block_number: u64) -> Result<Option<Block>> {
         let id = BlockId::Number(block_number);
         Ok(self.client.block(&id).map(|block| {
-            Block::from_core(block.decode(), self.client.common_params(Some(block_number)).unwrap().network_id())
+            Block::from_core(block.decode(), self.client.common_params(block_number.into()).unwrap().network_id())
```
HoOngEe: We can use predefined variable `id` instead of `block_number.into()`

rpc/src/v1/impls/chain.rs
```diff
@@ -219,16 +220,16 @@ where
...
     }
 
     fn get_block_by_hash(&self, block_hash: H256) -> Result<Option<Block>> {
         let id = BlockId::Hash(block_hash);
         Ok(self.client.block(&id).map(|block| {
             let block = block.decode();
-            let block_number = block.header.number();
-            Block::from_core(block, self.client.common_params(Some(block_number)).unwrap().network_id())
+            let block_hash = block.header.hash();
+            Block::from_core(block, self.client.common_params(block_hash.into()).unwrap().network_id())
```
HoOngEe: We can use predefined variable `id` instead of `block_hash.into()`

rpc/src/v1/impls/chain.rs
sgkim126: I make it use the parameters of the parent block.

Issue #1547, opened 2019-05-20
getTransactionByTracker RPC doesn't return WrapCCC transactions 


Kais-DkM:
Indexer reports that an [AssetTransfer transaction](https://corgi.codechain.io/explorer/tx/0x9772d8a7785a9395b9c28aaf53d11c0e86bb1b078ef81c07ee970e830e8aaf43) spent a [WrapCCC transaction](https://corgi.codechain.io/explorer/tx/0x71765bbdedf9e22bcb053c41ce5a66852508ca4d0b066d0e943800c7e4719c32), but CodeChain returns null when `getTransactionByTracker` RPC is requested with the WrapCCC transaction's tracker.

You can try the request with the following command:
`curl https://corgi-rpc.codechain.io -H 'Content-Type: application/json'     -d '{"jsonrpc": "2.0", "method": "chain_getTransactionByTracker", "params": ["0x033e49e6e7eeb612694eedfd2b6274c4f0d5b21ce866710c02b4fb8c62f9572b"], "id": null}'`

HoOngEe:
Current implementation stores (tracker, transactionAddress) into `body_db` only for `AssetTransaction`. Because `AssetTransaction` does not include `WrapCCC`, the `WrapCCC` transaction cannot be tracked back from the corresponding `tracker`.
After the patch, existing blocks should be re-synced to apply the correct tracking function.

sgkim126:
Fixed by https://github.com/CodeChain-io/codechain/pull/1548

PR #1548, opened 2019-05-20
Write WrapCCC tracker to db

HoOngEe:
It fixes https://github.com/CodeChain-io/codechain/issues/1547

### DIFF
--- a/core/src/blockchain/body_db.rs
+++ b/core/src/blockchain/body_db.rs
@@ -18,7 +18,6 @@ use std::collections::{HashMap, HashSet};
 use std::mem;
 use std::sync::Arc;
 
-use ctypes::transaction::ShardTransaction;
 use kvdb::{DBTransaction, KeyValueDB};
 use parking_lot::{Mutex, RwLock};
 use primitives::{Bytes, H256};
@@ -347,9 +346,9 @@ fn transaction_address_entries(
     parcel_hashes: impl IntoIterator<Item = UnverifiedTransaction>,
 ) -> impl Iterator<Item = TransactionHashAndAddress> {
     parcel_hashes.into_iter().enumerate().filter_map(move |(parcel_index, parcel)| {
-        Option::<ShardTransaction>::from(parcel.action.clone()).map(|tx| {
+        parcel.tracker().map(|tracker| {
             (
-                tx.tracker(),
+                tracker,
                 TransactionAddresses::new(TransactionAddress {
                     block_hash,
                     index: parcel_index,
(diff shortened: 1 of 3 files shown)


sgkim126:
@HoOngEe Would you check the use case of [Transaction::tracker](https://github.com/CodeChain-io/codechain/blob/master/types/src/transaction/transaction.rs)?
Fixing this function might be better, but please make sure where the function is used.

HoOngEe:
@sgkim126 There are two use cases of [Transaction::tracker](https://github.com/CodeChain-io/codechain/blob/fd0294fba015f1b26dd1c7dc53b8df5685830383/types/src/transaction/transaction.rs#L54). One is for pushing invoice information to block(our target usage) and the other is for calculating `asset_type` in [`devels.rs`](https://github.com/CodeChain-io/codechain/blob/fd0294fba015f1b26dd1c7dc53b8df5685830383/rpc/src/v1/impls/devel.rs#L233).

HoOngEe:
But in the second case, the `asset_type` operates only for `MintAsset` transactions. I think I can safely change the [Transaction::tracker](https://github.com/CodeChain-io/codechain/blob/fd0294fba015f1b26dd1c7dc53b8df5685830383/types/src/transaction/transaction.rs#L54).

HoOngEe:
After the patch, I synchronized local node to Corgi network and got the rpc response previously didn't work.
```
curl http://localhost:8080 -H 'Content-Type: application/json' -d '{"jsonrpc": "2.0", "method": "chain_getTransactionByTracker", "params": ["0x033e49e6e7eeb612694eedfd2b6274c4f0d5b21ce866710c02b4fb8c62f9572b"], "id": null}'
{"jsonrpc":"2.0","result":{"action":{"lockScriptHash":"0x5f5960a7bca6ceeeb0c97bc717562914e7a1de04","parameters":["d5b0e27e1e87d9446b0a45cbe38821ccc1430380"],"payer":"wccqxxz05354l57u877989aus855y0v7wual54n86yf","quantity":"0x3e8","shardId":0,"type":"wrapCCC"},"blockHash":"0x9ae96d65e866ce760617f24a94c2febbc8f8c2c09e9ddb60081a1514f36be1e9","blockNumber":290457,"fee":"0x186a0","hash":"0x71765bbdedf9e22bcb053c41ce5a66852508ca4d0b066d0e943800c7e4719c32","networkId":"wc","result":true,"seq":8,"sig":"0x145e5005e2a4a03ba218a43be01289b8541c81f6eefa781ad024834eebcc3a0719a2a9835e4d94d233dda3e5705e53d74ee7f1518c60019272e37e0162f7a54d01","transactionIndex":1},"id":null}
```

sgkim126:
@HoOngEe Thank you for checking

PR #1549, opened 2019-05-21
 Fix chain_getMinTransactionFee to take null as the block number

sgkim126:


sgkim126:
It depens on #1546

sgkim126:
@majecty The dependencies are merged. Please review it.

PR #1552, opened 2019-05-21
Update the term id and the block id when the term changes

sgkim126:


core/src/consensus/stake/action_data.rs
```diff
@@ -184,6 +188,32 @@ impl<'a> Delegation<'a> {
     }
 }
 
+#[derive(Default, RlpEncodable, RlpDecodable, Debug, PartialEq)]
+pub struct Metadata {
+    pub block_number: u64,
```
foriequal0: I think this `Metadata` should move to the `TopLevelState`. `Term` can be used outside of the stake custom handler.

sgkim126: Fixed.

sgkim126:
@foriequal0 Please review it again. I changed https://github.com/CodeChain-io/codechain/pull/1552/commits/1b589f4d2b293052b1197bbabb6bdd67a409559c to use `Range`.

state/src/item/metadata.rs
```diff
@@ -93,28 +108,56 @@ const PREFIX: u8 = super::METADATA_PREFIX;
 
 impl Encodable for Metadata {
     fn rlp_append(&self, s: &mut RlpStream) {
+        let mut len = 4;
```
foriequal0: Is this means `param_changed` implies they should fill the gap for `term` with a default value?
I think then this would can express the intention more.
```rust
const INITIAL_LEN: usize = 4;
const TERM_LEN: usize = 2;
const PARAMS_LEN: usize = 2;
let mut len = INITIAL_LEN;
if term_changed || params_chagned { 
    // Fill default values for term if params_changed even if !term_changed
    len += TERM_LEN;
}
if params_chagned {
    len += PARAMS_LEN;
}
```
and

```rust
s.begin_list(len)
    .append(&PREFIX)
    .append(&self.number_of_shards)
    .append(&self.number_of_initial_shards)
    .append_list(&self.hashes);
if term_changed {
    // append term
} else if params_changed {
    // append default value for term
}
if params_changed {
    // append params
}
```

sgkim126: @foriequal0 fixed

foriequal0: I was pointing out the control flow, not only labeling the constant. However, labeling makes it much better than before.

PR #1555, opened 2019-05-23
Refactor fee distrubte logic and fix the potential bug in tendermint module

sgkim126:


core/src/consensus/solo/mod.rs
```diff
@@ -81,23 +81,23 @@ impl ConsensusEngine for Solo<CodeChainMachine> {
...
         let stakes = stake::get_stakes(block.state()).expect("Cannot get Stake status");
-        for (address, share) in stake::fee_distribute(&author, min_fee, &stakes) {
+        for (address, share) in stake::fee_distribute(total_min_fee, &stakes) {
             self.machine.add_balance(block, &address, share)?
         }
-        if total_reward != min_fee {
-            self.machine.add_balance(block, &author, total_reward - min_fee)?
-        }
+        let stakeholders_share = stake::stakeholders_share(total_min_fee, &stakes);
+        self.machine.add_balance(block, &author, total_reward - stakeholders_share)?;
```
foriequal0: Why iterate stakes twice?
Is this more clear to express the intention that "remaining fee due to the rounding goes to the author"
```rust
let mut distributor = stake::fee_distribute(total_min_fee, &stakes);
for (address, share) in distributor {
  ...
}
self.mahcine.add_balance(block, &author, total_reward - total_min_fee + distributor.get_remaining_fee())?;
```

sgkim126: Because the block author doesn't take the fee at block close after term change introduced.

foriequal0: Will it be greatly different from this?
```rust
stake.add_pending_reward(&author, total_reward - total_min_fee + distributor.get_remaining_fee());
```

sgkim126: Two things are the same because `stakeholders_share` uses the iteration internally.
But when we move the fee distribution to term change, there will be no iterators, so reusing the iterator doesn't have merit at all.

foriequal0: I meant that it 'iterates' HashMap twice, not meant that it doesn't 'reuse iterator'. Anyway, it is a minor thing. I'll approve this PR.

sgkim126: I meant you cannot make it iterate once.
From the next patch, you cannot call `add_pending_reward` here, you should recalculate the reward of every block that generates in a term.

sgkim126: The pending_reward is a field that the former validators will receive.
If you want to store the block reward in a term to avoid calculating it twice, you should introduce a new field.

sgkim126: I think it has merit to store the intermediate reward in states.
```
intermediate_rewards = [ [ address, quantity ] * ]
```
@foriequal0 How about you?

PR #1559, opened 2019-05-24
Pay the block propose rewards at the end of the term

sgkim126:


sgkim126:
It depens on https://github.com/CodeChain-io/codechain/pull/1535

sgkim126:
The last commit based on https://github.com/CodeChain-io/codechain/pull/1562.
I'll change the commit if the spec changes.

majecty:
`on_new_block` is only called in the proposer node, not in other validator nodes.
IMO other validators also need to call `on_new_block` when they received a proposal block or received a block from the sync extension to create the same state with the proposer.

sgkim126:
@majecty @foriequal0 
I removed `on_new_block` and made the tendermint engine pay the reward after one term.

core/src/consensus/tendermint/engine.rs
```diff
@@ -263,3 +290,267 @@ impl ConsensusEngine for Tendermint {
...
+    mut reduced_rewards: u64,
+    missed_signatures: HashMap<Address, (usize, usize)>,
+    mut f: F,
+) -> Result<(), Error> {
+    let lcm_of_proposed = missed_signatures
+        .values()
+        .map(|(proposed, _)| *proposed)
+        .collect::<HashSet<_>>()
+        .into_iter()
+        .fold(1usize, |base, proposed| base.checked_mul(proposed).unwrap());
```
foriequal0: It would explode in the worst case.
When each of the verifiers (30) proposed a unique number of blocks that fill a term (720), then the `missed_signatures` could be [9, 10, 11, ..., 38] (sum = 705). However, the product of them is almost 1.19 * 2^133.

core/src/consensus/tendermint/engine.rs
```diff
@@ -165,11 +172,40 @@ impl ConsensusEngine for Tendermint {
...
+        let client = self
+            .client
+            .read()
+            .as_ref()
+            .ok_or(EngineError::CannotOpenBlock)?
+            .upgrade()
+            .ok_or(EngineError::CannotOpenBlock)?;
+
+        let (start_of_the_next_term, start_of_the_current_term) = {
+            let end_of_the_previous_term = block.state().metadata()?.unwrap().last_term_finished_block_num();
```
majecty: I'm confusing what is the {current, next, previous} terms.
`term id` is calculated by the header.timestamp. The `block` in this function has different `term` which is calculated from the header, and the block's term is `term id - 1`. Am I right?

What is the current term? Is it from the block's timestamp or from the block?

core/src/consensus/tendermint/engine.rs
```diff
@@ -263,3 +290,267 @@ impl ConsensusEngine for Tendermint {
...
+) -> Result<(), Error> {
+    let lcm_of_proposed = missed_signatures
+        .values()
+        .map(|(proposed, _)| *proposed)
+        .collect::<HashSet<_>>()
+        .into_iter()
+        .fold(1usize, |base, proposed| base.checked_mul(proposed).unwrap());
+    let sorted_validators = missed_signatures
+        .into_iter()
+        .map(|(address, (proposed, missed))| (address, missed.checked_mul(lcm_of_proposed / proposed).unwrap()))
```
foriequal0: I understand that `missed / proposed` cannot be calculated precisely without the least common multiplier. However, since we are comparing the non-negative ratio, we can compare `(missed/proposed, missed%proposed)` to eliminate `lcm_of_proposed`

core/src/consensus/tendermint/engine.rs
sgkim126: Okay. I'll replace it with `U256`

core/src/consensus/tendermint/engine.rs
sgkim126: LGTM. I'll change it

core/src/consensus/tendermint/engine.rs
sgkim126: Yes. It's the most confusing part because the current block is the end of the term but the term Id in the metadata is not updated.
The current term in this context is the term that this block closes. The reason which uses the current block to calculate `end_of_the_previous_term` is that because the last block of the previous term is stored in states.

core/src/consensus/tendermint/engine.rs
sgkim126: I changed it to use `num-rational`.

core/src/consensus/tendermint/engine.rs
```diff
@@ -165,11 +173,40 @@ impl ConsensusEngine for Tendermint {
...
+        let rewards = stake::drain_previous_rewards(&mut block.state_mut())?;
+        let client = self
+            .client
+            .read()
+            .as_ref()
+            .ok_or(EngineError::CannotOpenBlock)?
+            .upgrade()
+            .ok_or(EngineError::CannotOpenBlock)?;
+
+        let (start_of_the_next_term, start_of_the_current_term) = {
```
majecty: IMO, 
`start_of_the_current_term == end_of_the_previous_term + 1`

core/src/consensus/stake/action_data.rs
```diff
@@ -212,14 +269,23 @@ fn decode_map<K, V>(data: Option<&ActionData>) -> BTreeMap<K, V>
...
-            assert_eq!(2, record.item_count());
-            result.insert(key, value);
-        }
+        decode_map_impl(rlp)
+    } else {
+        Default::default()
+    }
+}
+
+fn decode_map_impl<K, V>(rlp: Rlp) -> BTreeMap<K, V>
```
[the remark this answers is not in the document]
sgkim126: I issued. https://github.com/CodeChain-io/codechain/issues/1575

core/src/consensus/tendermint/engine.rs
majecty: This line should be `let (start_of_the_current_term, end_of_the_prev_prev_term) = {`

sgkim126:
@majecty I updated the PR.

core/src/consensus/tendermint/engine.rs
```diff
@@ -254,3 +292,261 @@ impl ConsensusEngine for Tendermint {
...
+        .into_iter()
+        .map(|(address, (proposed, missed))| (address, Ratio::new(missed, proposed)))
+        .fold(BTreeMap::<Ratio<usize>, Vec<Address>>::new(), |mut map, (address, average_missed)| {
+            map.entry(average_missed).or_default().push(address);
+            map
+        });
+    for validators in sorted_validators.values() {
+        let reward = reduced_rewards / (u64::try_from(validators.len()).unwrap() + 1);
+        if reward == 0 {
+            break
```
majecty: What happens when `reduced_rewards < (u64::try_from(validators.len()).unwrap() + 1)` 
Should we give the remainder to someone?

sgkim126: No. we burns the remaining CCC.

core/src/consensus/tendermint/engine.rs
```diff
@@ -254,3 +292,261 @@ impl ConsensusEngine for Tendermint {
...
+
+    let mut missed_signatures = HashMap::<Address, (usize, usize)>::with_capacity(30);
+    let mut signed_blocks = HashMap::<Address, usize>::with_capacity(30);
+
+    let mut header = chain.block_header(&start_of_the_current_term.into()).unwrap();
+    loop {
+        for index in TendermintSealView::new(&header.seal()).bitset()?.true_index_iter() {
+            // FIXME: Change it after implementing ban
+            *signed_blocks.entry(authors[index]).or_default() += 1;
+        }
```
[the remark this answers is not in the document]
sgkim126: fixed

PR #1561, opened 2019-05-27
Send an email when there are error logs

### LINKED ISSUE #1557: Send an email when there are error logs

majecty:
This PR fixes #1557


codechain/config/mod.rs
```diff
@@ -691,6 +703,44 @@ impl Stratum {
...
+        }
+
+        Ok(())
+    }
+}
+
+impl Default for EmailAlarm {
+    fn default() -> Self {
+        Self {
+            disable: Some(true),
```
majecty: This is needed to support config files generated before this commit.

codechain/run_node.rs
```diff
@@ -238,7 +238,17 @@ pub fn run_node(matches: &ArgMatches) -> Result<(), String> {
             .expect("Current time should be later than unix epoch")
             .subsec_nanos() as usize,
     );
-    clogger::init(&LoggerConfig::new(instance_id)).expect("Logger must be successfully initialized");
+    let email_alarm_config = if config.email_alarm.disable != Some(true) {
```
sgkim126: Isn't the type of `disable` Option to support the configuration that was created before this commit, is it?
I think it should be `config.email_alarm.disable == Some(false)`.

util/logger/src/logger.rs
```diff
@@ -58,6 +59,7 @@ impl Logger {
             instance_id: config.instance_id,
             filter: builder.build(),
             stderr_is_tty,
+            email_alarm: email_alarm_config.as_ref().map(|config| EmailAlarm::new(config)),
```
sgkim126:
```rust
email_alarm: email_alarm_config.as_ref().map(EmailAlarm::new),
```

util/logger/src/logger.rs
```diff
@@ -41,10 +41,11 @@ pub struct Logger {
     instance_id: usize,
     filter: Filter,
     stderr_is_tty: bool,
+    email_alarm: Option<EmailAlarm>,
 }
 
 impl Logger {
-    pub fn new(config: &Config) -> Self {
+    pub fn new(config: &Config, email_alarm_config: &Option<EmailAlarmConfig>) -> Self {
```
sgkim126: How about taking `Option<EmailAlarm>` instead of `&Option<EmailAlarmConfig>`?

util/logger/src/logger.rs
```diff
@@ -103,8 +105,14 @@ impl Log for Logger {
                 target: log_target.to_string(),
                 message: log_message.to_string(),
                 timestamp,
-                thread_name,
+                thread_name: thread_name.to_string(),
```
sgkim126: `thread_name` is already `String`. How about using `clone` explicitly?

codechain/run_node.rs
majecty: Since the default instance of the EmailAlarm is created when it does not exist, we don't need to check the Option type of the `disable` field.

I found that other codes read the `disable` field using `Option::unwrap`. I'll change the code to use the `unwrap` for consistency.

majecty:
@sgkim126 I fixed them.

PR #1563, opened 2019-05-28
No need to fnd the merge base to check change

sgkim126:
Because it is the purpose of `COMMIT_RANGE` to check the changed files from the master branch, I think we don't need to find the merge base.


foriequal0:
It'll create false positive when the branch is stale. Long-standing document PR will eat up Travis queue unless we rebase for every change.

sgkim126:
@foriequal0 
How about `COMMIT_RANGE="origin/master..${TRAVIS_COMMIT}"`?
It will not make the false positive for the stale branch.

foriequal0:
I didn't know the differences between `..` and `...`. `..` would be right choice for this.

sgkim126:
https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#_commit_ranges
The triple-dot contains the commits in the base branch, while the double-dot doesn't.
In other words, `A...B` show all commits that are reachable from `A` or `B` except the commits that are reachable from the merge base of `A` and `B`, while `A..B` show the commits that are reachable from `B` but not reachable from `A`.

sgkim126:
@foriequal0 I modified the commit to use the double dots.

Issue #1511, opened 2019-05-03
Early stage of sync is slow with many 'AlreadyQueued' log


foriequal0:
```bash
#!/usr/bin/env bash
set -e

cargo build --release

export ENABLE_DELEGATIONS=true
export ENABLE_ORDER=true
export RUST_LOG=info,mio=warn,tokio=warn,hyper=warn,rpc=info

./target/release/codechain -c corgi --jsonrpc-port 8080 --port 3485 --enable-devel-api --no-miner --no-discovery --bootstrap-addresses $BOOTSTRAP
```

Sync on corgi is slow (less than few dozens per sec) on its early stage (less than 10000 blocks).
After that suspicious logs such as 'AlreadyQueued' is reduced and synchronization speed become more faster (few hundreds per sec).

It is not an accurate report. 
We should investigate more on this.
sgkim suggested `--no-tx-relay` would help to speed up, but it seems not.

foriequal0:
https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide
Inspect rocksdb stats.

sgkim126:
@foriequal0 Would you test after increasing [max_blocks_to_import](https://github.com/CodeChain-io/codechain/blob/00faf71b4e298010d5fdf60b0ebcdc1efdb0a0d2/core/src/client/importer.rs#L93)?

foriequal0:
It got a few times faster after #1542. It can sync about 40 blocks per seconds at the early stage of sync. However, it still have more rooms to improve compared to the later stage of the sync.

sgkim126:
Okay. Let's keep it open.

HoOngEe:
perf report for
current implmentation:
![perf_before_separate](https://user-images.githubusercontent.com/20704471/58160982-66162780-7cba-11e9-8507-a3d44c63d41a.png)

after the separation `import lock` into `header_import_lock` and `body_import_lock`:
![perf_separated](https://user-images.githubusercontent.com/20704471/58161037-80500580-7cba-11e9-9871-28388a279de5.png)

HoOngEe:
@Kais-DkM Is there any reason for using not headers[1..].iter() but headers.iter() ? in [this part](https://github.com/CodeChain-io/codechain/blob/82c62217df6be05ed3b995061d60149022cc9a18/sync/src/block/downloader/header.rs#L146)?
I think changing headers to headers[1..] would reduce overlaps.

Kais-DkM:
[...]

PR #1567, opened 2019-05-28
Fix slow sync

HoOngEe:
Now the total time taken for a node to synchronize with the corgi blocks is about 145 minutes.
Queued cache has introduced distinct from the downloaded cache in `HeaderDownloader`.
I failed to analyze the correct reason but decreasing `MAX_HEADERS_TO_IMPORT` down to 1,000 helps reducing sync time.
And lastly, in the current structure it doesn't help to spawn many threads because the cpu intensive jobs should be sequential. So I decreased it down to `2`.

The following image shows the cpu time spent during 50,000 block synchronization.
with_queued_cache

Now the main cpu intensive jobs are all related to EC calculation for verification.

### DIFF
--- a/core/src/verification/queue/mod.rs
+++ b/core/src/verification/queue/mod.rs
@@ -23,7 +23,6 @@ use std::sync::{Arc, Condvar as SCondvar, Mutex as SMutex};
 use std::thread::{self, JoinHandle};
 
 use cio::IoChannel;
-use num_cpus;
 use parking_lot::{Mutex, RwLock};
 use primitives::{H256, U256};
 
@@ -36,8 +35,8 @@ use crate::types::{BlockStatus as Status, VerificationQueueInfo as QueueInfo};
 const MIN_MEM_LIMIT: usize = 16384;
 const MIN_QUEUE_LIMIT: usize = 512;
 
-// maximum possible number of verification threads.
-const MAX_VERIFIERS: usize = 8;
+// number of verification threads.
+const NUM_VERIFIERS: usize = 2;
 
 /// Type alias for block queue convenience.
 pub type BlockQueue = VerificationQueue<kind::Blocks>;
@@ -150,10 +149,9 @@ impl<K: Kind> VerificationQueue<K> {
         let empty = Arc::new(SCondvar::new());
         let more_to_verify = Arc::new(SCondvar::new());
 
-        let num_verifiers = cmp::min(num_cpus::get(), MAX_VERIFIERS);
-        let mut verifier_handles = Vec::with_capacity(num_verifiers);
+        let mut verifier_handles = Vec::with_capacity(NUM_VERIFIERS);
 
-        for i in 0..num_verifiers {
+        for i in 0..NUM_VERIFIERS {
             let engine = engine.clone();
             let verification = verification.clone();
             let more_to_verify = more_to_verify.clone();
--- a/sync/src/block/downloader/header.rs
+++ b/sync/src/block/downloader/header.rs
@@ -45,6 +45,7 @@ pub struct HeaderDownloader {
     pivot: Pivot,
     request_time: Option<Instant>,
     downloaded: HashMap<H256, Header>,
+    queued: HashMap<H256, Header>,
     trial: usize,
 }
 
@@ -69,6 +70,7 @@ impl HeaderDownloader {
             },
             request_time: None,
             downloaded: HashMap::new(),
+            queued: HashMap::new(),
             trial: 0,
         }
     }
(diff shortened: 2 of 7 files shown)


core/src/verification/queue/mod.rs
```diff
@@ -37,7 +37,7 @@ const MIN_MEM_LIMIT: usize = 16384;
 const MIN_QUEUE_LIMIT: usize = 512;
 
 // maximum possible number of verification threads.
-const MAX_VERIFIERS: usize = 8;
+const MAX_VERIFIERS: usize = 2;
```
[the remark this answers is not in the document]
HoOngEe: The change applied

sync/src/block/downloader/header.rs
```diff
@@ -100,12 +102,17 @@ impl HeaderDownloader {
...
-        match self.downloaded.get(&self.pivot.hash) {
+        match self.queued.get(&self.pivot.hash) {
             Some(header) => header.clone(),
-            None => self.client.block_header(&BlockId::Hash(self.pivot.hash)).unwrap(),
+            None => self
+                .downloaded
+                .get(&self.pivot.hash)
+                .map(Clone::clone)
+                .or_else(|| self.client.block_header(&BlockId::Hash(self.pivot.hash)))
+                .unwrap(),
```
Kais-DkM: Why did you use function chaining here?
Using `match` seems to be more readable for me.

Kais-DkM:
Please fix the commit message.
queued chache -> queued cache

core/src/verification/queue/mod.rs
```diff
@@ -23,7 +23,6 @@ use std::sync::{Arc, Condvar as SCondvar, Mutex as SMutex};
 use std::thread::{self, JoinHandle};
 
 use cio::IoChannel;
-use num_cpus;
```
sgkim126: As far as I know, it's the only use case of `num_cpus`. You can remove this library.

sync/src/block/downloader/header.rs
HoOngEe: I changed it to `match` style.

sync/src/block/extension.rs
```diff
@@ -680,19 +680,23 @@ impl Extension {
...
             match self.client.import_header(header.clone().into_inner()) {
-                Err(BlockImportError::Import(ImportError::AlreadyInChain)) => exists.push(header.hash()),
+                Err(BlockImportError::Import(ImportError::AlreadyInChain)) => exists.push(hash),
                 // FIXME: handle import errors
                 Err(err) => {
                     cwarn!(SYNC, "Cannot import header({}): {:?}", header.hash(), err);
                     break
                 }
-                _ => {}
+                _ => queued.push(hash),
```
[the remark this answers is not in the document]
HoOngEe: Now I'm pushing hashes to the queue after importing error occurs.

sync/src/block/downloader/header.rs
```diff
@@ -143,7 +148,7 @@ impl HeaderDownloader {
         if self.best_hash == self.pivot.hash {
             ctrace!(SYNC, "Ignore received headers, pivot already reached the best hash");
         } else if first_header_hash == self.pivot.hash {
-            for header in headers.iter() {
+            for header in headers[1..].iter() {
```
sgkim126: Would you explain it?

core/src/verification/queue/mod.rs
HoOngEe: The other usage of `num_cpus` is in `util/kvdb-rocksdb`, so I removed this library from `core`

sync/src/block/downloader/header.rs
HoOngEe: Currently, the responded header data always makes one overlap. I made `queued` to reduce re-importing already queued headers but previous routine still generates one overlap every 128 requested headers. So I decided to iterate from the index 1.

sgkim126: @Kais-DkM Could you explain why the response overlaps one header?

Kais-DkM: You can see my comment [here](https://github.com/CodeChain-io/codechain/issues/1511#issuecomment-495173807).
In short, no specific reason for that.

sgkim126: @HoOngEe I think it's better to keep the original code.
If you really want to improve the performance of it, you should make the response exclude the overlapped one, not just skipping the first one.

HoOngEe: @sgkim126 Ok, I reverted it.

sgkim126:
@HoOngEe Please resolve conflicts.

HoOngEe:
I resolved the conficts

PR #1571, opened 2019-05-29
Send an email when a pnic occurred

majecty:
The email doesn't contain any information to specify the sender.
I'm trying to add private IP or public IP to the email contents.


majecty:
@sgkim126 I added the machine's IP in the mail.

sgkim126:
@majecty Would you apply this PR to rc-2.0.0 and rc-1.4.0?

majecty:
OK, I'll

PR #1574, opened 2019-05-29
Implement dynamic validator set

sgkim126:


sgkim126:
It depends on https://github.com/CodeChain-io/codechain/pull/1559 and https://github.com/CodeChain-io/codechain/pull/1445

sgkim126:
It depends on https://github.com/CodeChain-io/codechain/pull/1612 and https://github.com/CodeChain-io/codechain/pull/1598

core/src/consensus/validator_set/dynamic_validator.rs
```diff
@@ -0,0 +1,195 @@
...
+        }
+    }
+
+    fn next_block_proposer(&self, parent: &H256, view: u64) -> Option<Address> {
+        if let Some(validators) = self.validators_pubkey(*parent) {
+            let n_validators = validators.len();
+            let nonce = view as usize % n_validators;
+            Some(public_to_address(validators.get(n_validators - nonce - 1).unwrap()))
+        } else {
+            self.initial_list.next_block_proposer(parent, view)
```
foriequal0: The algorithm of `DynamicValidator::next_block_proposer` is different from `ValidatorList::next_block_proposer`. I think `initial_list` is more appropriate to be called `static_round_robin_validators` or `ValidatorList` is renamed to `StaticRoundRobinValidators`. However it is too long.

sgkim126: I agree. I need another name that specifies it's behavior, but I cannot think.

sgkim126: How about simply RoundRobinValidator?
I think it's easy to guess the validator set is static because the other one is `DynamicValidator`.

PR #1585, opened 2019-05-30
Clean up memories when syncing

### LINKED ISSUE #1576: High memory usage while syncing
During the block synchronization, codechain uses main memory over 8GB.
And the occupied memory was not cleared after the synchronization finished.

HoOngEe:
The commits in this pr piled upon the #1567
It closes #1576

### DIFF
--- a/core/src/verification/queue/mod.rs
+++ b/core/src/verification/queue/mod.rs
@@ -408,6 +408,7 @@ impl<K: Kind> VerificationQueue<K> {
                 *td -= score;
             }
         }
+        processing.shrink_to_fit();
         processing.is_empty()
     }
 
@@ -444,6 +445,7 @@ impl<K: Kind> VerificationQueue<K> {
                 new_verified.push_back(output);
             }
         }
+        processing.shrink_to_fit();
 
         self.verification.sizes.verified.fetch_sub(removed_size, AtomicOrdering::SeqCst);
         *verified = new_verified;
--- a/sync/src/block/downloader/body.rs
+++ b/sync/src/block/downloader/body.rs
@@ -69,6 +69,7 @@ impl BodyDownloader {
                 self.downloaded.insert(hash, body);
             }
         }
+        self.downloading.shrink_to_fit();
     }
 
     pub fn add_target(&mut self, header: &Header, parent: &Header) {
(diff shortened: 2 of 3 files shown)


sgkim126:
@HoOngEe Would you shrink `targets` too?

HoOngEe:
Before shrinking `targets` the memory usage reduced from about `9.0GB` to `5.3GB`.

HoOngEe:
I ran memory profiling tool valgrin --tool=massif, used massif-visualizer and got the following graph.
memory_profile

This time I got data of only 68000 blocks synchronization. So I cannot certain that the data from these two key parts: `header_queue` of importer, `downloaded` of `HeaderDownloader` would be shrunk later. The running time increased about 20 times before attaching heap memory profiler so that I cannot observe in full-time domain. Now I'll check `header_queue` and `body_queue`

HoOngEe:
`processing: RwLock<HashMap<H256, U256>>` also should be shrinked

sync/src/block/downloader/body.rs
```diff
@@ -111,6 +115,7 @@ impl BodyDownloader {
                 break
             }
         }
+        self.downloaded.shrink_to_fit();
         self.targets.drain(0..result.len());
```
sgkim126: Please shrink `targets` too.

HoOngEe: locally I shrank `targets`. In next force-push the change will be included.

HoOngEe:
I got a full-time synchronization data through last weekends.
full-time

The purple area, which was allocated in the `block_header_data` of `HeaderChain` implementing `HeaderProvider`, accounts for almost memory uncleared even after synchronizing old blocks.
It uses `HashMap` named `header_cache` internally. But there is no entry clearing routine for this.

sgkim126:
@HoOngEe I know still there is a memory to cleaning up. However, I think it's hard to fix the problem, there is no apparent time to clean the cache. It should change the implementation to LRU like implementation.
So I propose to merge this PR and keep #1576 open until we fix the memory issue completely.
How about you?

HoOngEe:
@sgkim126 Ok, let's merge this pr now and fix the cache implementation later.

core/src/verification/queue/mod.rs
```diff
@@ -429,6 +430,7 @@ impl<K: Kind> VerificationQueue<K> {
                 *td -= score;
             }
         }
+        processing.shrink_to_fit();
```
sgkim126: This seems redundant.

HoOngEe: Yes, It's redundant. I fixed it. Thank you.

sgkim126:
@HoOngEe LGTM. Please apply it to rc-1.4.0 too.

PR #1586, opened 2019-05-30
Proposal for how to handle ties

sgkim126:


spec/Dynamic-Validator.md
```diff
@@ -67,8 +67,9 @@ The election is a process that elects validators of a term according to the foll
 
 1. Select the candidates who deposited **MIN_DEPOSIT** or more.
 2. Pick **MAX_NUM_OF_VALIDATORS** candidates in order of the amount of received delegations.
-3. Select **MIN_NUM_OF_VALIDATORS** accounts; they become validators.
-4. Among the rest of them, drop the accounts that received less than **DELEGATION_THRESHOLD**; the remaining accounts become validators.
+3. If there are candidates who have a tie with dropouts, drop the candidates either.
+4. Select **MIN_NUM_OF_VALIDATORS** accounts; they become validators.
```
majecty: Should we change the "3." and "4."?
If there are 100 accounts which have been delegated the same 1% CCS, all the accounts will be dropped in the number 3.

majecty: On my second thought, the case would be a very rare case, it is okay to use PR's method.

spec/Dynamic-Validator.md
```diff
@@ -67,8 +67,9 @@ The election is a process that elects validators of a term according to the foll
 
 1. Select the candidates who deposited **MIN_DEPOSIT** or more.
 2. Pick **MAX_NUM_OF_VALIDATORS** candidates in order of the amount of received delegations.
-3. Select **MIN_NUM_OF_VALIDATORS** accounts; they become validators.
-4. Among the rest of them, drop the accounts that received less than **DELEGATION_THRESHOLD**; the remaining accounts become validators.
+3. If there are candidates who have a tie with dropouts, drop the candidates either.
```
ScarletBlue: @sgkim126 I am not sure what this sentence means.

PR #1590, opened 2019-05-31
Apply commits that fix slow sync to rc-1.4.0

sgkim126:


sgkim126:
@HoOngEe I think this fix can be applied to 1.4.0

PR #1591, opened 2019-05-31
Refactor Tendermint to introduce the dynamic validator set

sgkim126:


sgkim126:
@majecty Would you review it?

core/src/consensus/tendermint/worker.rs
```diff
@@ -1244,11 +1240,7 @@ impl Worker {
                 })
             }
 
-            let prev_block_hash = self
-                .client()
-                .block_header(&BlockId::Number((message.on.step.height as u64) - 1))
-                .expect("self.height - 1 == the best block number")
-                .hash();
+            let prev_block_hash = self.prev_block_hash();
```
majecty: Tendermint module receives the previous height's messages in several purposes.
First, if the more previous height's messages are collected, other nodes follow faster. Second, the collected previous height's pre-commits will be included in the current height's proposal. The more pre-commits make more profit.

Therefore, prev_block_hash may not be the same as the message's parent block hash. We should get the prev_block_hash using the message's height.

sgkim126: @majecty Thank you for your comment. I checked other parts too. Would you double-check it?

PR #1595, opened 2019-06-03
Add stake parameters to CommonParams

sgkim126:

PR #1598, opened 2019-06-03
Handle nomination expiration properly

sgkim126:
Please see the last commit only.
It depends on https://github.com/CodeChain-io/codechain/pull/1597, https://github.com/CodeChain-io/codechain/pull/1594 and https://github.com/CodeChain-io/codechain/pull/1595

### DIFF
--- a/core/src/consensus/solo/mod.rs
+++ b/core/src/consensus/solo/mod.rs
@@ -120,7 +120,8 @@ impl ConsensusEngine for Solo<CodeChainMachine> {
         for (address, reward) in rewards {
             self.machine.add_balance(block, &address, reward)?;
         }
-        self.machine.increase_term_id(block, last_term_finished_block_num)?;
+
+        stake::on_term_close(block.state_mut(), last_term_finished_block_num)?;
         Ok(())
     }
 
--- a/core/src/consensus/stake/mod.rs
+++ b/core/src/consensus/stake/mod.rs
@@ -23,7 +23,7 @@ use std::collections::HashMap;
 
 use ccrypto::Blake;
 use ckey::{public_to_address, recover, Address, Public, Signature};
-use cstate::{ActionHandler, StateResult, TopLevelState, TopState};
+use cstate::{ActionHandler, StateResult, TopLevelState, TopState, TopStateView};
 use ctypes::errors::{RuntimeError, SyntaxError};
 use ctypes::util::unexpected::Mismatch;
 use ctypes::{CommonParams, Header};
@@ -96,7 +96,20 @@ impl ActionHandler for Stake {
             Action::SelfNominate {
                 deposit,
                 metadata,
-            } => self_nominate(state, sender, sender_public, deposit, 0, 0, metadata),
+            } => {
+                let (current_term, nomination_ends_at) = {
+                    let metadata = state.metadata()?.expect("Metadata must exist");
+                    const DEFAULT_NOMINATION_EXPIRATION: u64 = 24;
+                    let current_term = metadata.current_term_id();
+                    let expiration = metadata
+                        .params()
+                        .map(CommonParams::nomination_expiration)
+                        .unwrap_or(DEFAULT_NOMINATION_EXPIRATION);
+                    let nomination_ends_at = current_term + expiration;
+                    (current_term, nomination_ends_at)
+                };
+                self_nominate(state, sender, sender_public, deposit, current_term, nomination_ends_at, metadata)
+            }
             Action::ChangeParams {
                 metadata_seq,
                 params,
@@ -188,8 +201,6 @@ fn self_nominate(
     nomination_ends_at: u64,
     metadata: Bytes,
 ) -> StateResult<()> {
-    // TODO: proper handling of get_current_term
-    // TODO: proper handling of NOMINATE_EXPIRATION
     let blacklist = Banned::load_from_state(state)?;
     if blacklist.is_banned(&sender) {
         return Err(RuntimeError::FailedToHandleCustomAction("Account is blacklisted".to_string()).into())
@@ -281,8 +292,8 @@ fn change_params(
     Ok(())
 }
 
-#[allow(dead_code)]
-pub fn on_term_close(state: &mut TopLevelState, current_term: u64) -> StateResult<()> {
+pub fn on_term_close(state: &mut TopLevelState, last_term_finished_block_num: u64) -> StateResult<()> {
+    let current_term = state.metadata()?.expect("The metadata must exist").current_term_id();
     // TODO: total_slash = slash_unresponsive(headers, pending_rewards)
     // TODO: pending_rewards.update(signature_reward(blocks, total_slash))
 
@@ -309,6 +320,8 @@ pub fn on_term_close(state: &mut TopLevelState, current_term: u64) -> StateResul
     revert_delegations(state, &reverted)?;
 
     // TODO: validators, validator_order = elect()
+
+    state.increase_term_id(last_term_finished_block_num)?;
     Ok(())
 }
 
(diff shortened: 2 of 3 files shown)


core/src/consensus/solo/mod.rs
```diff
@@ -121,6 +121,9 @@ impl ConsensusEngine for Solo<CodeChainMachine> {
             self.machine.add_balance(block, &address, reward)?;
         }
         self.machine.increase_term_id(block, last_term_finished_block_num)?;
+
+        let current_term = block.state().metadata()?.unwrap().current_term_id();
+        stake::on_term_close(block.state_mut(), current_term)?;
```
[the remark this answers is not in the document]
sgkim126: The last block changes the term. In other words, the term changes after executing the last block of the term, not before the first block of the next term. So you should call `on_term_close` with the changed term id.
Please fix the test cases.

sgkim126:
I stopped Travis. I'll run it after merging #1594 and #1595

core/src/consensus/solo/mod.rs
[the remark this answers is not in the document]
foriequal0: What about
```rust
        let current_term = block.state().metadata()?.unwrap().current_term_id();
        self.machine.increase_term_id(block, last_term_finished_block_num)?;
        stake::on_term_close(block.state_mut(), current_term)?;
```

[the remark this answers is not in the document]
sgkim126: Looks good, but it seems that `on_term_close` doesn't use the `last_term_finished_block_num`, so the order is not important.
I made it increase the term id after `on_term_close` and there is no need to pass the term id because `on_term_close` takes the state.

sgkim126:
I resolved the conflicts.

core/src/consensus/stake/mod.rs
```diff
@@ -892,6 +916,8 @@ mod tests {
         let delegator = public_to_address(&address_pubkey);
 
         let mut state = helpers::get_temp_state();
+        assert_eq!(Ok(()), state.update_params(0, CommonParams::default_for_test()));
```
foriequal0: Why don't we add this `state.update_params(0, CommonParams::default_for_test());` to `helpers::get_temp_state();`? or let calling `update_params()` create metadata automatically?
And `state.update_params()` is not a subject for this tests. I think it can just `.unwrap()` rather than `assert_eq!()`

core/src/consensus/solo/mod.rs
```diff
@@ -120,6 +120,8 @@ impl ConsensusEngine for Solo<CodeChainMachine> {
         for (address, reward) in rewards {
             self.machine.add_balance(block, &address, reward)?;
         }
+
+        stake::on_term_close(block.state_mut())?;
```
foriequal0: `on_term_close` is always paired with `self.machine.increase_term_id()`. Why don't we increase term id in the `on_term_close` ?

core/src/consensus/stake/mod.rs
```diff
@@ -860,7 +882,8 @@ mod tests {
         // TODO: change with stake.execute()
         self_nominate(&mut state, &address, &address_pubkey, 200, 0, 30, b"".to_vec()).unwrap();
 
-        let result = on_term_close(&mut state, 29);
+        let result = on_term_close(&mut state);
+        increase_term_id_until(&mut state, 30);
```
foriequal0: If `on_term_close` handles increasing term id, these two lines will be `term_close_until(...)`

core/src/consensus/stake/mod.rs
```diff
@@ -846,12 +857,23 @@ mod tests {
         assert!(result.is_err(), "Cannot self-nominate without a sufficient balance");
     }
 
+    fn increase_term_id_until(state: &mut TopLevelState, term_id: u64) {
```
foriequal0: If `on_term_close` handles increasing term id, this will be `term_close_until(...)`

core/src/consensus/stake/mod.rs
```diff
@@ -1174,7 +1212,8 @@ mod tests {
         stake.execute(&action.rlp_bytes(), &mut state, &delegator, &delegator_pubkey).unwrap();
 
         for current_term in 0..=released_at {
-            on_term_close(&mut state, current_term).unwrap();
+            on_term_close(&mut state).unwrap();
```
foriequal0: If we make `on_term_close` handle increasing term id, just calling `term_close_until(..., released_at)` will replace this loop.

core/src/consensus/stake/mod.rs
sgkim126: There is a test whether the initial state is a root of the empty trie.

core/src/consensus/solo/mod.rs
sgkim126: Looks good. I'll fix it.

core/src/consensus/stake/mod.rs
```diff
@@ -860,7 +884,7 @@ mod tests {
         // TODO: change with stake.execute()
         self_nominate(&mut state, &address, &address_pubkey, 200, 0, 30, b"".to_vec()).unwrap();
 
-        let result = on_term_close(&mut state, 29);
+        let result = on_term_close(&mut state, pseudo_term_to_block_num_calculator(29));
```
foriequal0: `pseudo_term_to_block_num_calculator` is somewhat misleading. Actual bulk term increasing thing is done with `increase_term_id_until`, `on_term_close` is increasing it just one step. I think turn `increase_term_id_until` into `term_close_until` or something

core/src/consensus/stake/mod.rs
foriequal0: How about making a local test helper?

core/src/consensus/stake/mod.rs
sgkim126: It doesn't increase the term bulky.
The second parameter of `on_term_close` is the block number of closing block.
And you cannot rename `increase_term_id_until` to `term_close_until`, because it doesn't close term, it merely increases term id.

foriequal0: I understand it now that `pseudo_term_to_block_num_calculator` is not actually related term id. It was confusing for a while.
I think increasing term id by closing the term is more realistic than calling `increase_term_id` directly.

core/src/consensus/stake/mod.rs
sgkim126: I think it doesn't have worth to make a helper function.
And please prefer to use `assert_eq` if the argument derives `Debug` and `PartialEq`.
You should read the backtrace if you use `unwrap`, because the test fails in `Option`.

core/src/consensus/stake/mod.rs
sgkim126: No. It doesn't close terms. It just increases term id on the metadata

core/src/consensus/stake/mod.rs
sgkim126: It is the purpose of this function to set the states for testing not calling `on_term_close`.

core/src/consensus/stake/mod.rs
foriequal0: Okay then. I'll comment like `// Arrange`, `// Action`, `// Assert` to separate sections to convey the intention of a test in that situation.

core/src/consensus/stake/mod.rs
foriequal0: Is there a meaning to differentiate closing the term from increasing the term id?

sgkim126: Yes. It is hard to create test cases if we close terms to set the states.

foriequal0: Okay.

PR #1599, opened 2019-06-04
Use the genesis network id at the email title

sgkim126:
The best way is using the current network id, but CodeChain cannot know
the current network id when creating an email alarm.
Instead of the current network id, it can get the genesis network id
easily. Because there is no use case to change the network id yet, it
would be a fine solution for now.

And this PR also does some trivial refactoring.

### DIFF
--- a/util/logger/src/email.rs
+++ b/util/logger/src/email.rs
@@ -3,13 +3,15 @@ use sendgrid::v3 as sendgrid;
 pub struct EmailAlarmConfig {
     pub to: String,
     pub sendgrid_key: String,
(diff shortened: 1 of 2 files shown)


util/logger/src/email.rs
```diff
@@ -35,7 +39,7 @@ impl EmailAlarm {
         let m = sendgrid::Message::new()
             .set_from(sendgrid::Email::new().set_email("no-reply@codechain.io"))
             // FIXME: fill the network id
```
majecty: Let's remove the FIXME comment.

PR #1606, opened 2019-06-05
Increase the timeout of term change test

sgkim126:
Because it waits term change, sometimes it takes longer than 5 seconds.


sgkim126:
@foriequal0 I'll merge it first because other PRs are pending because of it.

PR #1613, opened 2019-06-07
Change mergify not to check clahub

sgkim126:
Because mergify itself didn't sign the clahub, the merge commit that
mergify creates fail to pass clahub check. This makes mergify check only
Travis and approved reviews.
It is the task of the reviewers to make the author sign to clahub.


sgkim126:
I merged without waiting for Travis because it just changes the rule of mergify not about behavior.
For now, the reviewers MUST NOT approve the commit when the author doesn't sign the clahub.

PR #1619, opened 2019-06-12
Add the spec of chain_getPossibleAuthors

sgkim126:


spec/JSON-RPC.md
```diff
@@ -1561,6 +1562,39 @@ No parameters
 
 [Back to **List of methods**](#list-of-methods)
 
+## chain_getPossibleAuthors
+Returns the list of the account that can generate the blocks at the given block number.
```
ScarletBlue: Returns the list of accounts that can generate the blocks at the given block number.

spec/JSON-RPC.md
```diff
@@ -1561,6 +1562,39 @@ No parameters
...
+
+It returns `null` if anyone can generate the block.
+Only PoW and Solo chains can return `null`, other chains never return it.
```
ScarletBlue: Only PoW and Solo chains can return `null`. Other chains never return `null`.

spec/JSON-RPC.md
```diff
@@ -1561,6 +1562,39 @@ No parameters
...
+
+The possible authors of the genesis block are always the list that contains only the author of the genesis block, regardless of the chain types.
```
ScarletBlue: The possible authors of the genesis block are always in the list that contains only the author of the genesis block, regardless of the chain types.

PR #1623, opened 2019-06-13
Renew the nomination expiration of the validators at the term end

sgkim126:


foriequal0:
Should Any validator who doesn't want to be a validator anymore be jailed or wait delegators to revoke?

sgkim126:
@foriequal0 Yes.

PR #1625, opened 2019-06-13
Update the weights of the validators when closing blocks

sgkim126:


core/src/consensus/stake/action_data.rs
```diff
@@ -315,10 +315,9 @@ impl Validators {
         Ok(())
     }
 
-    #[allow(dead_code)]
-    pub fn update(&mut self, block_author: Address, min_delegation: StakeQuantity) {
+    pub fn update(&mut self, block_author: &Address, min_delegation: StakeQuantity) {
         for (weight, _deposit, pubkey) in self.0.iter_mut().rev() {
-            if public_to_address(pubkey) == block_author {
+            if public_to_address(pubkey) == *block_author {
```
foriequal0: It doesn't make a big difference in the asm output when it is built with a release flag when I've tested it, however, this looks more comfortable for me. You can just resolve this suggestion.
```suggestion
            if &public_to_address(pubkey) == block_author {
```

sgkim126: I also like to compare the reference, but the clippy complains some of them https://rust-lang.github.io/rust-clippy/master/index.html#op_ref
To make the code consistent, I usually deref the reference when comparing.

PR #1626, opened 2019-06-13
Fix a bug that Tendermint fails when a locked proposal is not imported

majecty:


sgkim126:
@majecty Please apply this fix to 1.4.x and 1.5.0.

majecty:
I sent PRs to apply this commit to 1.4.x and 1.5.0.

PR #1641, opened 2019-06-19
Implement the missing features of the dynamic validators

sgkim126:
and fixing bugs.

### DIFF
--- a/core/src/consensus/stake/mod.rs
+++ b/core/src/consensus/stake/mod.rs
@@ -31,7 +31,9 @@ use primitives::{Bytes, H256};
 use rlp::{Decodable, UntrustedRlp};
 
 pub use self::action_data::{Banned, Validator, Validators};
-use self::action_data::{Candidates, Delegation, IntermediateRewards, Jail, ReleaseResult, StakeAccount, Stakeholders};
+use self::action_data::{
+    Candidates, Delegation, Deposit, IntermediateRewards, Jail, ReleaseResult, StakeAccount, Stakeholders,
+};
 use self::actions::Action;
 pub use self::distribute::fee_distribute;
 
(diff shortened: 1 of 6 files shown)


core/src/consensus/tendermint/engine.rs
```diff
@@ -339,6 +346,23 @@ fn block_number_if_term_changed(
     Some(header.number())
 }
 
+fn inactive_validators(
```
foriequal0: Does this function assume that every validator would get a chance to be a proposer at least once in a term?

core/src/consensus/stake/mod.rs
```diff
@@ -384,8 +386,12 @@ pub fn ban(state: &mut TopLevelState, criminal: Address) -> StateResult<()> {
     let mut jailed = Jail::load_from_state(state)?;
     let mut validators = Validators::load_from_state(state)?;
 
-    candidates.remove(&criminal);
-    jailed.remove(&criminal);
+    let deposit = match (candidates.remove(&criminal), jailed.remove(&criminal)) {
+        (Some(_), Some(_)) => unreachable!(),
```
foriequal0: There should be a reason why this is unreachable.

core/src/consensus/tendermint/engine.rs
sgkim126: Yes, it is.

core/src/consensus/stake/mod.rs
sgkim126: fixed

core/src/consensus/tendermint/engine.rs
sgkim126: The current term is 3600 seconds; there must be 360 blocks if no missing round.
Because the validator must have been delegated at least 1% of the total stakes, the validator must have a chance per 100 blocks in this worst case.

PR #1645, opened 2019-06-21
Do not call verify_block_external in the header verification

majecty:
`verify_block_external` could use the parent block's state. In the
Tendermint consensus, `verify_block_external` uses the parent block's
state to calculate the current validator set.


majecty:
I'm fixing failed test cases.

majecty:
I fixed the test case.

majecty:
@sgkim126 Could you check this PR again? I fixed the integration test.

PR #1647, opened 2019-06-21
Do not use the same timestamp with the parent block

majecty:


sgkim126:
@majecty Please apply this commit to rc-1.5.0 and rc-1.4.x branch.

majecty:
Okay, I'll.

PR #1651, opened 2019-06-26
Change the versions of master branch to 0.0.0

sgkim126:
The version of the master branch doesn't make sense because we use the release branch.
I suggest fixing the master branch version to 0.0.0.


majecty:
LGTM

PR #1672, opened 2019-07-01
Add a comment about the safety of ratio comparison in our module

### LINKED ISSUE #1670: Add a comment about comparing Ratio is safe
When comparing two Ratio instances, `0/2` and `0/3` are not equal and can cause problems.
However, when used in CodeChain, there is no problem since both are used after being converted to `0/1`. As others can see this again, add a comment on the reason in the code.

https://github.com/CodeChain-io/codechain/blob/e039f3e4b0fb6b126182cf8ad56b7d53958603d4/core/src/consensus/tendermint/engine.rs#L474

HoOngEe:
It closes #1670

### DIFF
--- a/core/src/consensus/tendermint/engine.rs
+++ b/core/src/consensus/tendermint/engine.rs
@@ -472,6 +472,12 @@ fn give_additional_rewards<F: FnMut(&Address, u64) -> Result<(), Error>>(
     let sorted_validators = missed_signatures
         .into_iter()
         .map(|(address, (proposed, missed))| (address, Ratio::new(missed, proposed)))
+        // When one sees the Ratio crate's Order trait implementation, he/she can easily realize that the
+        // comparing routine is erroneous. It inversely compares denominators when the numerators are the same.
+        // That's problematic when it makes comparisons between ratios such as Ratio{numer: 0, denom:2} and Ratio{numer: 0, denom: 3}.
+        // But Ratio::new() calls Ratio::reduce() from within so that the numerator and the denominator are
+        // divided equally by the gcd. Therefore the comparison between above two instances would never happen and only
+        // comparisons between the reduced ones like Ratio{numer:0, denom:1} and Ratio{numer:0, denom: 1} should occur.
         .fold(BTreeMap::<Ratio<usize>, Vec<Address>>::new(), |mut map, (address, average_missed)| {
             map.entry(average_missed).or_default().push(address);
             map


core/src/consensus/tendermint/engine.rs
```diff
@@ -472,6 +472,12 @@ fn give_additional_rewards<F: FnMut(&Address, u64) -> Result<(), Error>>(
     let sorted_validators = missed_signatures
         .into_iter()
         .map(|(address, (proposed, missed))| (address, Ratio::new(missed, proposed)))
+        // When one sees the Ratio crate's Order trait implementation, he/she can easily realize the
```
ScarletBlue: // When one sees the Ratio crate's Order trait implementation, he/she can easily realize that the

core/src/consensus/tendermint/engine.rs
```diff
@@ -472,6 +472,12 @@ fn give_additional_rewards<F: FnMut(&Address, u64) -> Result<(), Error>>(
...
+        // comparing routine is erroneous. It inversely compares denominators when the numerators are same.
```
ScarletBlue: // comparing routine is erroneous. It inversely compares denominators when the numerators are the same.

core/src/consensus/tendermint/engine.rs
```diff
@@ -472,6 +472,12 @@ fn give_additional_rewards<F: FnMut(&Address, u64) -> Result<(), Error>>(
...
+        // That's problematic when it compares such as Ratio{numer: 0, denom:2} and Ratio{numer: 0, denom: 3}.
```
ScarletBlue: // That's problematic when it makes comparisons between ratios such as Ratio{numer: 0, denom:2} and Ratio{numer: 0, denom: 3}.

core/src/consensus/tendermint/engine.rs
```diff
@@ -472,6 +472,12 @@ fn give_additional_rewards<F: FnMut(&Address, u64) -> Result<(), Error>>(
...
+        // But the Ratio::new() calls Ratio::reduce() inside it so that the numerator and the denominator are
```
ScarletBlue: // But Ratio::new() calls Ratio::reduce() from within so that the numerator and the denominator are

core/src/consensus/tendermint/engine.rs
```diff
@@ -472,6 +472,12 @@ fn give_additional_rewards<F: FnMut(&Address, u64) -> Result<(), Error>>(
...
+        // divided equally by gcd. Therefore the comparison between above two instances will never happen and only
```
ScarletBlue: // divided equally by the gcd. Therefore the comparison between above two instances would never happen and only

core/src/consensus/tendermint/engine.rs
```diff
@@ -472,6 +472,12 @@ fn give_additional_rewards<F: FnMut(&Address, u64) -> Result<(), Error>>(
...
+        // the comparison between the reduced ones like Ratio{numer:0, denom:1} and Ratio{numer:0, denom: 1} will happen.
```
ScarletBlue: // comparisons between reduced ones like Ratio{numer:0, denom:1} and Ratio{numer:0, denom: 1} should occur.

core/src/consensus/tendermint/engine.rs
HoOngEe: Accepted

core/src/consensus/tendermint/engine.rs
HoOngEe: Accepted

core/src/consensus/tendermint/engine.rs
HoOngEe: Accepted

core/src/consensus/tendermint/engine.rs
HoOngEe: Accepted

core/src/consensus/tendermint/engine.rs
HoOngEe: Accepted

core/src/consensus/tendermint/engine.rs
HoOngEe: Accepted

majecty:
@HoOngEe If you applied all feedbacks from @ScarletBlue. Request him to recheck this PR.

HoOngEe:
@majecty I request him for approval.

PR #1678, opened 2019-07-10
Add the first test for dynamic validator (N->N')

majecty:
This PR depends on #1677.

### DIFF
--- a/test/src/e2e.dynval/dv.n'.test.ts
+++ b/test/src/e2e.dynval/dv.n'.test.ts
@@ -1,0 +1,139 @@
+// Copyright 2019 Kodebox, Inc.
+// This file is part of CodeChain.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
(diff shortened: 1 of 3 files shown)


test/src/e2e.dynval/dv.n'.test.ts
```diff
@@ -0,0 +1,375 @@
...
+            for (let i = 0; i < otherDynNodes.length; i++) {
+                await otherDynNodes[i].waitForTx(changeTx);
+            }
+
+            const margin = 1.1;
+
+            await waitForTermChange(1, TERM_SECONDS * margin);
+
+            {
+                const authors = await otherDynNodes[0].sdk.rpc.sendRpcRequest(
```
foriequal0: `const authors = await otherDynNodes[0].sdk.rpc.chain.getPossibleAuthors();`

majecty: `getPossibleAuthors` not in the current SDK version.

foriequal0: How about using `stake.getPossibleAuthors`?

majecty: I updated!.

PR #1681, opened 2019-07-11
Add test cases to the test suite "N to N-1"

HoOngEe:


majecty:
@HoOngEe Please fix the conflict.

test/src/e2e.dynval/dv.n-1.test.ts
```diff
@@ -0,0 +1,184 @@
...
+import * as stake from "codechain-stakeholder-sdk";
+import "mocha";
+
+import { validators as originalValidators } from "../../tendermint.dynval/constants";
+import { faucetAddress, faucetSecret } from "../helper/constants";
+import { PromiseExpect } from "../helper/promise";
+import { withNodes } from "./setup";
+import CodeChain from "../helper/spawn";
+
+async function getPossibleAuthors(
```
majecty: Other tests using `stake.getPossibleAuthors`, please use it.

test/src/e2e.dynval/dv.n-1.test.ts
```diff
@@ -0,0 +1,184 @@
...
+                blockNumber
+            );
+            const currentTermInitialBlockNumber =
+                termMedata!.lastTermFinishedBlockNumber + 1;
+            const validatorsAfter = await getPossibleAuthors(
+                checkingNode.sdk,
+                currentTermInitialBlockNumber
+            );
+
+            expect(validatorsAfter).contains.all.members(
```
majecty: Please check if Alice is missing.
Please check that the number of the validator list has decreased.

test/src/e2e.dynval/dv.n-1.test.ts
```diff
@@ -0,0 +1,184 @@
...
+            const validatorsAfter = await getPossibleAuthors(
+                checkingNode.sdk,
+                currentTermInitialBlockNumber
+            );
+
+            expect(termMedata!.currentTermId).to.be.equals(2);
+            expect(validatorsAfter).contains.all.members(
+                otherDynValidators.map(
+                    validator => validator.platformAddress.value
+                )
```
majecty: ditto.

test/src/e2e.dynval/setup.ts
```diff
@@ -57,6 +57,7 @@ export function withNodes(
             result.map(node => node.testFailed(this.currentTest!.fullTitle()));
         }
         await Promise.all(result.map(node => node.clean()));
+        options.promiseExpect.checkFulfilled();
```
foriequal0: `promiseExpect` is expected to be checked in `afterEach` of a test which it is declared.

test/src/e2e.dynval/dv.n-1.test.ts
```diff
@@ -0,0 +1,184 @@
...
+        }
+    });
+
+    describe("A node is imprisoned to jail", async function() {
+        it("alice should be dropped out from validator list", async function() {
+            const margin = 1.2;
+            const aliceNode = nodes[0];
+            const checkingNode = nodes[1];
+
+            await aliceNode.clean();
```
foriequal0: Alice is cleaned after the first term is started. It might lead to a flaky test. (What if Alice is the first validator and, there's a race between `clean()` and the term's first block? There are only 5 seconds to do `clean()`). This might be why you've left a comment `// not to make timing issue, wait until the term 3.`, and this is why we have `onBeforeEnable` in `withNode`'s parameter. 

Can you move `const nodes = withNodes(...)` to this inner `describe` from outer `describe`, and add `clean()` to `onBeforeEnable` for this case? 
It'll remove unnecessary waiting for term 3. 
You can share parameters like this to avoid duplications: https://github.com/CodeChain-io/codechain/pull/1690/files#diff-d5a0bb0ab12e8c7ae2dfbe269d53dceeR44. 
Duplicating it might also acceptable because it is not quite long.

test/src/e2e.dynval/dv.n-1.test.ts
HoOngEe: I replaced it.

test/src/e2e.dynval/dv.n-1.test.ts
```diff
@@ -0,0 +1,184 @@
...
+                    )
+                );
+            })
+            .catch(reject);
+    });
+}
+
+chai.use(chaiAsPromised);
+
+const [alice, , , ...otherDynValidators] = originalValidators;
```
foriequal0: Would you change this?
```ts
const otherDynValidators = originalValidators.slice(0, 7);
const betty = originalValidators[7];
```
It is more clear that there are 7 validators in `otherDynValidators`, and it is easier than counting empty commas.

test/src/e2e.dynval/setup.ts
HoOngEe: Ok, I added promiseExpect checking in `AfterEach` hook.

test/src/e2e.dynval/dv.n-1.test.ts
HoOngEe: I changed it to use slice

test/src/e2e.dynval/dv.n-1.test.ts
HoOngEe: I Introduced `onBeforeEnable`, thanks.

test/src/e2e.dynval/dv.n-1.test.ts
HoOngEe: I added both checks

test/src/e2e.dynval/dv.n-1.test.ts
```diff
@@ -0,0 +1,155 @@
...
+            })),
+            onBeforeEnable: async allDynNodes => {
+                const aliceNode = allDynNodes[0];
+                // Kill the alice node first to make alice not to participate in the term 1.
+                await aliceNode.clean();
+            }
+        });
+
+        it("alice should be dropped out from validator list", async function() {
+            const checkingNode = nodes[1];
```
majecty: Let's check Alice was in the validator list for the first time.

test/src/e2e.dynval/dv.n-1.test.ts
```diff
@@ -0,0 +1,89 @@
...
+            currentTermInitialBlockNumber
+        ))!.map(platformAddr => platformAddr.toString());
+
+        expect(termMedata!.currentTermId).to.be.equals(2);
+        expect(validatorsAfter.length).to.be.equals(otherDynValidators.length);
+        expect(validatorsAfter).not.to.includes(
+            alice.platformAddress.toString()
+        );
+        expect(
+            validatorsAfter!.map(platformAddr => platformAddr.toString())
```
Remagpie: You don't need `toString` here. Elements of `validatorsAfter` are already converted to a string.

HoOngEe: Thank you

test/src/e2e.dynval/dv.n-1.test.ts
HoOngEe: I added the routine.

test/src/e2e.dynval/dv.n-1.test.ts
```diff
@@ -0,0 +1,153 @@
...
+import { withNodes } from "./setup";
+
+chai.use(chaiAsPromised);
+
+const alice = originalValidators[0];
+const otherDynValidators = originalValidators.slice(1, 8);
+const allDynValidators = [alice, ...otherDynValidators];
+
+describe("Dynamic Validator N -> N-1", function() {
+    this.timeout(0);
```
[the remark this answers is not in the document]
HoOngEe: I added timeout

PR #1696, opened 2019-07-17
Add a nomination test

majecty:


test/src/e2e.dynval/dv.nomination.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+            const beforeCandidates = await stake.getCandidates(
+                otherDynNodes[0].sdk
+            );
+
+            expect(
+                beforeCandidates.map(candidate => candidate.pubkey.toString())
+            ).to.includes(H512.ensure(alice.publicKey).toString());
+
+            await otherDynNodes[0].waitForTermChange(4, 20 * margin * 3);
```
foriequal0: Hard-coded `time` seems unreliable. Can you specify `TERM_SECONDS` manually?

test/src/e2e.dynval/dv.nomination.test.ts
```diff
@@ -0,0 +1,99 @@
...
+            const [validators, banned, candidates, jailed] = await Promise.all([
+                stake.getValidators(otherDynNodes[0].sdk),
+                stake.getBanned(otherDynNodes[0].sdk),
+                stake.getCandidates(otherDynNodes[0].sdk),
+                stake.getJailed(otherDynNodes[0].sdk)
+            ]);
+
+            expect(
+                validators.map(validator => validator.pubkey.toString())
+            ).not.to.includes(H512.ensure(alice.publicKey).toString());
```
foriequal0: Why do you roundtrip pubkey like string -> H512 -> string?

test/src/e2e.dynval/setup.ts
```diff
@@ -180,9 +180,10 @@ export async function createNodes(options: {
                 )
                 .sign({
                     secret: faucetSecret,
-                    seq: faucetSeq2 + i,
+                    seq: faucetSeq2,
```
foriequal0: I've changed same line 🤔 How about this?
https://github.com/CodeChain-io/codechain/pull/1703/files#diff-2ad9cd6449c41e2610357eef55525900R190

test/src/e2e.dynval/setup.ts
```diff
@@ -236,6 +236,39 @@ export async function createNodes(options: {
...
+    validator: ValidatorConfig["signer"],
+    deposit: number
+): Promise<H256> {
+    const tx = stake.createSelfNominateTransaction(sdk, deposit, "").sign({
+        secret: validator.privateKey,
+        seq: await sdk.rpc.chain.getSeq(validator.platformAddress),
+        fee: 10
+    });
+
+    return await sdk.rpc.chain.sendSignedTransaction(tx);
```
Remagpie: How about calling `waitForTx` too?

test/src/e2e.dynval/setup.ts
majecty: Looks good. It's so functional!

test/src/e2e.dynval/setup.ts
majecty: To do it, we need to receive a value of type `CodeChain` that is a very heavy object in arguments.
Is it too complicated object to pass to this function?

Remagpie: `changeParams` already receives the `CodeChain` instance.
It's too big component though. Let's leave it as is for now.

PR #1686, opened 2019-07-15
Add test cases to the test suite "changeParams"

HoOngEe:

PR #1697, opened 2019-07-17
Add test cases to the test suite "changeParams"

Remagpie:
- Added test for changing the maximum number of validators
- Added test for changing the maximum size of candidate metadata
Depends on: https://github.com/CodeChain-io/codechain/pull/1686

### DIFF
--- a/test/src/e2e.dynval/dv.changeParams.test.ts
+++ b/test/src/e2e.dynval/dv.changeParams.test.ts
@@ -36,6 +36,9 @@ describe("Change commonParams", function() {
     const promiseExpect = new PromiseExpect();
     const nodes = withNodes(this, {
         promiseExpect,
+        overrideParams: {
+            maxCandidateMetadataSize: 128
+        },
         validators: allDynValidators.map((signer, index) => ({
             signer,
             delegation: 5_000,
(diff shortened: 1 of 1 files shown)


test/src/e2e.dynval/dv.changeParams.test.ts
```diff
@@ -221,4 +220,95 @@ describe("Change commonParams", function() {
...
+            );
+        }
+
+        it("Should apply larger metadata limit after increment", async function() {
+            const [alice] = allDynValidators;
+            const checkingNode = nodes[0];
+            const overrideParams = {
+                maxCandidateMetadataSize: 256
+            };
+            const changeTxHash = await changeParams(checkingNode, 1, {
```
foriequal0: See: https://github.com/CodeChain-io/codechain/pull/1686#discussion_r305658182

test/src/e2e.dynval/dv.changeParams.test.ts
```diff
@@ -221,4 +220,95 @@ describe("Change commonParams", function() {
...
+            }
+        });
+
+        it("Should apply smaller metadata limit after decrement", async function() {
+            const [alice] = allDynValidators;
+            const checkingNode = nodes[0];
+            const overrideParams = {
+                maxCandidateMetadataSize: 64
+            };
+            const changeTxHash = await changeParams(checkingNode, 1, {
```
foriequal0: ditto

test/src/e2e.dynval/dv.changeParams.test.ts
Remagpie: Fixed it.

test/src/e2e.dynval/dv.changeParams.test.ts
```diff
@@ -202,4 +202,101 @@ describe("Change commonParams", function() {
...
+            const margin = 1.2;
+
+            this.slow(termSeconds * margin * 1000);
+            this.timeout(termSeconds * 2 * 1000);
+
+            const [alice] = allDynValidators;
+            const checkingNode = nodes[0];
+            const changeTxHash = await changeParams(checkingNode, 1, {
+                ...defaultParams,
+                maxCandidateMetadataSize: 256
```
majecty: I think it would be nice to give `maxCandidateMetadataSize` as a parameter directly to the function `withNodes` as 128. The default value is 128, but I hope the test will work even if the default changes later.

test/src/e2e.dynval/dv.changeParams.test.ts
```diff
@@ -199,7 +199,159 @@ describe("Change commonParams", function() {
...
+            const margin = 1.2;
+
+            this.slow(termSeconds * margin * 1000);
+            this.timeout(termSeconds * 2 * 1000);
+
+            const [alice] = allDynValidators;
+            const checkingNode = nodes[0];
+            const changeTxHash = await changeParams(checkingNode, 1, {
+                ...defaultParams,
+                maxCandidateMetadataSize: 64
```
majecty: ditto

test/src/e2e.dynval/dv.changeParams.test.ts
Remagpie: How do you think about using `defaultParams.maxCandidateMetadataSize * 2` or similar instead of constant value?

Remagpie: I think adding `maxCandidateMetadataSize` to `withNodes` is good too. I'll fix it this way.

PR #1717, opened 2019-07-25
Fix the crash that is caused when there is only one block in a term

majecty:


core/src/miner/miner.rs
```diff
@@ -474,12 +475,26 @@ impl Miner {
...
+            match self.engine.seals_internally() {
+                Some(false) => panic!("If a signer is not prepared, prepare_block should not be called"),
+                None => panic!("Exteranl sealing is not seals_first"),
+                Some(true) => {}
+            };
+            let seal = self.engine.generate_seal(None, &parent_header.decode());
+            if let Some(seal_bytes) = seal.seal_fields() {
+                open_block.seal(self.engine.borrow(), seal_bytes).expect("Sealing always success");
+            } else {
+                panic!("Seal should not be none")
```
majecty: I'm not confident that panic is better or not than throwing an error.

core/src/miner/miner.rs
```diff
@@ -580,37 +595,34 @@ impl Miner {
...
-                        .seal(&*self.engine, seal)
-                        .map(|sealed| chain.import_sealed_block(&sealed).is_ok())
-                        .unwrap_or_else(|e| {
-                            cwarn!(MINER, "ERROR: seal failed when given internally generated seal: {}", e);
-                            false
-                        })
-                }
-            }
-            None => {
+        if self.engine.seals_internally() != Some(true) {
```
majecty: This code is hard to read. Please tell me if you don't understand it.

majecty:
@foriequal0 ping

PR #1728, opened 2019-07-31
Automatically create a ReportDoubleVote transaction

Remagpie:
This patch only covers double votes for precommit and prevote messages.

### DIFF
--- a/core/src/consensus/signer.rs
+++ b/core/src/consensus/signer.rs
@@ -16,7 +16,7 @@
 
 use std::sync::Arc;
 
-use ckey::{Address, Public, SchnorrSignature};
+use ckey::{Address, Public, SchnorrSignature, Signature};
 use ckeystore::DecryptedAccount;
 use primitives::H256;
 
@@ -77,9 +77,27 @@ impl EngineSigner {
         Ok(result)
     }
 
+    /// Sign a message hash with ECDSA.
+    pub fn sign_ecdsa(&self, hash: H256) -> Result<Signature, AccountProviderError> {
+        let address = self.signer.map(|(address, _public)| address).unwrap_or_else(Default::default);
+        let result = match &self.decrypted_account {
+            Some(account) => account.sign(&hash)?,
+            None => {
+                let account = self.account_provider.get_unlocked_account(&address)?;
+                account.sign(&hash)?
+            }
+        };
+        Ok(result)
+    }
+
     /// Public Key of signer.
     pub fn public(&self) -> Option<&Public> {
         self.signer.as_ref().map(|(_address, public)| public)
+    }
+
+    /// Address of signer.
+    pub fn address(&self) -> Option<&Address> {
+        self.signer.as_ref().map(|(address, _)| address)
     }
 
     /// Check if the given address is the signing address.
--- a/test/src/e2e.dynval/doubleVote.test.ts
+++ b/test/src/e2e.dynval/doubleVote.test.ts
@@ -1,0 +1,75 @@
+// Copyright 2019 Kodebox, Inc.
+// This file is part of CodeChain.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import * as chai from "chai";
+import { expect } from "chai";
+import * as chaiAsPromised from "chai-as-promised";
+import * as stake from "codechain-stakeholder-sdk";
+import "mocha";
+
+import { Mock } from "../helper/mock";
+import { validators } from "../../tendermint.dynval/constants";
+import { PromiseExpect } from "../helper/promise";
+import { setTermTestTimeout, withNodes } from "./setup";
+
+chai.use(chaiAsPromised);
+
+describe("Double vote detection", function() {
+    const promiseExpect = new PromiseExpect();
+
+    const [alice, betty, charlie] = validators;
+
+    const { nodes } = withNodes(this, {
+        promiseExpect,
+        validators: [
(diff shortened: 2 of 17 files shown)


core/src/consensus/signer.rs
```diff
@@ -77,11 +77,29 @@ impl EngineSigner {
         Ok(result)
     }
 
+    /// Sign a message hash with ECDSA.
+    pub fn sign_ecdsa(&self, hash: H256) -> Result<Signature, AccountProviderError> {
```
foriequal0: Why not 'hash: &H256`?

test/src/helper/mock/tendermintMessage.ts
```diff
@@ -0,0 +1,257 @@
...
+
+function readOptionalRlp<T>(bytes: [] | [Buffer], decoder: (b: Buffer) => T) {
+    if (bytes.length === 0) {
+        return null;
+    } else {
+        return decoder(bytes[0]);
+    }
+}
+
+function readUIntRLP(bytes: Buffer) {
```
foriequal0: Would you change the function name? It is not related to the RLP.

test/src/helper/mock/tendermintMessage.ts
```diff
@@ -0,0 +1,257 @@
...
+                    height: readUIntRLP(decoded[1]),
+                    view: readUIntRLP(decoded[2]),
+                };
+                break;
+            }
+            default: {
+                throw new Error(`Unexpected message id ${id}`);
+            }
+        }
+        Emitter.emit(message.type, message);
```
foriequal0: `emit` in `fromBytes` is awkward.

test/src/helper/mock/index.ts
```diff
@@ -18,6 +18,7 @@ import { SignedTransaction } from "codechain-sdk/lib/core/SignedTransaction";
 import { BlockSyncMessage, Emitter, IBodiesq, IHeadersq, MessageType, ResponseMessage } from "./blockSyncMessage";
 import { Header } from "./cHeader";
 import { P2pLayer } from "./p2pLayer";
+import { Emitter as TendermintEmitter, TendermintMessage } from "./tendermintMessage";
```
foriequal0: `TendermintEmitter` is not used in this commit.

test/src/helper/mock/tendermintMessage.ts
```diff
@@ -0,0 +1,257 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+import { H256, U64 } from "codechain-primitives";
+import { EventEmitter } from "events";
+import { compressSync, uncompressSync } from "snappy";
+
+const RLP = require("rlp");
+
+export const Emitter = new EventEmitter();
```
foriequal0: Are you sure about a singleton `Emitter`?

test/src/e2e.dynval/setup.ts
```diff
@@ -326,7 +326,7 @@ export const defaultParams = {
     minSetShardOwnersCost: 10,
     minSetShardUsersCost: 10,
     minWrapCccCost: 10,
-    minCustomCost: 10,
+    minCustomCost: 0,
```
foriequal0: Would you use `overrideParams` in `withNodes`?

test/src/e2e.dynval/doubleVote.test.ts
```diff
@@ -32,15 +33,15 @@ describe("Double vote detection", function() {
...
+    const { nodes, initialParams } = withNodes(this, {
         promiseExpect,
         validators: [
             { signer: alice, delegation: 5000, deposit: 10_000_000 - 0 },
             { signer: betty, delegation: 5000, deposit: 10_000_000 - 1 },
             { signer: charlie, delegation: 5000, deposit: 10_000_000 - 2 }
         ],
         overrideParams: {
-            termSeconds: 20,
+            termSeconds: 15,
```
foriequal0: Why do you override it when they are not different with the default?

test/src/e2e.dynval/doubleVote.test.ts
```diff
@@ -52,19 +53,53 @@ describe("Double vote detection", function() {
         mock = new Mock("0.0.0.0", aliceNode.port, aliceNode.sdk.networkId);
     });
 
-    it("Should create ReportDoubleVote transaction if double vote is detected", async function() {
-        const termWaiter = setTermTestTimeout(this, { terms: 1 });
+    it("Should report if double vote for prevote is detected", async function() {
+        const termWaiter = setTermTestTimeout(this, {
+            terms: 1,
+            params: initialParams
```
foriequal0: ditto.

test/src/e2e.dynval/doubleVote.test.ts
```diff
@@ -52,19 +53,53 @@ describe("Double vote detection", function() {
...
+        const banned = await stake.getBanned(aliceNode.sdk);
+        expect(banned.map(b => b.toString())).to.include(
+            betty.platformAddress.toString()
+        );
+    });
+
+    it("Should report if double vote for precommit is detected", async function() {
+        const termWaiter = setTermTestTimeout(this, {
+            terms: 1,
+            params: initialParams
```
foriequal0: ditto.

core/src/consensus/signer.rs
Remagpie: To keep consistency with the function `sign` in the same file.
I wonder why it was implemented like that though.

test/src/helper/mock/tendermintMessage.ts
Remagpie: How about `readUIntBE`?

test/src/helper/mock/tendermintMessage.ts
Remagpie: I agree, but the other part of `mock` is implemented like this.
I think refactoring this should be in separated PR since it's too large to fit in this one.

test/src/e2e.dynval/setup.ts
Remagpie: Why? I thought making this cost to `0` would make our test closer to the real-world environment.

test/src/e2e.dynval/doubleVote.test.ts
Remagpie: I'll remove this one.

test/src/helper/mock/tendermintMessage.ts
Remagpie: No, I'm not. But I have no better idea without refactoring the whole mock structure.

Issue #1724, opened 2019-07-30
Seal field is empty in the `calculate_pending_rewards_of_the_previous_term` function


majecty:
Crash log

```
  10:     0x55b1ee0c4946 - core::option::expect_failed::he602aa67d6d2d670
                        at src/libcore/option.rs:1036
  11:     0x55b1ec52e64c - core::option::Option<T>::expect::h2c96e71edb287caf
                        at /rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c/src/libcore/option.rs:314
  12:     0x55b1ec77f1ce - codechain_core::consensus::tendermint::types::TendermintSealView::bitset::hbf54f677e638737d
                        at core/src/consensus/tendermint/types.rs:202
  13:     0x55b1ec8a348c - codechain_core::consensus::tendermint::engine::calculate_pending_rewards_of_the_previous_term::h8ad064a75756d7a6
                        at core/src/consensus/tendermint/engine.rs:403
  14:     0x55b1ec62285b - codechain_core::consensus::tendermint::engine::<impl codechain_core::consensus::ConsensusEngine for codechain_core::consensus::tendermint::Tendermint>::on_close_block::hea5d425445cba026
                        at core/src/consensus/tendermint/engine.rs:224
  15:     0x55b1ec399772 - codechain_core::block::OpenBlock::close_and_lock::ha19574b5840e3bf2
                        at core/src/block.rs:268
  16:     0x55b1ec789720 - codechain_core::block::enact::h44d7d7d2380cd396
                        at core/src/block.rs:512
  17:     0x55b1ec7afd40 - codechain_core::client::importer::Importer::check_and_close_block::h29ee3dd1a94d735f
                        at core/src/client/importer.rs:287
  18:     0x55b1ec7ade72 - codechain_core::client::importer::Importer::import_verified_blocks::h9a089cbfa80daac8
                        at core/src/client/importer.rs:119
  19:     0x55b1ec5589fe - codechain_core::client::client::Client::import_verified_blocks::hfc25cc178364e20d
                        at core/src/client/client.rs:212
  20:     0x55b1ec2f3f86 - <codechain_core::service::ClientIoHandler as codechain_io::IoHandler<codechain_core::service::ClientIoMessage>>::message::hcd9950eac567d4aa
                        at core/src/service.rs:95
  21:     0x55b1ec27eb81 - codechain_io::service::IoChannel<Message>::send_sync::hf6598a3907377ca8
[...]

PR #1733, opened 2019-08-08
Populate the seal in OpenBlock for external blocks

Remagpie:
Fixes https://github.com/CodeChain-io/codechain/issues/1724


test/src/e2e.dynval/dv.oneblockoneterm.test.ts
```diff
@@ -23,7 +23,7 @@ import { withNodes } from "./setup";
 
 chai.use(chaiAsPromised);
 
-const alice = originalDynValidators[0];
```
majecty: Let's add a comment why we added the "bob".
It is not clear why bob is needed.

PR #1734, opened 2019-08-08
Fix the command for formatting the test scripts

Remagpie:


majecty:
@Remagpie The only manually modified file is the package.json?
Did the tslint or prettier change other files?

Remagpie:
@majecty Yes. All the files other than the `package.json` are changed by the formatters.

PR #1736, opened 2019-08-13
Assert that transaction is executed in the expected term

majecty:
Currently, some tests are sometiles failed. I suspect that term is
changed unintentionally. This error log will help to find the code
that needs more timeout.

### DIFF
--- a/test/src/e2e.dynval/setup.ts
+++ b/test/src/e2e.dynval/setup.ts
@@ -464,3 +464,22 @@ export function setTermTestTimeout(
         }
     };
 }
+
+export async function termThatIncludeTransaction(
+    sdk: SDK,
+    txHash: H256
+): Promise<number> {
+    const transaction = await sdk.rpc.chain.getTransaction(txHash);
+    const minedBlock = transaction!.blockNumber!;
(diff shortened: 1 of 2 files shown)


foriequal0:
If you want to assert that a transaction is executed in a certain term, it would be better to check directly (by getting termId of tx's block), rather than relying on uncertain proving method.

majecty:
@foriequal0 I updated the PR.

test/src/e2e.dynval/setup.ts
```diff
@@ -464,3 +464,36 @@ export function setTermTestTimeout(
...
 }
+
+export async function termThatIncludeTransaction(
+    sdk: SDK,
+    txHash: H256
+): Promise<number> {
+    // Diable lint rule to extract let and const variable from Promise.all.
+    // tslint:disable-next-line:prefer-const
+    let [termMetadata, transaction] = await Promise.all([
+        stake.getTermMetadata(sdk),
```
foriequal0: You can utilize `blockNumber` parameter

majecty: Good!
I changed the code to use the blockNumber parameter.

PR #1749, opened 2019-08-28
Fix a memory leak in header downloader's downloaded

majecty:
The header downloader has two queues. The first one is the downloaded
queue, and the second one is the "queued" queue. If a header is
downloaded, downloader enqueues the header to the downloaded
queue. When starting to import the header, the downloader removes the
header from the downloaded queue and enqueues the header to the
"queued" queue. When the importing is done, the downloader removes the
header from the "queued" queue.

However, if a header(let's call it A) is downloaded and is imported
already, A is not removed from the downloaded queue forever. This
commit removes A from the downloaded header.


majecty:
@HoOngEe I added it.

majecty:
@Remagpie Please approve it!

PR #1750, opened 2019-08-28
Do not request older headers if re-organization is not occurred

majecty:
Header downloader has "best_hash," which is peer's best hash, and
"pivot," which is the best hash of received headers from a peer.

When header downloader receives headers from a peer, it checks the
first header's hash and the pivot's hash. If they are not the same,
the downloader concludes that there is a reorganization and request
older headers.

However, if the pivot is changed between header request and response,
the header downloader requests older hashes even though there is no
re-organization. This commit changes the header downloader check
re-organization more concisely.

### DIFF
--- a/sync/src/block/downloader/header.rs
+++ b/sync/src/block/downloader/header.rs
@@ -142,7 +142,10 @@ impl HeaderDownloader {
     /// Imports headers and mark success
     /// Expects importing headers matches requested header
     pub fn import_headers(&mut self, headers: &[Header]) {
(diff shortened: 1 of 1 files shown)


Remagpie:
I think there might be more memory leaks related to the `update` method in the header downloader.
I haven't considered the case where the pivot is updated other than when it receives the header response.

majecty:
@Remagpie Thanks for your comment.

I checked the `downloaded` and `queued` again. Since if a header is imported or already imported, `mark_as_imported` is always called, all imported `downloaded` and `queued` is cleared always.

PR #1755, opened 2019-08-29
Replace hashmap cache to LRU cache

### LINKED ISSUE #1576: High memory usage while syncing

HoOngEe:
it closes #1576

### DIFF
--- a/core/src/blockchain/body_db.rs
+++ b/core/src/blockchain/body_db.rs
@@ -19,6 +19,7 @@ use std::mem;
 use std::sync::Arc;
 
 use kvdb::{DBTransaction, KeyValueDB};
+use lru_cache::LruCache;
 use parking_lot::{Mutex, RwLock};
 use primitives::{Bytes, H256};
 use rlp::RlpStream;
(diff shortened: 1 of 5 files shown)


core/src/blockchain/headerchain.rs
```diff
@@ -99,7 +100,7 @@ impl HeaderChain {
             best_header_hash: RwLock::new(best_header_hash),
             best_proposal_header_hash: RwLock::new(best_proposal_header_hash),
 
-            header_cache: RwLock::new(HashMap::new()),
+            header_cache: Mutex::new(LruCache::new(1000)),
```
foriequal0: Ditto

core/src/blockchain/body_db.rs
```diff
@@ -48,7 +49,7 @@ impl BodyDB {
     /// Create new instance of blockchain from given Genesis.
     pub fn new(genesis: &BlockView, db: Arc<KeyValueDB>) -> Self {
         let bdb = Self {
-            body_cache: RwLock::new(HashMap::new()),
+            body_cache: Mutex::new(LruCache::new(1000)),
```
foriequal0: Would you separate the constant out?

core/src/blockchain/headerchain.rs
HoOngEe: Done

core/src/blockchain/body_db.rs
HoOngEe: Ok, I applied

PR #1792, opened 2019-09-24
Stop generating blocks if the consensus cannot make a seal

majecty:
If the Tendermint consensus receives an outdated block request, it returns
an empty seal. The miner should stop block generation if the seal is empty.

### DIFF
--- a/core/src/miner/miner.rs
+++ b/core/src/miner/miner.rs
@@ -451,7 +451,7 @@ impl Miner {
         &self,
         parent_block_id: BlockId,
         chain: &C,
-    ) -> Result<(ClosedBlock, Option<H256>), Error> {
+    ) -> Result<Option<(ClosedBlock, Option<H256>)>, Error> {
         let (transactions, mut open_block, original_work_hash, block_number) = {
             let sealing_work = self.sealing_work.lock();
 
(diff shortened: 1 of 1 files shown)


core/src/miner/miner.rs
```diff
@@ -491,7 +491,7 @@ impl Miner {
             if let Some(seal_bytes) = seal.seal_fields() {
                 open_block.seal(self.engine.borrow(), seal_bytes).expect("Sealing always success");
             } else {
-                panic!("Seal should not be none")
+                return Err(BlockError::InvalidSeal.into())
```
foriequal0: `InvalidSeal` usually used for block validating functions. How about returning `Some(None)`?

majecty: You mean changing the function signature to return `Result<Option<(ClosedBlock, Option<H256>)>, Error>`

foriequal0: Yes

majecty: I changed it.

core/src/miner/miner.rs
```diff
@@ -837,13 +840,17 @@ impl MinerService for Miner {
         let parent_block_number = chain.block_header(&parent_block).expect("Parent is always exist").number();
         if self.requires_reseal(parent_block_number) {
             let (block, original_work_hash) = match self.prepare_block(parent_block, chain) {
-                Ok((block, original_work_hash)) => {
+                Ok(Some((block, original_work_hash))) => {
```
foriequal0: BTW, `block` and `original_work_hash` is duplicated 3 times. It can be reduced with
```rust
let (block, original_work_hash) = match ... {
  Ok(Some(prepared @ (block, _))) => { 
    ...
    prepared
  }
  ...
};
```

Issue #1773, opened 2019-09-17
Don't assume that valid proposal is always one in one round


majecty:
If a proposer double voted in one round, there could be two or more proposal blocks in the round.
One of the blocks could be finalized in the height.

kseo:
@majecty what happens if there are two proposal blocks in one round?

majecty:
I'm not sure what is the effect of it. I created this issue since the code depends on an invalid assumption.

majecty:
If there are two proposal blocks in one round, the finalization time in a network disaster will be increased.

PR #1793, opened 2019-09-24
Handle multiple proposals in a view correctly in Tendermint

majecty:
The previous code assumes that there is only one proposal in a
view. Tendermint was always using the first proposal even though there
are many proposals in a view. This commit makes Tendermint find the
proper proposal for that situation.

Fixes https://github.com/CodeChain-io/codechain/issues/1773

### DIFF
--- a/core/src/consensus/tendermint/vote_collector.rs
+++ b/core/src/consensus/tendermint/vote_collector.rs
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{BTreeMap, HashMap};
 use std::iter::Iterator;
 
 use ckey::SchnorrSignature;
@@ -35,7 +35,7 @@ pub struct VoteCollector {
 struct StepCollector {
     voted: HashMap<usize, ConsensusMessage>,
     block_votes: HashMap<Option<H256>, BTreeMap<usize, SchnorrSignature>>,
-    messages: HashSet<ConsensusMessage>,
+    messages: Vec<ConsensusMessage>,
 }
 
 #[derive(Debug)]
@@ -64,7 +64,8 @@ impl StepCollector {
     /// Returns Some(&Address) when validator is double voting.
     fn insert(&mut self, message: ConsensusMessage) -> Option<DoubleVote> {
         // Do nothing when message was seen.
-        if self.messages.insert(message.clone()) {
+        if !self.messages.contains(&message) {
+            self.messages.push(message.clone());
             if let Some(previous) = self.voted.insert(message.signer_index(), message.clone()) {
                 // Bad validator sent a different message.
                 return Some(DoubleVote {
@@ -217,11 +218,11 @@ impl VoteCollector {
     }
 
     pub fn get_all(&self) -> Vec<ConsensusMessage> {
-        self.votes.iter().flat_map(|(_round, collector)| collector.messages.iter()).cloned().collect()
+        self.votes.iter().flat_map(|(_round, collector)| collector.messages.clone()).collect()
     }
 
     pub fn get_all_votes_in_round(&self, round: &VoteStep) -> Vec<ConsensusMessage> {
-        self.votes.get(round).map(|c| c.messages.iter().cloned().collect()).unwrap_or_default()
+        self.votes.get(round).map(|c| c.messages.clone()).unwrap_or_default()
     }
 
     pub fn get_all_votes_and_indices_in_round(&self, round: &VoteStep) -> Vec<(usize, ConsensusMessage)> {
--- a/core/src/consensus/tendermint/vote_collector.rs
+++ b/core/src/consensus/tendermint/vote_collector.rs
@@ -207,6 +207,15 @@ impl VoteCollector {
             .unwrap_or_else(Vec::new)
     }
 
+    pub fn has_votes_for(&self, round: &VoteStep, block_hash: H256) -> bool {
+        let votes = self
+            .votes
+            .get(round)
+            .map(|c| c.block_votes.keys().cloned().filter_map(|x| x).collect())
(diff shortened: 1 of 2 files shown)


core/src/consensus/tendermint/worker.rs
```diff
@@ -683,9 +683,9 @@ impl Worker {
                 } else {
                     let parent_block_hash = self.prev_block_hash();
                     if self.is_signer_proposer(&parent_block_hash) {
-                        if let TwoThirdsMajority::Lock(lock_view, _) = self.last_two_thirds_majority {
+                        if let TwoThirdsMajority::Lock(lock_view, locked_block_hash) = &self.last_two_thirds_majority {
```
foriequal0: `TwoThirdsMajority` derives `Copy`. Would you check again it requires to take a reference?

core/src/consensus/tendermint/worker.rs
```diff
@@ -796,14 +796,15 @@ impl Worker {
         }
     }
 
-    fn locked_proposal_block(&self, locked_view: View) -> Result<encoded::Block, String> {
+    fn locked_proposal_block(&self, locked_view: View, locked_proposal_hash: H256) -> Result<encoded::Block, String> {
         let vote_step = VoteStep::new(self.height, locked_view, Step::Propose);
-        let locked_proposal_hash = self.votes.get_block_hashes(&vote_step).first().cloned();
+        let received_locked_block =
+            self.votes.get_block_hashes(&vote_step).into_iter().any(|block_hash| block_hash == locked_proposal_hash);
```
foriequal0: The pattern is repeating. Make a function `has_votes_for(vote_step, hash)`

core/src/consensus/tendermint/worker.rs
```diff
@@ -971,8 +972,9 @@ impl Worker {
 
         let current_height = self.height;
         let vote_step = VoteStep::new(self.height, self.view, self.step.to_step());
-        let proposal_at_current_view = self.votes.get_block_hashes(&vote_step).first().cloned();
-        if proposal_at_current_view == Some(proposal.hash()) {
+        let is_current_proposal =
+            self.votes.get_block_hashes(&vote_step).into_iter().any(|block_hash| block_hash == proposal.hash());
+        if is_current_proposal {
```
foriequal0: Putting a verb prefix in a local boolean variable is awkward.
Microsoft's conventions "suggests" that boolean "object property" can have 'is', 'has', 'can' prefix[1]. However, the convention is not blindly applicable to another context.
Local boolean variables don't have an "object" to provide contexts. Also when the method/field is not only about an object that contains it, verb prefix becomes awkward [2]
So, `current_proposal_exists` would be more readable name.

[1] https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-properties
[2] https://softwareengineering.stackexchange.com/a/102738

core/src/consensus/tendermint/worker.rs
```diff
@@ -1005,16 +1007,16 @@ impl Worker {
             self.move_to_height(height);
             let prev_block_view = TendermintSealView::new(proposal.seal()).previous_block_view().unwrap();
             self.save_last_confirmed_view(prev_block_view);
-            let proposal_at_view_0 = self
+            let is_view0_proposal = self
```
foriequal0:
```suggestion
            let view0_proposal_exists = self
```

core/src/consensus/tendermint/worker.rs
```diff
@@ -454,7 +454,8 @@ impl Worker {
         self.validators.next_block_proposer(prev_block_hash, view)
     }
 
-    pub fn proposal_at(&self, height: Height, view: View) -> Option<(SchnorrSignature, usize, Bytes)> {
+    /// If there are multiple proposals, return the first one.
```
foriequal0: Should we log it?

core/src/consensus/tendermint/worker.rs
majecty: I want to check that the `proposal` variable is for the current step. `current_proposal_exists` seems there is another proposal in the current step. How about `current_proposal` as the variable name.

core/src/consensus/tendermint/worker.rs
majecty: Same as the upper comment. `view0_proposal_exists` seems that there is another proposal for view0 other than the `proposal` parameter. How about using `view0_proposal` as the name.

core/src/consensus/tendermint/worker.rs
majecty: Okay let's remove it.

core/src/consensus/tendermint/worker.rs
```diff
@@ -454,7 +454,7 @@ impl Worker {
         self.validators.next_block_proposer(prev_block_hash, view)
     }
 
-    pub fn proposal_at(&self, height: Height, view: View) -> Option<(SchnorrSignature, usize, Bytes)> {
+    pub fn first_proposal_at(&self, height: Height, view: View) -> Option<(SchnorrSignature, usize, Bytes)> {
```
foriequal0: `VoteCollector` and `StepCollector` don't guarantee the ordering of underlying votes. You should change the prefix to `any_` or fix the vote collector

core/src/consensus/tendermint/vote_collector.rs
```diff
@@ -64,7 +64,8 @@ impl StepCollector {
     /// Returns Some(&Address) when validator is double voting.
     fn insert(&mut self, message: ConsensusMessage) -> Option<DoubleVote> {
         // Do nothing when message was seen.
-        if self.messages.insert(message.clone()) {
+        if self.messages.contains(&message) {
```
foriequal0:
```suggestion
        if !self.messages.contains(&message) {
            self.messages.push(message.clone());
        } else {
```

core/src/consensus/tendermint/worker.rs
majecty: I made the vote_collecter returns messages in the insertion order.

PR #1795, opened 2019-09-25
Do not panic when the height, view or step changes while creating a block

majecty:


core/src/consensus/tendermint/worker.rs
```diff
@@ -1103,6 +1103,23 @@ impl Worker {
     }
 
     fn proposal_generated(&mut self, sealed_block: &SealedBlock) {
+        let proposal_height = sealed_block.header().number();
+        let proposal_seal = sealed_block.header().seal();
+        let proposal_view = TendermintSealView::new(proposal_seal)
+            .consensus_view()
+            .expect("Generated proposal should have a valid seal");
+        if proposal_height < self.height || (proposal_height == self.height && proposal_view != self.view) {
```
foriequal0: What if `proposal_height > self.height`?

majecty: I added an assertion!

core/src/consensus/tendermint/worker.rs
```diff
@@ -1103,6 +1103,24 @@ impl Worker {
     }
 
     fn proposal_generated(&mut self, sealed_block: &SealedBlock) {
+        let proposal_height = sealed_block.header().number();
+        let proposal_seal = sealed_block.header().seal();
+        let proposal_view = TendermintSealView::new(proposal_seal)
+            .consensus_view()
+            .expect("Generated proposal should have a valid seal");
+        assert!(proposal_height <= self.height, "A generated proposal cannot be generated on the future height");
```
foriequal0: Would you touch the sentence? 'generated' is repeated.
```suggestion
        assert!(proposal_height <= self.height, "A proposal cannot be generated on the future height");
```

PR #1803, opened 2019-10-04
Extract the crypto module to a new repository

sgkim126:
The module is moved to https://github.com/CodeChain-io/rust-codechain-crypto.

This PR closes https://github.com/CodeChain-io/codechain/issues/1500.


majecty:
@sgkim126 
Should we use the git URL to find the crypto library?
What do you think about uploading the Crypto package in the Cargo registry?

sgkim126:
@majecty I think registering in crates will merely increase the management cost with little benefit. Others except for CodeChain will rarely use the library since we customized it for CodeChain.

kseo:
I agree with @sgkim126.

PR #1806, opened 2019-10-07
Use rust 1.37.0

sgkim126:


3
```diff
@@ -0,0 +1,553 @@
+// Copyright 2019 Kodebox, Inc.
```
majecty: Why is this file added?

sgkim126: Mistake. I updated the PR.

sgkim126:
@kseo Please review it.

Issue #1805, opened 2019-10-04
Stop generating a block if view is changed while creating the block.


majecty:
The `generate_seal` function assumes that the generating block is generated for the current view. If the block is generated in the previous view, Tendermint crashes by an assertion in worker.rs#L1047 line.

https://github.com/CodeChain-io/codechain/blob/c24b6f76951965e5cee6b893e18d096034f6f832/core/src/consensus/tendermint/worker.rs#L1047

PR #1807, opened 2019-10-07
Do not generate a seal if the block is generated from a past view

majecty:
If generating a block takes too much time, the view could be changed
before the miner module requests the signature. If the Tendermint
module receives a signature request of an old view, it should ignore
the message. Before this commit, the Tendermint module was
crashing when it gets a signature request from an old view.

It fixes https://github.com/CodeChain-io/codechain/issues/1805.

### DIFF
--- a/core/src/consensus/tendermint/worker.rs
+++ b/core/src/consensus/tendermint/worker.rs
@@ -1044,7 +1044,13 @@ impl Worker {
             return Seal::None
         }
 
-        assert!(self.is_signer_proposer(&parent_hash));
+        // We don't know at which view the node starts generating a block.
+        // If this node's signer is not proposer at the current view, return none.
+        if !self.is_signer_proposer(&parent_hash) {
+            cwarn!(ENGINE, "Seal request for an old view");
+            return Seal::None
+        }
+
         assert_eq!(Proposal::None, self.proposal);
         assert_eq!(height, self.height);
 


core/src/consensus/tendermint/worker.rs
```diff
@@ -1044,7 +1044,13 @@ impl Worker {
             return Seal::None
         }
 
-        assert!(self.is_signer_proposer(&parent_hash));
+        // We don't know at which view the node starts generating a block.
+        // If this node's signer is not proposer at the current view, return none.
+        if self.is_signer_proposer(&parent_hash) {
```
foriequal0:
```suggestion
        if !self.is_signer_proposer(&parent_hash) {
```

core/src/consensus/tendermint/worker.rs
```diff
@@ -1044,7 +1044,13 @@ impl Worker {
...
+            cwarn!(ENGINE, "Proposer is generated after view change");
```
foriequal0:
```suggestion
            cwarn!(ENGINE, "View is changed after requesting the seal");
```

majecty: View is not changed after the seal request. View is changed after the block generation request.
What do you think about "Seal request for an old view"?

foriequal0: That's good.

PR #1811, opened 2019-10-10
Fix the condition to check whether the new block could change the best block

majecty:
When a new block is imported in Tendermint consensus, CodeChain checks the conditions below to judge whether it is safe to change the best block: 

First, whether the score is higher than before. 
Second, whether the height of a new block is higher than the finalized block's height.
This commit fixes the code that is calculating the finalized block's height erroneously.

### DIFF
--- a/core/src/consensus/mod.rs
+++ b/core/src/consensus/mod.rs
@@ -266,10 +266,15 @@ pub trait ConsensusEngine: Sync + Send {
         header.hash()
     }
 
-    /// In PoW consensus, the higher scored block became the best block.
+    /// In PoW consensus, the higher scored block becomes the best block.
     /// In Tendermint consensus, the highest scored block may not be the best block.
-    /// Only the child of the current best block could be the next best block in Tendermint consensus.
-    fn can_change_canon_chain(&self, _header: &HeaderView) -> bool {
+    /// Only the descendant of the current best block could be the next best block in Tendermint consensus.
+    fn can_change_canon_chain(
+        &self,
+        _new_header: &HeaderView,
+        _previous_best_hash: H256,
+        _previous_best_proposal_hash: H256,
+    ) -> bool {
         true
     }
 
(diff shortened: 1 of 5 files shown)


core/src/consensus/tendermint/engine.rs
```diff
@@ -308,15 +308,9 @@ impl ConsensusEngine for Tendermint {
...
-        self.inner
-            .send(worker::Event::AllowedHeight {
-                result,
-            })
-            .unwrap();
-        let allowed_height = receiver.recv().unwrap();
-        header.number() >= allowed_height
+
+    fn can_change_canon_chain(&self, new_header: &HeaderView, prev_best_hash: H256) -> bool {
+        new_header.parent_hash() == prev_best_hash
```
majecty: In Tendermint consensus, only the child of the parent block could be the best block.

core/src/consensus/mod.rs
```diff
@@ -266,7 +266,10 @@ pub trait ConsensusEngine: Sync + Send {
         header.hash()
     }
 
-    fn can_change_canon_chain(&self, _header: &HeaderView) -> bool {
+    /// In PoW consensus, the higher scored block became the best block.
```
ScarletBlue: became -> becomes?

core/src/consensus/tendermint/worker.rs
```diff
@@ -307,16 +304,6 @@ impl Worker {
                             }) => {
                                 inner.set_signer(ap, address);
                             }
-                            Ok(Event::AllowedHeight {
-                                result,
-                            }) => {
-                                let allowed_height = if inner.step.is_commit() {
-                                    inner.height + 1
```
majecty: This was the cause of a bug.
If a proposal is not imported in Commit state yet, the code should return `inner.height` instead of `inner.height + 1`.

core/src/consensus/mod.rs
majecty: You're right. I'll change it to `becomes`.

PR #1813, opened 2019-10-14
Enable all grandchild of the best block to become a best proposal block

majecty:
Before this commit, only the child of the best block and the child of the best proposal block could be a new best proposal block. A block that is a grandchild of the best block but not the child of the best proposal block should be able to become a best proposal block.

After this commit, CodeChain checks whether a new block is the best block's child or grandchild.


foriequal0:
It seems to be very ad-hoc. Is there any chance to receive grand-grand children?

majecty:
@foriequal0 
No. only the child of the best block and grandchild of the best block could be received.
If a valid block is imported, its parent block became a new best block.

Here is an example.
1. Block A is the best block.
2. Block B is a proposal block and B is a child of A.
3. A block C, which is a child of B, is imported(At that time A is the best block).
4. Since C has precommits of B, B became the new best block.

PR #1827, opened 2019-10-25
Introduce minimum fees configured by machines

### LINKED ISSUE #1758: Configure minimum fee in a node
If a validator mines a transaction with network's minimum fee, the validator couldn't get any reward from the transaction. The validator should be able to set minimum fees for their node for profit.

HoOngEe:
It resolves #1758 
It enables validators to customize fee thresholds for transactions to be included in their mem pools.


majecty:
Let's add an integration test with the scenario below:
1. Node A's minimum fee for a pay transaction is 100CCC
2. Node B's minimum fee for a pay transaction is 200CCC
3. Node A creates a Block with a transaction whose fee is 140CCC.
4. Node B verifies the block successfully.

HoOngEe:
I implemented the integration test you commented.

HoOngEe:
> You've added codechain binary here: remove it from the repository. [811ab50](https://github.com/CodeChain-io/codechain/commit/811ab5043c7124fbfe6a84aa576279a495dd97d0)

Thank you, I removed it from the commit.

core/src/miner/mem_pool.rs
```diff
@@ -781,17 +770,18 @@ impl MemPool {
         origin: TxOrigin,
         client_account: &AccountDetails,
     ) -> Result<(), Error> {
-        if origin != TxOrigin::Local && tx.fee < self.minimal_fee {
+        let action_min_fee = self.minimum_fees.min_cost(&tx.action);
```
foriequal0: Why `origin != TxOrigin::Local` condition is dropped?

HoOngEe: Because I think that's the correct behavior. Some services let users send their transactions via RPC directly to the nodes. In this case, I think the mem pool minimum fees should be applied to such transactions.

foriequal0: Oh. The spec is changed. okay.

test/src/e2e.long/mempoolMinfee.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        await node1_with_min_pay_fee_150.connect(node2_with_min_pay_fee_200);
+        await node1_with_min_pay_fee_150.connect(node3);
+        await node2_with_min_pay_fee_200.connect(node3);
+
+        await node1_with_min_pay_fee_150.waitPeers(2);
+        await node2_with_min_pay_fee_200.waitPeers(2);
+        await node3.waitPeers(2);
+    });
+
+    it("two nodes with different minimum pay costs1", async function() {
```
foriequal0: There is actually three nodes. The title is not descriptive. You should change the title to "should be a node that accepts a tx whose fee is greater than the node's min fee criteria, and the tx should be propagated through the network" or some similar and concise one.

test/src/e2e.long/mempoolMinfee.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        expect(
+            await node2_with_min_pay_fee_200.sdk.rpc.chain.containsTransaction(
+                tx.hash()
+            )
+        ).to.be.true;
+        expect(await node3.sdk.rpc.chain.containsTransaction(tx.hash())).to.be
+            .true;
+    });
+
+    it("two nodes with different minimum pay costs2", async function() {
```
foriequal0: There is actually three nodes. The title is not descriptive. You should change the title to "should block the tx whose fee is less than min_fee" or some similar and concise one.

core/src/miner/mem_pool.rs
majecty: I think that we should remain the original behavior. RPC should not be open to the internet. We didn't consider any safety in RPC. All the people who want to send a transaction should have its own CodeChain instance.

majecty: How about discussing this with other team members?

HoOngEe: Ok then, I'll summary it and make a conversation. Until now, the mem pool's minimal fee is always set to zero.

test/src/e2e.long/mempoolMinfee.test.ts
HoOngEe: Thank you, I've changed the code but forgot to change the title.

HoOngEe: Changed the title to a descriptive one.

test/src/e2e.long/mempoolMinfee.test.ts
HoOngEe: Changed the title to a descriptive one.

core/src/miner/mem_pool.rs
HoOngEe: I changed the behavior

core/src/miner/mem_pool.rs
```diff
@@ -1538,6 +1552,83 @@ pub mod test {
...
+            timestamp: None,
+        };
+
+        let inserted_block_number = 1;
+        let inserted_timestamp = 100;
+        let inputs = vec![
+            MemPoolInput::new(create_signed_pay_with_fee(0, 200, keypair), TxOrigin::External, no_timelock),
+            MemPoolInput::new(create_signed_pay_with_fee(1, 140, keypair), TxOrigin::External, no_timelock),
+            MemPoolInput::new(create_signed_pay_with_fee(1, 160, keypair), TxOrigin::External, no_timelock),
+            MemPoolInput::new(create_signed_pay_with_fee(2, 149, keypair), TxOrigin::External, no_timelock),
```
foriequal0: Would you add test cases for `TxOrigin::Local`?

HoOngEe: I added a unit test.

PR #1842, opened 2019-11-04
[ibc] Add CreateClient transaction

majecty:


core/src/ibc/client_02/manager.rs
```diff
@@ -0,0 +1,37 @@
...
+use ibc;
+
+pub struct Manager {}
+
+impl Manager {
+    pub fn new() -> Self {
+        Manager {}
+    }
+
+    pub fn create(&self, ctx: &dyn ibc::Context, id: &str, cs: &dyn ConsensusState) -> Result<Box<dyn State>, String> {
```
majecty: `Manager` is the main point of ICS-02.
I'll add `ChannelManager` and `ConnectionManager`.

core/src/ibc/context.rs
```diff
@@ -0,0 +1,60 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+use super::kv_store;
+use cstate::TopLevelState;
+use ibc::KVStore;
+
+pub trait Context {
```
majecty: I want to isolate IBC related code for easy test environment.
IBC code only depends on the `Context` type to create any side effect.

PR #1845, opened 2019-11-05
[ibc] Implement KVStore

majecty:


majecty:
@joojis 
I think that ActionData is a demo data type, and it should be used only for the HitHandler.
Do you think reusing ActionData is better?

joojis:
@majecty ActionData is for the custom transactions as far as I know. Maybe I'm confused.

Adding a new one is ok to me but I was wondering because it seems `IBC` is no different from `ActionData`.

majecty:
Aha, you are right. I found that the stake module is using the action_data. I'll change it to use the action_data.

PR #1849, opened 2019-11-07
[ICS] Implement IBC client RPCs

majecty:


majecty:
We implemented the ICS PoC version in Foundry.

PR #1855, opened 2019-11-11
Add state to sync extension

Remagpie:


util/merkle/src/triedb.rs
```diff
@@ -105,6 +105,27 @@ impl<'db> TrieDB<'db> {
             None => Ok(None),
         }
     }
+
+    /// Check if every leaf of the trie exists
+    pub fn is_complete(&self) -> bool {
+        self.is_complete_aux(self.root)
```
foriequal0: You should check that `self.root == BLAKE_RLP_NULL`.

PR #1864, opened 2019-11-15
Rename total_score to nonce in sync message

Remagpie:


Remagpie:
> It's good to use nonce instead of the total score. total_score isn't saved in the state and need the whole history to calculate it. It is hard to manage and verify `total_scorea`.
> 
> This commit removes last_request and uses nonce instead. Sync extension is using last_requset to give a unique id of requests. The sync extension will work even though we use nonce instead. However, one variable used by two different purposes makes maintenance hard. IMO using two variables for each purpose will make code maintainable.

Agreed. Let me separated those variables.

sync/src/block/extension.rs
```diff
@@ -339,13 +341,7 @@ impl NetworkExtension<Event> for Extension {
...
 
                         for id in peer_ids {
-                            let peer_score = if let Some(peer) = self.header_downloaders.get(&id) {
-                                peer.total_score()
-                            } else {
-                                U256::zero()
-                            };
-
-                            if peer_score > best_proposal_score {
+                            if self.header_downloaders.get(&id).map_or(false, |d| !d.is_lagging()) {
```
majecty: `is_lagging` is not correct to use as a condition for `send_body_request`. When a node's peer has the same best block with the node, the node should not send a body request.

```rust
let is_lagging = {
  let peer_status = self.client.block_status(&BlockId::Hash(self.best_hash));
  let best_hash = self.client.block_hash(&BlockId::Latest).expect("Best hash always exists");
  peer_status == BlockStatus::InChain && self.best_hash != best_hash
}
```

correct condition
```rust
let is_lagging = {
  let peer_status = self.client.block_status(&BlockId::Hash(self.best_hash));
  let best_hash = self.client.block_hash(&BlockId::Latest).expect("Best hash always exists");
  peer_status == BlockStatus::InChain
}
```

sync/src/block/downloader/header.rs
```diff
@@ -108,25 +89,28 @@ impl HeaderDownloader {
...
     pub fn is_idle(&self) -> bool {
-        let can_request = self.request_time.is_none() && self.total_score > self.pivot.total_score;
+        let can_request = self.request_time.is_none() && self.best_hash != self.pivot;
 
         self.is_valid() && (can_request || self.is_expired())
     }
 
+    pub fn is_lagging(&self) -> bool {
+        let peer_status = self.client.block_status(&BlockId::Hash(self.best_hash));
+        let best_hash = self.client.block_hash(&BlockId::Latest).expect("Best hash always exists");
```
majecty: Is `block_hash` the header chain's best hash?

sync/src/block/extension.rs
```diff
@@ -759,14 +756,12 @@ impl Extension {
...
             State::SnapshotChunk(..) => {}
             State::Full => {
-                let (mut completed, pivot_score_changed) = if let Some(peer) = self.header_downloaders.get_mut(from) {
-                    let before_pivot_score = peer.pivot_score();
+                let (mut completed, is_lagging) = if let Some(peer) = self.header_downloaders.get_mut(from) {
                     let encoded: Vec<_> = headers.iter().map(|h| EncodedHeader::new(h.rlp_bytes().to_vec())).collect();
                     peer.import_headers(&encoded);
-                    let after_pivot_score = peer.pivot_score();
-                    (peer.downloaded(), before_pivot_score != after_pivot_score)
+                    (peer.downloaded(), peer.is_lagging())
```
majecty: What do you think about a variable like `pivot_hash_changed`. Then the behavior is the same as before.

Remagpie: I updated this part to use `is_caught` method, which compares the pivot with the best hash.

sync/src/block/downloader/header.rs
Remagpie: It's block chain's best hash.

sync/src/block/extension.rs
Remagpie: Thx. I fixed it.

sync/src/block/extension.rs
```diff
@@ -792,7 +793,7 @@ impl Extension {
                     peer.mark_as_imported(exists);
                     peer.create_request()
                 });
-                if pivot_score_changed {
+                if peer_is_caught_up {
```
majecty:
```suggestion
                if !peer_is_caught_up {
```

Remagpie: OMG! Thx for noticing this!

PR #1865, opened 2019-11-15
Import only desired bootstrap headers

Remagpie:


sync/src/block/extension.rs
```diff
@@ -743,17 +743,21 @@ impl Extension {
     fn on_header_response(&mut self, from: &NodeId, headers: &[Header]) {
         ctrace!(SYNC, "Received header response from({}) with length({})", from, headers.len());
         match self.state {
-            State::SnapshotHeader(..) => {
+            State::SnapshotHeader(hash, _) => {
                 for header in headers {
```
foriequal0: Should we transform this into like:
```rust
match headers {
	[header] if header.hash() == hash => {
	  ... import ...
	}
	_ => {
	  cwarn!(SYNC, "Unexpected response")
	}
}
```

sync/src/block/extension.rs
```diff
@@ -743,20 +743,22 @@ impl Extension {
...
                         // FIXME: handle import errors
                         Err(err) => {
                             cwarn!(SYNC, "Cannot import header({}): {:?}", header.hash(), err);
-                            break
                         }
                         _ => {}
                     }
                 }
-            }
+                _ => cdebug!(SYNC, "Peer {} didn't have the snapshot header {}", from, hash),
```
foriequal0: The log message is misleading. inversion of `[header] if header.hash() == hash` is "headers is not `[header]`" (such as `[]`, `[h1, ...]`), or `header.hash() != hash`. You should change it to `"Peer {} responded with the invalid response. requested hash: {}, response  length: {}"`

Remagpie: I thought it is checked in `is_valid_response`, but it seems it's not.
I'll fix this.

PR #1866, opened 2019-11-15
Set finalized_view to None when restoring in the Commit step

majecty:
In the Tendermint restoring process, if the backup state is the Commit
step, CodeChain set the step to the Precommit step and handles the
votes. The purpose of the behavior is that calling functions that are
called when enters the Commit state.

After we introduce the `finalized_view_of_current_block` variable, the
variable should be changed to `None` when CodeChain sets the state to
Precommit in the restoring process.


foriequal0:
Oh we should've made rc-2.2.x branch first according to the Release manual.

PR #1871, opened 2019-11-18
Do not assert the finalized_view while restoring

majecty:
In the Tendermint restoring process, if the step variable is backed up by the Commit value, CodeChain sets the step to the Precommit step and handles the votes. The purpose of this behavior is to call functions that are called when the node enters the Commit state.

Changing the "step" in the restoring process is fragile. It is easy to miss setting some variables. The `finalized_view_of_current_block` variable should be changed when the step is changed. Until now, it was forgotten due to human error.

This commit fixes the problem by ignoring the assertion check in the restore process. Although it is not a perfect solution, it is consistent. There is some code already that does different things in the restore process, depending on whether the code is called or not.

### DIFF
--- a/core/src/consensus/tendermint/worker.rs
+++ b/core/src/consensus/tendermint/worker.rs
@@ -612,8 +612,11 @@ impl Worker {
     }
 
     /// Do we need this function?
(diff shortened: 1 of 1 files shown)


foriequal0:
I'll merge this manually.

PR #1872, opened 2019-11-18
Collect snapshot chunks from peers

Remagpie:


sync/src/block/extension.rs
```diff
@@ -193,6 +234,14 @@ impl Extension {
                 })
                 .collect();
 
+            let chunk_requests: Vec<RequestMessage> = requests
+                .iter()
+                .filter_map(|r| match r {
+                    (_, RequestMessage::StateChunk(..)) => Some(r.1.clone()),
```
foriequal0: Binding is more clear.
`(_, msg@RequestMessage::StateChunk(..)) => Some(msg.clone()),`

util/merkle/src/snapshot/mod.rs
```diff
@@ -91,6 +90,12 @@ impl<'a> Restore<'a> {
     }
 }
 
+impl std::fmt::Debug for Restore {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        write!(f, "Restore {{ pending: {:?} }}", self.pending)
```
foriequal0: You can use this: https://doc.rust-lang.org/std/fmt/struct.DebugStruct.html
```rust
        fmt.debug_struct("Restore")
           .field("pending", &self.pending)
           .field("unresolved", "<...>")
           .finish()
```

sync/src/block/extension.rs

util/merkle/src/snapshot/mod.rs
Remagpie: Fixed.

PR #1873, opened 2019-11-19
Add RPC snapshot_getList

Remagpie:
`snapshot_getList` RPC is introduced by this PR.
The RPC checks the `snapshot_dir` directory provided by the config and inspects its entries.
An entry is recognized as a snapshot if it satisfies the following rules:
* Is a directory
* Directory name is a valid 256bit hex string
* There exists a block which has a hash same as the directory name

### DIFF
--- a/codechain/rpc.rs
+++ b/codechain/rpc.rs
@@ -15,8 +15,8 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 use std::io;
-use std::net::SocketAddr;
 
+use crate::config::Config;
 use crate::rpc_apis;
 use crpc::{
     jsonrpc_core, start_http, start_ipc, start_ws, HttpServer, IpcServer, MetaIoHandler, Middleware, WsError, WsServer,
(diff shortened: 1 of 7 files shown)


codechain/rpc_apis.rs
```diff
@@ -35,6 +35,7 @@ impl ApiDependencies {
         use crpc::v1::*;
         handler.extend_with(ChainClient::new(Arc::clone(&self.client)).to_delegate());
         handler.extend_with(MempoolClient::new(Arc::clone(&self.client)).to_delegate());
+        handler.extend_with(SnapshotClient::new(Arc::clone(&self.client)).to_delegate());
```
foriequal0: Why don't we just pass Config to this function? Why take a detour through `Client` to pass a `snapshot_path`?

rpc/src/v1/impls/snapshot.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+impl<C> Snapshot for SnapshotClient<C>
+where
+    C: BlockChainClient + CoreSnapshotClient + 'static,
+{
+    fn get_snapshot_list(&self) -> Result<Vec<BlockNumberAndHash>> {
+        let ids = {
+            if let Some(snapshot_dir) = self.client.snapshot_dir() {
```
foriequal0: Change this to early exit pattern.
```rust
let snapshot_dir = if let Some(snapshot_dir) = self.client.snapshot_dir() { 
  snapshot_dir 
} else { 
  return Ok(Vec::new()) 
};
```

codechain/rpc_apis.rs
Remagpie: We can't pass the full config object since it introduces a circular dependency.
We can pass String instead, but I can't see any explicit benefit here.
Do you have any reason for changing it?

rpc/src/v1/impls/snapshot.rs
Remagpie: I think an early return is not a good choice here, but I think I can improve this. Let me fix it.

codechain/rpc_apis.rs
foriequal0: What do you mean by a circular dependency? An `Arc` circular dependency or a module dependency?

Remagpie: A module dependency.
`SnapshotClient` is in `rpc` crate and `Config` is in `codechain` crate.
Both crates need to depend on each other if we have to pass `Config` to `SnapshotClient`.

foriequal0: Okay. But I still don't think punching through the client is a good idea. We put too much information into the client. We might try to separate the config module?

codechain/rpc.rs
```diff
@@ -32,39 +32,27 @@ pub struct RpcHttpConfig {
     pub hosts: Option<Vec<String>>,
 }
 
-pub fn rpc_http_start(
-    cfg: RpcHttpConfig,
-    enable_devel_api: bool,
-    deps: &rpc_apis::ApiDependencies,
-) -> Result<HttpServer, String> {
-    let url = format!("{}:{}", cfg.interface, cfg.port);
+pub fn rpc_http_start(config: &Config, deps: &rpc_apis::ApiDependencies) -> Result<HttpServer, String> {
```
sgkim126: I have a question about this refactoring.
Why do you think using `Config` is better than using the specific configuration trait?

Remagpie: `ApiDependencies::extend_api` receives `Config` as an argument, so we need to pass it anyways.
Speciailized configs (e.g. `RpcHttpConfig`) can be derived from `Config`, so I thought it'd be better to reduce duplication.

foriequal0: I think it would be better to refactor the code to be:
```rust
let rpc_server = {
  if !config.rpc.disable.unwrap() {
    let rpc_server = setup_rpc_server(config, deps);
    Some(rpc_http_start(rpc_server, config.rpc_http_config()))
  } else {
    None
  }
};
```

Remagpie: It becomes a bit verbose but I like the idea here. @sgkim126 What do you think about it?

sgkim126: I like @foriequal0's idea too.

Remagpie: Applied the suggestion.

PR #1879, opened 2019-11-21
Download the full snapshot target block

Remagpie:
Original snapshot sync implementation downloaded only the header of the snapshot target.
It introduced lots of problems, so I and @foriequal0 decided to download the body of the snapshot target too.

`SnapshotBody` state is added to the `State` enum in sync extension.
Also, `import_bootstrap_header` in `Client` is changed to `import_trusted_header`, `import_trusted_block` and `force_update_best_block`.

### DIFF
--- a/core/src/client/importer.rs
+++ b/core/src/client/importer.rs
@@ -371,16 +371,33 @@ impl Importer {
         imported.len()
     }
 
-    pub fn import_bootstrap_block<'a>(&'a self, block: &'a Block, client: &Client, _importer_lock: &MutexGuard<()>) {
+    pub fn import_trusted_header<'a>(&'a self, header: &'a Header, client: &Client, _importer_lock: &MutexGuard<()>) {
+        let hash = header.hash();
+        ctrace!(CLIENT, "Importing trusted header #{}-{:?}", header.number(), hash);
+
+        {
+            let chain = client.block_chain();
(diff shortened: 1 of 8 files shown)


sync/src/block/extension.rs
```diff
@@ -858,20 +810,20 @@ impl Extension {
         ctrace!(SYNC, "Received header response from({}) with length({})", from, headers.len());
         match self.state {
             State::SnapshotHeader(hash, _) => match headers {
-                [header] if header.hash() == hash => {
+                [parent, header] if header.hash() == hash => {
```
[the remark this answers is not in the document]
Remagpie: It's checked in `is_valid_response`.

core/src/client/importer.rs
```diff
@@ -371,26 +371,52 @@ impl Importer {
         imported.len()
     }
 
-    pub fn import_bootstrap_block<'a>(&'a self, block: &'a Block, client: &Client, _importer_lock: &MutexGuard<()>) {
+    pub fn import_trusted_header<'a>(&'a self, header: &'a Header, client: &Client, _importer_lock: &MutexGuard<()>) {
```
foriequal0: Why it takes unused `_importer_lock`?

Remagpie: I don't know why. I just used the same interface as `import_headers`.

PR #1881, opened 2019-11-21
Use plain_hasher in crates.io

sgkim126:


sgkim126:
https://github.com/sgkim126/codechain/commit/41fc75ca55d57d619285cc450444bb9c9d8cc5dd
I already checked there are no differences between the upstream and our version.

PR #1883, opened 2019-11-22
Detect bootstrap header in tendermint worker

Remagpie:
Tendermint worker skips some process for the genesis block since it doesn't have a parent, and this should be applied to the bootstrap header too.


core/src/consensus/tendermint/worker.rs
```diff
@@ -958,7 +959,7 @@ impl Worker {
     }
 
     pub fn on_imported_proposal(&mut self, proposal: &Header) {
-        if proposal.number() < 1 {
+        if self.client().block_status(&BlockId::Hash(*proposal.parent_hash())) == BlockStatus::Unknown {
```
majecty: What do you think about checking whether the header is a snapshot header or not? The existence of a header's parent does not mean the header is a snapshot header.

Remagpie: I added some comments explaining this near the updated code.

PR #1870, opened 2019-11-17
Upgrade rlp library to 0.4.0

sgkim126:
This version removed the dependency on elastic_array.

PR #1885, opened 2019-11-23
Use the upstream kvdb

sgkim126:
There are two major changes.
- kvdb-rocksdb removed WAL support
- kvdb::Error is replaced with io::Error


Remagpie:
[kvdb 0.1.1](https://crates.io/crates/kvdb) still uses `elastic-array` crate.
I'm a bit concerned about it considering that we merged https://github.com/CodeChain-io/codechain/pull/1870.

sgkim126:
@Remagpie It's not a concern, because kvdb uses ElasticArray internally and doesn't expose it.

Remagpie:
[`DBValue`](https://docs.rs/kvdb/0.1.1/kvdb/type.DBValue.html) is exposed, and is an alias of `ElasticArray128<u8>`.
I think we will be using this type a lot.

sgkim126:
@Remagpie I missed it.
However, I think it's better to keep track of the upstream.

rpc/src/v1/impls/devel.rs
```diff
@@ -79,7 +79,7 @@ where
     }
 
     fn get_state_trie_value(&self, key: H256) -> Result<Vec<Bytes>> {
-        match self.db.get(COL_STATE, &key).map_err(|e| errors::kvdb(&e))? {
+        match self.db.get(COL_STATE, &key).map_err(errors::transaction_core)? {
```
Remagpie: Is this correct?

core/src/client/error.rs
```diff
@@ -16,14 +16,12 @@
 
 use std::fmt::{Display, Error as FmtError, Formatter};
 
-use kvdb;
 use util_error::UtilError;
 
 /// Client configuration errors.
 #[derive(Debug)]
 pub enum Error {
     /// Database error
```
Remagpie: Please remove this comment too.

rpc/src/v1/impls/devel.rs
sgkim126: Yes, because there is no more kvdb error.
The error becomes io::Error.

Remagpie: I think it'll be interpreted as `unknown_error` in `transaction_core`, which has a message `"Unknown error when sending transaction"`.
I think using `errors::core` is better here.

sgkim126: Looks good. I changed.

Remagpie:
@sgkim126 Is it still work in progress?

sgkim126:
@Remagpie It's done. I'll merge it when Travis passes.

PR #1896, opened 2019-11-26
Refactor `on_close_block`, take #2

Remagpie:


core/src/consensus/tendermint/engine.rs
```diff
@@ -273,7 +267,7 @@ impl ConsensusEngine for Tendermint {
 
         let inner = self.inner.clone();
         let extension = service.register_extension(move |api| TendermintExtension::new(inner, timeouts, api));
-        let client = Weak::clone(self.client.read().as_ref().unwrap());
+        let client = Arc::downgrade(&self.client().unwrap());
```
foriequal0: Note that this patch will changes the behavior (by small change)
The previous one panics only when the `self.client == None`, however, the changed one will panics 1. `self.client == None`, 2. `Weak::upgrade(self.client.unwrap()).is_none()`

core/src/consensus/tendermint/engine.rs
```diff
@@ -137,12 +137,16 @@ impl ConsensusEngine for Tendermint {
     fn on_close_block(
         &self,
         block: &mut ExecutedBlock,
-        parent_header: &Header,
-        parent_common_params: &CommonParams,
         term_common_params: Option<&CommonParams>,
```
foriequal0: Can you remove this parameter too?

Remagpie: I tried to do that, and I thought it'd be a bit complex for now.
How about removing this after https://github.com/CodeChain-io/codechain/pull/1895 is merged?

core/src/consensus/tendermint/engine.rs
Remagpie: Well, it does change the behavior, but I don't think it matters.
`register_network_extension_to_service` is called only once at the initialization, and I think there's no possibility of panic in that situation.

PR #1900, opened 2019-11-28
Precalculate the fee at on_term_open

Remagpie:
Depends on https://github.com/CodeChain-io/codechain/pull/1895

This PR introduces `on_term_open`, and calculates the final fee distribution there.
The detailed design is as follows:
### 1. State structure
State leaf at `makeKey("IntermediateRewards")` contains `[previous, current]` value, and they mean the previous and current intermediate reward that the validators collected.

In era=1, it now contains `[current, calculated]`, and they mean the _current intermediate reward_, and the _calculated result of (reward - penalty)_.

### 2. `on_term_open`
`on_term_open` is called at the first block of the term, and before any of the transactions in the block is processed.
The term start is detected by the following condition:
`block_number == metadata.last_term_finished_block_num() + 1`

When the era is zero, it does nothing.
When the era is 1, it calculates the penalty that each validator should take, and substracts it from the _current_ reward in the state, and saves the result in _calculated_ part.
After that, it clears the _current_ reward.

### 3. `on_term_close`
When the era is 1, the _calculated_ fee is read from the state and CCC is distributed according to that value.
Note that although it is not strictly required to do this, the _calculated_ part is cleared to empty map here.

### DIFF
--- a/core/src/consensus/stake/action_data.rs
+++ b/core/src/consensus/stake/action_data.rs
@@ -18,7 +18,6 @@ use std::cmp::Ordering;
 use std::collections::btree_map::{BTreeMap, Entry};
 use std::collections::btree_set::{self, BTreeSet};
 use std::collections::{btree_map, HashMap, HashSet};
-use std::mem;
 use std::ops::Deref;
 use std::vec;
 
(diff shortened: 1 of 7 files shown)


core/src/consensus/stake/mod.rs
```diff
@@ -322,25 +325,57 @@ pub fn get_validators(state: &TopLevelState) -> StateResult<Validators> {
...
+            let mut rewards = IntermediateRewardsV1::load_from_state(state)?;
+            rewards.add_quantity(address, reward);
+            rewards.save_to_state(state)?;
+        }
+        _ => unimplemented!(),
+    }
     Ok(())
 }
 
 pub fn drain_previous_rewards(state: &mut TopLevelState) -> StateResult<BTreeMap<Address, u64>> {
```
foriequal0: I think it is more clear to the outside if it is namespaced.
There would be some approaches.
We can make them private and publicly reexport in the versioned mod.
```rust
fn add_intermediate_rewards;
fn drain_previous_rewards;
fn drain_current_rewards;

pub mod v0 {
	pub use super::add_intermediate_rewards;
	pub use super::drain_previous_rewards;
	...
}

pub mod v1 {
	pub use super::add_intermediate_rewards;
	pub use super::drain_current_rewards
	...
}
```
The other approach can be C++ style inline namespace style.
https://foonathan.net/2018/11/inline-namespaces/

```rust
pub mod v0 {
	pub fn add_intermediate_rewards;
	pub fn drain_previous_rewards;
	...
}

pub mod v1 {
	pub fn add_intermediate_rewards;
	pub fn drain_current_rewards
	...
}

pub fn add_intermediate_rewards(...) {
	if era == 0 {
		v0::add_intermediate_rewards(...);
	} else {
		v1::add_intermediate_rewards(...);
	}
}
pub use v1::{drain_current_rewards, ...};
```

core/src/consensus/stake/action_data.rs
```diff
@@ -1129,46 +1182,85 @@ mod tests {
...
+        assert_eq!(origin_rewards, recovered_rewards);
+        recovered_rewards.drain_current();
+        recovered_rewards.save_to_state(&mut state).unwrap();
+        let mut final_rewards = IntermediateRewardsV1::load_from_state(&state).unwrap();
+        assert_eq!(BTreeMap::new(), final_rewards.current);
+        final_rewards.drain_calculated();
+        assert_eq!(BTreeMap::new(), final_rewards.calculated);
+    }
+
     #[test]
```
foriequal0: Can you add a test save in V0 -> load V1?

core/src/consensus/stake/action_data.rs
```diff
@@ -456,6 +456,59 @@ impl IntermediateRewards {
     }
 }
 
+#[derive(Default, Debug, PartialEq)]
+pub struct IntermediateRewardsV1 {
```
foriequal0: (just a comment) I understand that the operational semantics are changed without changing the format. However, it triggers me somehow.

core/src/consensus/stake/mod.rs
Remagpie: Fixed.

core/src/consensus/stake/action_data.rs
Remagpie: Fixed. It required some `pub(super)` modifiers, so the code became a bit ugly :(

Remagpie:
I tested it in the local network and found out that there is some problem with fee distribution.
The distributed amount should be the same with before, but fee-monitor says there are some difference.
I'll try to figure out what's going on.

core/src/consensus/tendermint/engine.rs
```diff
@@ -134,6 +134,49 @@ impl ConsensusEngine for Tendermint {
...
+                    let start_of_the_previous_term = {
+                        let end_of_the_two_level_previous_term = client
+                            .last_term_finished_block_num((metadata.last_term_finished_block_num() - 1).into())
+                            .unwrap();
+
+                        end_of_the_two_level_previous_term + 1
+                    };
+
+                    let banned = stake::Banned::load_from_state(block.state())?;
+                    let start_of_the_current_term_header =
```
majecty: Does the header contains seal here?

Remagpie: I checked the related parts, and I'm pretty sure that the seal exists at this point.

Remagpie:
Finished testing it on the local network, and the fee distribution is correctly working as I expected.
Please review the changed parts again.

PR #1904, opened 2019-11-30
Some miscellaneous changes of e2e tests

sgkim126:


test/src/e2e.dynval/1/dv.changeParams.test.ts
```diff
@@ -313,7 +314,7 @@ describe("Change commonParams that doesn't affects validator set", function() {
                 await checkingNode.sdk.rpc.chain.sendSignedTransaction(
                     largeNomination.sign({
                         secret: alice.privateKey,
-                        seq,
+                        seq: seq + 1,
```
sgkim126: I'll give you some more explanation for it.
This test sends an invalid transaction and checks whether the RPC returns an error.
Because the validation step is before checking seq, the seq doesn't affect the behavior in practical.
However, I think we need to set up seq properly so that the test behaves like a real execution.

PR #1905, opened 2019-12-01
Implement Drop for SnapshotService

sgkim126:
Currently, the main thread will exit while the snapshot service is creating snapshots.

### DIFF
--- a/sync/src/snapshot/service.rs
+++ b/sync/src/snapshot/service.rs
@@ -14,13 +14,16 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+use std::collections::HashMap;
 use std::io::ErrorKind;
 use std::path::PathBuf;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
-use std::thread::spawn;
+use std::thread::{spawn, JoinHandle};
 
 use ccore::{BlockChainClient, BlockChainTrait, BlockId, ChainNotify, Client, DatabaseClient};
 use ctypes::BlockHash;
+use parking_lot::Mutex;
 
 use super::error::Error;
 use super::snapshot::{Snapshot, WriteSnapshot};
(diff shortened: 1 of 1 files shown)


sync/src/snapshot/service.rs
```diff
@@ -66,11 +73,26 @@ impl ChainNotify for Service {
...
+            let joins = Arc::clone(&self.joins);
+            let join = spawn(move || {
+                match Snapshot::try_new(path).map(|s| s.write_snapshot(db.as_ref(), &root)) {
+                    Ok(_) => {}
+                    Err(Error::FileError(ErrorKind::AlreadyExists)) => {}
+                    Err(e) => cerror!(SNAPSHOT, "{}", e),
+                }
+                joins.lock().remove(&id);
             });
+            self.joins.lock().insert(id, join);
```
foriequal0: There might be a race condition where the spawned thread exits first, so `joins.remove()` is called first and `joins.insert()` is called later.

How about cleaning it lazily?
```rust
fn new_blocks() {
	...
	let arc = Arc::new(());
	let weak = Arc::downgrade(&arc);
	let join = spawn(move || {
		let arc = arc; // If the stack dies, the arc dies. 
	    ...
	}
	let mut joins = joins.lock();
	joins.push((weak, join));
	joins.retain(|(weak, _)| Weak::upgrade(weak).is_some());
	...
}
```

Remagpie:
I think we merged a similar implementation in https://github.com/CodeChain-io/codechain/pull/1869 (in `snapshot` branch).
Would you check it first?

sync/src/snapshot/service.rs
Remagpie: I think it can be solved by locking `self.joins` in the upper line.
```rust
let joins = Arc::clone(&self.joins);
let self_join = self.joins.lock();
let join = spawn(move || { ... });
self_join.insert(id, join);
```
The variable name is a bit awkward, but I think you'll get the point.

sgkim126: @foriequal0 @Remagpie How about it?
```rust
let id = self.thread_ids.fetch_add(1, Ordering::SeqCst);
self.joins.lock().insert(id, join);
let joins = Arc::clone(&self.joins);
let join = spawn(move || {
```
I think it's simple but working.

sgkim126: IMO, @Remagpie's looks simple. @foriequal0 how about it?

sgkim126:
> I think we merged a similar implementation in #1869 (in `snapshot` branch).
> Would you check it first?

Ah. Yes, it is fixed in the snapshot branch. I'll close it.

Issue #1839, opened 2019-11-01
Can't specify custom chain file in the config file


Remagpie:
Currently, there are two ways of selecting which chain to use.
1. Through `codechain.chain` field in the config file
1. Through `-c` option in the command line argument

The latter can accept `*.json` file as a custom chain spec, but the former doesn't accept values other than the built-in chains.

PR #1907, opened 2019-12-03
Accept arbitrary string as a chain type

Remagpie:
Fix https://github.com/CodeChain-io/codechain/issues/1839

### DIFF
--- a/codechain/config/chain_type.rs
+++ b/codechain/config/chain_type.rs
@@ -19,17 +19,16 @@ use std::{fmt, fs};
 
 use ccore::Scheme;
 use never_type::Never;
+use serde::de::{Error, Visitor};
+use serde::{Deserialize, Deserializer};
 
-#[derive(Clone, Debug, PartialEq, Deserialize)]
-#[serde(rename_all = "snake_case")]
+#[derive(Clone, Debug, PartialEq)]
 pub enum ChainType {
     Mainnet,
     Solo,
-    #[serde(rename = "simple_poa")]
     SimplePoA,
     Tendermint,
     Cuckoo,
-    #[serde(rename = "blake_pow")]
     BlakePoW,
     Husky,
     Saluki,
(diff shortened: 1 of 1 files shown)


sgkim126:
What's the difference between using a cli flag and a config file?

Remagpie:
There should be no difference, and the cli flag overrides the config file's parameter.

sgkim126:
What I asked is why the current code accepts an arbitrary file name for a cli flag, but not for a config file.

Remagpie:
I think our usage of serde's `#[dervie(Deserialize)]` on the type `ChainType` was not valid.
The current implementation shows the following error when I pass a file path as a configuration.
```
Error: "Error while parsing TOML: Error { inner: ErrorInner { kind: Custom, line: None, col: 0, message: \"unknown variant `Custom(./akita.json)`, expected one of `mainnet`, `solo`, `simple_poa`, `tendermint`, `cuckoo`, `blake_pow`, `husky`, `saluki`, `corgi`, `beagle`, `custom`\", key: [\"codechain\", \"chain\"] } }"
```
And if I pass `"custom"` as a configuration, it shows the following error.
```
Error: "Error while parsing TOML: Error { inner: ErrorInner { kind: Custom, line: None, col: 0, message: \"invalid type: unit variant, expected newtype variant\", key: [\"codechain\", \"chain\"] } }
```

I'm not sure about the exact format the serde wants, but it seems that it recognized the `Custom(string)` enum in the `ChainType` as a format.

The cli options are fetched directly as a string with the `clap` library, so it doesn't suffer from the problem above.

sgkim126:
I think it's the difference between `toml` and `serde_json`. I'll try to figure out.

sgkim126:
I confused. CLI doesn't use serde_json. It parses the chain type manually with `FromStr`.
It seems that `sede_derive` doesn't support a fallback field with data for now.

Remagpie:
@sgkim126 Thanks for the review.
BTW, why did you think that we're using `serde_json` for cli option parsing?
We're not using json as a cli argument.

PR #1908, opened 2019-12-03
Extract aggregate_work_info function

Remagpie:
Depends: https://github.com/CodeChain-io/codechain/pull/1916

### DIFF
--- a/core/src/consensus/tendermint/engine.rs
+++ b/core/src/consensus/tendermint/engine.rs
@@ -47,6 +47,13 @@ use crate::views::HeaderView;
 use crate::BlockId;
 use rlp::Encodable;
 
+#[derive(Default)]
+struct WorkInfo {
+    proposed: usize,
+    missed: usize,
+    signed: u64,
+}
+
 impl ConsensusEngine for Tendermint {
     fn name(&self) -> &str {
         "Tendermint"
@@ -193,14 +200,6 @@ impl ConsensusEngine for Tendermint {
                 let start_of_the_current_term = metadata.last_term_finished_block_num() + 1;
 
                 if term > 1 {
-                    let start_of_the_previous_term = {
-                        let end_of_the_two_level_previous_term = client
-                            .last_term_finished_block_num((metadata.last_term_finished_block_num() - 1).into())
-                            .unwrap();
-
-                        end_of_the_two_level_previous_term + 1
-                    };
-
                     let banned = stake::Banned::load_from_state(block.state())?;
                     let start_of_the_current_term_header = if block_number == start_of_the_current_term {
                         encoded::Header::new(block.header().clone().rlp_bytes().to_vec())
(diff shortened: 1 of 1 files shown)


core/src/consensus/tendermint/engine.rs
```diff
@@ -425,64 +411,78 @@ fn inactive_validators(
...
-    let mut pending_rewards = HashMap::<Address, u64>::with_capacity(MAX_NUM_OF_VALIDATORS);
-    let mut missed_signatures = HashMap::<Address, (usize, usize)>::with_capacity(MAX_NUM_OF_VALIDATORS);
-    let mut signed_blocks = HashMap::<Address, usize>::with_capacity(MAX_NUM_OF_VALIDATORS);
+    start_of_the_next_term_header: encoded::Header,
+) -> Result<HashMap<Address, WorkInfo>, Error> {
+    let mut work_info = HashMap::<Address, WorkInfo>::new();
+
+    let start_of_the_current_term = {
+        let end_of_the_last_term =
+            chain.last_term_finished_block_num((start_of_the_next_term_header.number() - 2).into()).unwrap();
```
sgkim126: Is `- 2` right?

core/src/consensus/tendermint/engine.rs
```diff
@@ -515,22 +515,22 @@ fn final_rewards(intermediate_reward: u64, number_of_signatures: u64, number_of_
         );
         (0, 0)
     };
-    let numerator = i128::from(intermediate_reward)
-        * (a * i128::from(number_of_signatures) + b * i128::from(number_of_blocks_in_term));
+    let numerator =
+        i128::from(intermediate_reward) * (a * (number_of_signatures as i128) + b * (number_of_blocks_in_term as i128));
```
sgkim126: You'd better prefer to use `from` instead of `as`. Using `as` coerces into type conversion even it's not safe.

core/src/consensus/tendermint/engine.rs
Remagpie: Yes, it's right.
End of the current term = block@(Start of the next term - 1)
End of the last term = block@((End of the current term - 1).metadata.lastTermFinished)

It becomes a bit different if the current term has only one block, but surprisingly, the final formula remains the same.

core/src/consensus/tendermint/engine.rs
Remagpie: `From<usize>` is not implemented for `i128`. Do you have any idea for solving this?

sgkim126: You can use `TryFrom` with `unwarp`. `TryFrom<usize> for i128` is safe to unwrap because it always returns `Ok`

Remagpie: Hmm. Let me try it.

Remagpie:
I rebased the PR to @foriequal0 's refactoring branch.
Only the last commit is the new commit.

sgkim126:
@foriequal0 It looks good to me, but would you double-check?

Issue #1744, opened 2019-08-19
Account subcommand doesn't recognize chain option at the front


Remagpie:
```
# This works
codechain account -c corgi import 'keystore-file.json'
# This doesn't work
codechain -c corgi account import 'keystore-file.json'
```

I felt this behavior confusing. We have to recognize the chain option wherever it is.

PR #1909, opened 2019-12-03
Make --chain argument global

Remagpie:
Fix https://github.com/CodeChain-io/codechain/issues/1744

Now the `-c`(or `--chain`) argument is a global command-line argument.
Every subcommand accepts it as an option, and the position of the argument doesn't matter anymore.

### DIFF
--- a/codechain/codechain.yml
+++ b/codechain/codechain.yml
@@ -3,6 +3,12 @@ version_short: "v"
 author: CodeChain Team <hi@codechain.io>
 about: CodeChian client
 args:
+    - chain:
+        short: c
+        long: chain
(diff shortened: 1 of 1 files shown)


codechain/codechain.yml
```diff
@@ -3,6 +3,12 @@ version_short: "v"
 author: CodeChain Team <hi@codechain.io>
 about: CodeChian client
 args:
+    - chain:
+        short: c
+        long: chain
+        help: Set the blockchain type out of solo, simple_poa, tendermint, cuckoo, blake_pow, corgi, mainnet or a path to chain scheme file.
+        takes_value: true
+        global: true
```
majecty: Wow there is a "global" option.

PR #1919, opened 2019-12-07
Remove Trie::get_with

sgkim126:
We prefer using an interface instead of a concrete object to make code open and generic.
But the current code reveals TrieDB that is an implementation of Trie because the Trait which has a method having generic parameters cannot be an object.
This code removes the get_with method and replaces TrieDB with Trie.

I think we have the other solution to this problem.
The other solution is removing Trie and TrieMut traits and reveals TrieDB and TrieDBMut because they are the only implementation of each one.


sgkim126:
@majecty I resolved conflicts.

PR #1920, opened 2019-12-08
Cleanup db code

sgkim126:


util/journaldb/src/archivedb.rs
```diff
@@ -72,14 +72,9 @@ impl HashDB for ArchiveDB {
...
-                Entry::Occupied(mut entry) => {
-                    *entry.get_mut() += refs;
-                }
-                Entry::Vacant(entry) => {
-                    entry.insert(refs);
-                }
-            }
+            let rc = ret.entry(key).or_default();
+            *rc += refs;
+            assert!(*rc >= -1, "rc should be larger than -1, but {}", rc);
```
joojis: `equal to or greater than`?

sgkim126: fixed

sgkim126:
@joojis The memory db is used as an overlay of the ArchiveDB. So the meaning of the reference count that is less than 0 means it will remove the value in the backend db, and the ArchiveDB has an assertion that the count should not be less than -1.
It makes sense because the archive db should keep the value in the backend when we removing a field twice and inserting the field after that.
However, if the reference count doesn't saturate to -1, the archive db will remove the value because the count is -1.

PR #1922, opened 2019-12-09
Save current block's validator in the state

Remagpie:
This PR introduces the `CurrentValidator` state value, which represents the set of validators for this block.
It allows the tendermint consensus module to verify blocks without reading their parents.
`CurrentValidator` is updated at `on_open_block` for every block that is in the 1st era.

### DIFF
--- a/core/src/consensus/tendermint/worker.rs
+++ b/core/src/consensus/tendermint/worker.rs
@@ -35,7 +35,7 @@ use super::backup::{backup, restore, BackupView};
 use super::message::*;
 use super::network;
 use super::params::TimeGapParams;
-use super::stake::CUSTOM_ACTION_HANDLER_ID;
+use super::stake::{CurrentValidators, CUSTOM_ACTION_HANDLER_ID};
 use super::types::{Height, Proposal, Step, TendermintSealView, TendermintState, TwoThirdsMajority, View};
 use super::vote_collector::{DoubleVote, VoteCollector};
 use super::vote_regression_checker::VoteRegressionChecker;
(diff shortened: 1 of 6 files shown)


core/src/consensus/tendermint/worker.rs
```diff
@@ -1244,13 +1244,18 @@ impl Worker {
...
-            .parent_hash();
+        let parent = self.client().block_header(&(*header.parent_hash()).into()).expect("The parent block must exist");
+        let grand_parent_hash = parent.parent_hash();
         for (bitset_index, signature) in seal_view.signatures()? {
-            let public = self.validators.get(&grand_parent_hash, bitset_index);
+            let public = {
+                let state = self.client().state_at(parent.hash().into()).expect("The parent state must exist");
+                let validators = CurrentValidators::load_from_state(&state)?;
+                if validators.is_empty() {
+                    self.validators.get(&grand_parent_hash, bitset_index)
```
foriequal0: Can you add a comment that this is a fallback for era 0?

Remagpie: Added comment for that. Please check the change.

core/src/consensus/validator_set/dynamic_validator.rs
```diff
@@ -77,6 +77,58 @@ impl DynamicValidator {
...
+
+    pub fn check_enough_votes_with_header(&self, header: &Header, votes: &BitSet) -> Result<(), EngineError> {
+        let client: Arc<dyn ConsensusClient> =
+            self.client.read().as_ref().and_then(Weak::upgrade).expect("Client is not initialized");
+
+        let validators = {
+            match client.state_at(header.hash().into()).map(|s| CurrentValidators::load_from_state(&s)) {
+                Some(Ok(current_validators)) if !current_validators.is_empty() => {
+                    let mut result = (*current_validators).clone();
+                    result.reverse();
```
foriequal0: Reversing the vector outside of the `self.validators` seems an abstraction leakage.

PR #1925, opened 2019-12-12
Remove an unguaranteed transaction order

HoOngEe:
To guarantee that the delegate transaction is executed after the self-nomination transaction, the transactions should be sent to the same nodes.


majecty:
Could you add the description you wrote in this PR to the commit message?

HoOngEe:
> Could you add the description you wrote in this PR to the commit message?

I added it to the commit message.

PR #1926, opened 2019-12-12
Add a restriction comment to the verify_block_family function

HoOngEe:
I misused this function based on the wrong assumption: This function will be invoked after the two corresponding bodies imported


HoOngEe:
I changed "have imported" to "have been imported"

PR #1928, opened 2019-12-12
Do not export TrieDB and TrieDBMut

sgkim126:
This patch makes codechain use Trie and TrieMut which are abstracted
interfaces instead of concrete implementations.

### DIFF
--- a/util/merkle/benches/triedb.rs
+++ b/util/merkle/benches/triedb.rs
@@ -29,7 +29,7 @@ extern crate test;
 use std::path::Path;
 use std::sync::Arc;
 
-use cmerkle::{Trie, TrieDB, TrieDBMut, TrieMut};
+use cmerkle::{Trie, TrieFactory, TrieMut};
 use kvdb::DBTransaction;
 use primitives::H256;
 use rand::random;
@@ -61,7 +61,7 @@ impl TestDB {
         let mut root = H256::new();
         {
             let hashdb = journal.as_hashdb_mut();
-            let mut trie = TrieDBMut::new(hashdb, &mut root);
+            let mut trie = TrieFactory::create(hashdb, &mut root);
             for i in 0..size {
                 trie.insert(&i.to_be_bytes(), &i.to_be_bytes()).unwrap();
             }
@@ -92,14 +92,14 @@ impl TestDB {
         }
     }
 
-    fn trie(&self) -> TrieDB {
+    fn trie<'db>(&'db self) -> impl Trie + 'db {
         let hashdb = self.journal.as_hashdb();
-        TrieDB::try_new(hashdb, &self.root).unwrap()
+        TrieFactory::readonly(hashdb, &self.root).unwrap()
     }
 
-    fn trie_mut(&mut self) -> TrieDBMut {
+    fn trie_mut<'db>(&'db mut self) -> impl TrieMut + 'db {
         let hashdb = self.journal.as_hashdb_mut();
-        TrieDBMut::new(hashdb, &mut self.root)
+        TrieFactory::create(hashdb, &mut self.root)
     }
 
     fn flush(&mut self) {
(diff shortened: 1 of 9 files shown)


Remagpie:
Nice patch! I totally forgot that we had `TrieFactory`.

util/merkle/benches/triedb.rs
```diff
@@ -139,9 +139,11 @@ fn bench_read_multiple(b: &mut Bencher) {
 fn bench_write_single(b: &mut Bencher) {
     let mut db = TestDB::new(DB_SIZE);
     b.iter(|| {
-        let mut trie = db.trie_mut();
-        let item = random::<usize>() % DB_SIZE + DB_SIZE;
-        let _ = trie.insert(&item.to_be_bytes(), &item.to_be_bytes()).unwrap();
+        {
```
Remagpie: Why did you add a new block?

util/merkle/src/triedb.rs
```diff
@@ -40,12 +40,12 @@ use crate::{Trie, TrieError};
...
 /// let mut root = H256::new();
-/// TrieDBMut::new(&mut memdb, &mut root).insert(b"foo", b"bar").unwrap();
-/// let t = TrieDB::try_new(&memdb, &root).unwrap();
+/// TrieFactory::create(&mut memdb, &mut root).insert(b"foo", b"bar").unwrap();
+/// let t = TrieFactory::readonly(&memdb, &root).unwrap();
 /// assert!(t.contains(b"foo").unwrap());
 /// assert_eq!(t.get(b"foo").unwrap().unwrap(), b"bar".to_vec());
 /// ```
-pub struct TrieDB<'db> {
+pub(crate) struct TrieDB<'db> {
```
Remagpie: Is this required?

util/merkle/benches/triedb.rs
```diff
@@ -92,14 +92,14 @@ impl TestDB {
         }
     }
 
-    fn trie(&self) -> TrieDB {
+    fn trie<'db>(&'db self) -> Box<dyn Trie + 'db> {
```
Remagpie: The return type became a lot verbose than before, and it seems it's coming from the implementation of `TrieFactory`'s methods.
I suggest simplifying this by refactoring `TrieFactory` and remove those `Box` and lifetimes if possible.

sgkim126: You cannot remove Box because the size of the return type should be known at the compile time.

Remagpie: Oh, I see.

sgkim126: Ah, I think `impl Trait` can work. I'll try it.

util/merkle/benches/triedb.rs
sgkim126: To drop trie before flushing db.

sgkim126: Both `trie_mut` and `flush` mutably borrow `db`.

Remagpie: It seems it's not related to the commit title "Do not export TrieDB and TrieDBMut".
Could you extract it as a new commit?

util/merkle/src/triedb.rs
sgkim126: Yes. The TrieDB should not be exported out of the crate, but it's used inside the crate.

util/merkle/benches/triedb.rs
sgkim126: It's related. The return type of `trie_mut` was not bounded to the lifetime of `db`, but this patch makes it bounded.

util/merkle/src/triedb.rs
Remagpie: Isn't it exported only when we `pub use` in the `mod.rs`?
I think removing `pub` in `mod.rs` is enough.

util/merkle/benches/triedb.rs
Remagpie: Oh, I thought that there would be no compile error even if we flatten the block.
It seems we need it.

util/merkle/src/triedb.rs
sgkim126: I want to guarantee that this doesn't go out of the crate.

util/merkle/benches/triedb.rs
sgkim126: `impl Trait` worked.

util/merkle/src/triedb.rs
Remagpie: Hmm. I think that we didn't "have to" introduce this change.
You can leave it as is if you want, but I'd like to suggest minimizing the change if possible.

PR #1929, opened 2019-12-13
Import shard level states in the snapshot synchronization

Remagpie:


sync/src/block/extension.rs
```diff
@@ -1044,10 +1116,15 @@ impl Extension {
 
     fn on_chunk_response(&mut self, from: &NodeId, roots: &[H256], chunks: &[Vec<u8>]) {
         let (block, restore) = match self.state {
-            State::SnapshotChunk {
+            State::SnapshotTopChunk {
                 block,
```
sgkim126: It's not related to this commit, but the current code copies a block hash unnecessary. You can fix it with `ref block`.

sync/src/snapshot/mod.rs
```diff
@@ -94,12 +95,26 @@ impl Service {
         }
     }
 }
+fn snapshot(db: &StateDB, block_hash: BlockHash, root: H256, dir: &str) -> Result<(), SnapshotError> {
+    snapshot_trie(db.as_hashdb(), block_hash, root, dir)?;
+
+    let shard_roots: Vec<_> = {
```
sgkim126:
```rust
    let top_state = TopLevelState::from_existing(db.clone(&root), root)?;
    let shard_roots = {
        let metadata = top_state.metadata()?.expect("Metadata must exist for snapshot block");
        let shard_num = *metadata.number_of_shards();
        (0..shard_num).map(|n| top_state.shard_root(n))
    };
```
You don't need to collect the shard roots.

sync/src/block/extension.rs
sgkim126: Ah, this PR contains this change too. I saw https://github.com/CodeChain-io/codechain/pull/1929/commits/5a6853c59cb86060d41cd92b6649e0bf6ad0eb9b first, so I thought it was not part of this PR.

sgkim126: Ignore these comments, I found that `send_chunk_request` mutably borrows `self`.

sync/src/block/extension.rs
```diff
@@ -999,59 +999,60 @@ impl Extension {
...
-                if c.is_empty() {
-                    cdebug!(SYNC, "Peer {} sent empty response for chunk request {}", from, r);
+        let (block, restore) = match self.state {
+            State::SnapshotChunk {
+                block,
+                ref mut restore,
+            } => (block, restore),
+            _ => return,
+        };
+        for (r, c) in roots.iter().zip(chunks) {
```
sgkim126: What happens if the size of `roots` and `chunks` are different?

sync/src/block/extension.rs
```diff
@@ -75,6 +75,64 @@ enum State {
...
 }
 
+impl State {
+    fn initial(client: &Client, snapshot_target: Option<(BlockHash, u64)>) -> Self {
+        let (hash, num) = match snapshot_target {
+            Some(target) => target,
+            None => return State::Full,
+        };
+        let header = match client.block_header(&num.into()) {
+            Some(ref h) if h.hash() == hash => h.clone(),
```
sgkim126: Cloning `header` only needed when the block body doesn't exist to return `State::SnapshotBody`, and it's an exceptional case.
I think it's better to remove unnecessary cloning because cloning header is an expensive job.

sgkim126:
```rust
        match client.block_header(&num.into()) {
            Some(ref header) if header.hash() == hash => {
                if client.block_body(&hash.into()).is_none() {
                    let parent_hash = header.parent_hash();
                    let parent =
                        client.block_header(&parent_hash.into()).expect("Parent header of the snapshot header must exist");
                    return State::SnapshotBody {
                        header: header.clone(),
                        prev_root: parent.transactions_root(),
                    }
                }

                let state_db = client.state_db().read();
                let state_root = header.state_root();
                match TrieFactory::readonly(state_db.as_hashdb(), &state_root) {
                    Ok(ref trie) if trie.is_complete() => State::Full,
                    _ => State::SnapshotChunk {
                        block: hash,
                        restore: SnapshotRestore::new(state_root),
                    },
                }
            },
            _ => State::SnapshotHeader(hash, num),
        }
```

sync/src/block/extension.rs
```diff
@@ -138,34 +196,29 @@ impl Extension {
...
-            let parent_hash = header.parent_hash();
-            let parent =
-                client.block_header(&parent_hash.into()).expect("Parent header of the snapshot header must exist");
-            return State::SnapshotBody {
-                header,
-                prev_root: parent.transactions_root(),
+    fn move_state(&mut self) {
+        let next_state = self.state.next(&self.client);
+        cdebug!(SYNC, "Transitioning the state to {:?}", next_state);
+        if discriminant(&next_state) == discriminant(&State::Full) {
```
foriequal0: You might want to
```rust
if let State::Full = next_state { ... }
```

sync/src/block/extension.rs
```diff
@@ -138,34 +196,29 @@ impl Extension {
...
+                State::SnapshotHeader(hash, _) => *hash,
+                State::SnapshotBody {
+                    header,
+                    ..
+                } => header.hash(),
+                State::SnapshotChunk {
+                    block,
+                    ..
+                } => *block,
+                State::Full => unreachable!("Trying to transition state from State::Full"),
```
foriequal0: `unreachable` seems to be used as a debug designator in a pretty local scope. 
https://doc.rust-lang.org/std/macro.unreachable.html
`panic` would be more precise if you meant to throw an error on the violation of interface assumption.

sync/src/block/extension.rs
Remagpie: It won't happen since it is already checked in `is_valid_response`.

sync/src/block/extension.rs
Remagpie: I think the performance is not an important thing to consider since this function is executed only once at the initialization.
I'd like to emphasize the readability of this part, and I tried to make the function check-and-return for each enum items.
What do you think in this perspective?

sync/src/block/extension.rs
Remagpie: I copied this part from the existing code, and I think we should change that part too if we're fixing it.
Do you think your suggestion is better?

sync/src/block/extension.rs

sync/src/block/extension.rs
foriequal0: I thought I've used it for matching complex enum. Not using `discriminant` is better for the short enum.

Remagpie: It's used the same way as I used. ref: https://github.com/CodeChain-io/codechain/blob/snapshot/sync/src/block/extension.rs#L178
I'll change them altogether in a new commit then.

sync/src/block/extension.rs
sgkim126: If then, please add an assertion that the size of them is equal.

sync/src/block/extension.rs
sgkim126: I think the performance can be a matter even it's executed only at the initialization.
The slow startup sometimes makes a problem.
Moreover, I don't think the previous code is more readable.

sgkim126: It suddenly come to mind that the if guard allows move since rust 1.39.0.
You can change it to
```rust
        let header = match client.block_header(&num.into()) {
            Some(h) if h.hash() == hash => h,
```

Remagpie: Seems nice. I'll change it like that.

sync/src/block/extension.rs
Remagpie: Added `assert_eq` here.

Remagpie:
@foriequal0 ping

PR #1932, opened 2019-12-20
Changing mempool_getPendingTransactions to include future transactions.

MSNTCS:
The new change is made. The change is supposed to include the future queue in pending transactions as well. Previously, we only had the current queue. Now, `mempool_getPendingTransactions` can return either the `current queue` or both the `current` and `future queues`. In order to have `future queue`, the boolean `future_included` parameter is considered.


test/src/helper/spawn.ts
```diff
@@ -859,7 +859,7 @@ export default class CodeChain {
     public async waitForTermChange(target: number, timeout?: number) {
         const start = Date.now();
         while (true) {
-            const termMetadata = await stake.getTermMetadata(this.sdk);
+            const termMetadata = await stake.getTermMetadata(this.sdk as any);
```
joojis: It seems to be unrelated. Please remove this change from the PR.

test/src/e2e/mempool.test.ts
```diff
@@ -46,26 +47,54 @@ describe("Sealing test", function() {
...
-        expect(await node.sdk.rpc.chain.getSeq(faucetAddress)).to.equal(seq);
-        await node.sendPayTx({ seq: seq + 2 });
-        expect(await node.sdk.rpc.chain.getSeq(faucetAddress)).to.equal(seq);
-        await node.sendPayTx({ seq: seq + 1 });
-        expect(await node.sdk.rpc.chain.getSeq(faucetAddress)).to.equal(seq);
-        await node.sendPayTx({ seq: seq });
-        expect(await node.sdk.rpc.chain.getSeq(faucetAddress)).to.equal(
-            seq + 4
-        );
+    it.only("all transaction in both queues should be included", async function() {
```
joojis: You need to remove `only` here.

joojis: Plus, in this case you need to add a new test,  not changing the existing one.

test/src/helper/spawn.ts
MSNTCS: Done ✅

test/src/e2e/mempool.test.ts
MSNTCS: Done ✅

core/src/transaction.rs
```diff
@@ -240,6 +240,7 @@ pub struct LocalizedTransaction {
 impl LocalizedTransaction {
     /// Returns transaction signer.
     /// Panics if `LocalizedTransaction` is constructed using invalid `UnverifiedTransaction`.
+
```
joojis: Could you remove this change?

This is the only change in the file but unnecessary.

core/src/client/test_client.rs
```diff
@@ -542,10 +542,17 @@ impl BlockChainClient for TestBlockChainClient {
         self.miner.ready_transactions(range)
     }
 
+    fn future_ready_transactions(&self, range: Range<u64>) -> PendingSignedTransactions {
+        self.miner.future_ready_transactions(range)
```
joojis: I'm curious if there is a better name.

I think, in this context, "ready" means the "current". For me, it's hard to make guess the behavior from this name.

core/src/miner/mem_pool.rs
```diff
@@ -934,7 +934,105 @@ impl MemPool {
...
+                current_size < size_limit
+            })
+            .collect();
+
+
+        let mut current_signed_tx: Vec<SignedTransaction> = pending_items.iter().map(|t| t.tx.clone()).collect();
+        let mut future_signed_tx: Vec<SignedTransaction> = future_pending_items.iter().map(|t| t.tx.clone()).collect();
+        for tx in &mut future_signed_tx {
+            current_signed_tx.push(tx.clone());
+        }
```
joojis: I think there is a better way to concatenating 2 vectors. but, I'm not sure at this moment.

You can look at https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.chain Or ask anybody else.
These days I don't write Rust code much. 😓

test/src/e2e.long/mempool.test.ts
```diff
@@ -38,7 +38,10 @@ describe("Memory pool size test", function() {
             sending.push(nodeA.sendPayTx({ seq }));
         }
         await Promise.all(sending);
-        const pendingTransactions = await nodeA.sdk.rpc.chain.getPendingTransactions();
+        const pendingTransactions = await nodeA.sdk.rpc.sendRpcRequest(
+            "mempool_getPendingTransactions",
+            [null, null, true]
```
joojis: Before this patch, the test would send `[null, null, false]`. I guess this was the cause of test failure.

core/src/miner/mem_pool.rs
```diff
@@ -934,7 +934,106 @@ impl MemPool {
...
+            })
+            .collect();
+
+
+        let mut current_signed_tx: Vec<SignedTransaction> = pending_items.iter().map(|t| t.tx.clone()).collect();
+        let mut future_signed_tx: Vec<SignedTransaction> = future_pending_items.iter().map(|t| t.tx.clone()).collect();
+        current_signed_tx.append(&mut future_signed_tx);
+        //        for tx in &mut future_signed_tx {
+        //            current_signed_tx.push(tx.clone());
+        //        }
```
joojis: Would you remove the comments?

core/src/miner/mem_pool.rs
```diff
@@ -934,7 +934,106 @@ impl MemPool {
...
+        let transactions: Vec<SignedTransaction> = current_signed_tx;
+        let last_timestamp = future_pending_items.into_iter().map(|t| t.inserted_timestamp).max();
```
joojis: I think this would not work since you changed it to use `append`.

`future pending items` is an empty list at this point.

PR #1939, opened 2019-12-27
Convert constants to discriminated enum

### LINKED ISSUE #1143: Use the discriminated enum for the constants
We internally manage constants like this:
```rust
const ERROR_ID_AAA: u8 = 1;
const ERROR_ID_BBB: u8 = 2;
```
You may not notice duplicate numbers, or you may not non-exhausitive pattern.
Use a discriminated enum.
```rust
#[repr(u8)]
enum ErrorId {
    Aaa = 1,
    Bbb = 2,
}
```
It can detect duplicated numbers, or non-exhausitive pattern, etc.
You can automate conversion between number and enum using `num` package (https://github.com/rust-num/num)

somniumism:
I converted constant type to discriminated enum type.

Using enum type provides the safety for type error rather than constant type.

Fixed #1143

### DIFF
--- a/core/src/consensus/stake/actions.rs
+++ b/core/src/consensus/stake/actions.rs
@@ -24,13 +24,39 @@ use primitives::{Bytes, H256};
 use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
 use std::sync::Arc;
 
-const ACTION_TAG_TRANSFER_CCS: u8 = 1;
-const ACTION_TAG_DELEGATE_CCS: u8 = 2;
-const ACTION_TAG_REVOKE: u8 = 3;
-const ACTION_TAG_SELF_NOMINATE: u8 = 4;
-const ACTION_TAG_REPORT_DOUBLE_VOTE: u8 = 5;
-const ACTION_TAG_REDELEGATE: u8 = 6;
-const ACTION_TAG_CHANGE_PARAMS: u8 = 0xFF;
+#[derive(Clone, Copy)]
+#[repr(u8)]
+enum ActionTag {
+    TransferCCS = 1,
+    DelegateCCS = 2,
+    Revoke = 3,
+    SelfNominate = 4,
+    ReportDoubleVote = 5,
+    Redelegate = 6,
+    ChangeParams = 0xFF,
+}
+
+impl Encodable for ActionTag {
+    fn rlp_append(&self, s: &mut RlpStream) {
+        s.append_single_value(&(*self as u8));
+    }
+}
+
+impl Decodable for ActionTag {
+    fn decode(rlp: &Rlp) -> Result<Self, DecoderError> {
+        let tag = rlp.as_val()?;
+        match tag {
+            1u8 => Ok(ActionTag::TransferCCS),
+            2 => Ok(ActionTag::DelegateCCS),
+            3 => Ok(ActionTag::Revoke),
+            4 => Ok(ActionTag::SelfNominate),
+            5 => Ok(ActionTag::ReportDoubleVote),
+            6 => Ok(ActionTag::Redelegate),
+            0xFF => Ok(ActionTag::ChangeParams),
+            _ => Err(DecoderError::Custom("Unexpected ActionTag Value")),
+        }
+    }
+}
 
 #[derive(Debug, PartialEq)]
 pub enum Action {
@@ -181,19 +207,19 @@ impl Encodable for Action {
                 address,
                 quantity,
             } => {
-                s.begin_list(3).append(&ACTION_TAG_TRANSFER_CCS).append(address).append(quantity);
+                s.begin_list(3).append(&ActionTag::TransferCCS).append(address).append(quantity);
             }
             Action::DelegateCCS {
                 address,
                 quantity,
             } => {
-                s.begin_list(3).append(&ACTION_TAG_DELEGATE_CCS).append(address).append(quantity);
+                s.begin_list(3).append(&ActionTag::DelegateCCS).append(address).append(quantity);
             }
             Action::Revoke {
                 address,
                 quantity,
             } => {
-                s.begin_list(3).append(&ACTION_TAG_REVOKE).append(address).append(quantity);
+                s.begin_list(3).append(&ActionTag::Revoke).append(address).append(quantity);
             }
             Action::Redelegate {
                 prev_delegatee,
@@ -201,7 +227,7 @@ impl Encodable for Action {
                 quantity,
             } => {
                 s.begin_list(4)
-                    .append(&ACTION_TAG_REDELEGATE)
+                    .append(&ActionTag::Redelegate)
                     .append(prev_delegatee)
                     .append(next_delegatee)
                     .append(quantity);
@@ -210,7 +236,7 @@ impl Encodable for Action {
                 deposit,
                 metadata,
             } => {
-                s.begin_list(3).append(&ACTION_TAG_SELF_NOMINATE).append(deposit).append(metadata);
+                s.begin_list(3).append(&ActionTag::SelfNominate).append(deposit).append(metadata);
             }
             Action::ChangeParams {
                 metadata_seq,
@@ -218,7 +244,7 @@ impl Encodable for Action {
                 signatures,
             } => {
                 s.begin_list(3 + signatures.len())
-                    .append(&ACTION_TAG_CHANGE_PARAMS)
+                    .append(&ActionTag::ChangeParams)
                     .append(metadata_seq)
                     .append(&**params);
                 for signature in signatures {
@@ -230,7 +256,7 @@ impl Encodable for Action {
                 message2,
             } => {
                 s.begin_list(3)
-                    .append(&ACTION_TAG_REPORT_DOUBLE_VOTE)
+                    .append(&ActionTag::ReportDoubleVote)
                     .append(message1.as_ref())
                     .append(message2.as_ref());
             }
@@ -242,7 +268,7 @@ impl Decodable for Action {
     fn decode(rlp: &Rlp) -> Result<Self, DecoderError> {
         let tag = rlp.val_at(0)?;
         match tag {
-            ACTION_TAG_TRANSFER_CCS => {
+            ActionTag::TransferCCS => {
                 let item_count = rlp.item_count()?;
                 if item_count != 3 {
                     return Err(DecoderError::RlpInvalidLength {
@@ -255,7 +281,7 @@ impl Decodable for Action {
                     quantity: rlp.val_at(2)?,
                 })
             }
-            ACTION_TAG_DELEGATE_CCS => {
+            ActionTag::DelegateCCS => {
                 let item_count = rlp.item_count()?;
                 if item_count != 3 {
                     return Err(DecoderError::RlpInvalidLength {
@@ -268,7 +294,7 @@ impl Decodable for Action {
                     quantity: rlp.val_at(2)?,
                 })
             }
-            ACTION_TAG_REVOKE => {
+            ActionTag::Revoke => {
                 let item_count = rlp.item_count()?;
                 if item_count != 3 {
                     return Err(DecoderError::RlpInvalidLength {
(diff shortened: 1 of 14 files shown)


kseo:
Use the imperative mood in the subject line

types/src/errors/runtime_error.rs
```diff
@@ -477,7 +477,7 @@ pub enum UnlockFailureReason {
 enum FailureReasonID {
     ScriptShouldBeBurnt = 1u8,
     ScriptShouldNotBeBurnt = 2u8,
-    ScriptError = 3u8,
+    ErrorForScript = 3u8,
```
somniumism: I changed a variant name 'ScriptError' to 'ErrorForScript' in enum. In the enum, all variants have the same prefix: `Script`, so an error occurred. But if subtract the prefix from all names, we cannot identify the role of the variable properly. So, only 'ScriptError' was changed; The prefix error is solved.

core/src/consensus/stake/actions.rs
```diff
@@ -26,13 +29,17 @@ use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
...
-const ACTION_TAG_DELEGATE_CCS: u8 = 2;
-const ACTION_TAG_REVOKE: u8 = 3;
-const ACTION_TAG_SELF_NOMINATE: u8 = 4;
-const ACTION_TAG_REPORT_DOUBLE_VOTE: u8 = 5;
-const ACTION_TAG_REDELEGATE: u8 = 6;
-const ACTION_TAG_CHANGE_PARAMS: u8 = 0xFF;
+#[derive(FromPrimitive)]
+#[repr(u8)]
+enum ActionTag {
+    TRANSFER_CCS = 1,
```
sgkim126: Usually, the Pascal case(camel case with the first capital letter) is used for enum variants.
https://doc.rust-lang.org/1.0.0/style/style/naming/README.html

core/src/consensus/tendermint/message.rs
```diff
@@ -190,8 +197,8 @@ impl Encodable for TendermintMessage {
 impl Decodable for TendermintMessage {
     fn decode(rlp: &Rlp) -> Result<Self, DecoderError> {
         let id = rlp.val_at(0)?;
-        Ok(match id {
-            MESSAGE_ID_CONSENSUS_MESSAGE => {
+        Ok(match FromPrimitive::from_u8(id) {
```
sgkim126:
```rust
Ok(match FromPrimitive::from_u8(id).ok_or_else(|| DecoderError::Custom("Unknown message id detected"))?) {
```

core/src/consensus/stake/actions.rs
```diff
@@ -14,10 +14,9 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::sync::Arc;
-use num_derive::FromPrimitive;    
+use num_derive::FromPrimitive;
```
sgkim126: Do you have a reason for this change?

core/Cargo.toml
```diff
@@ -38,6 +38,8 @@ rlp_derive = { git = "https://github.com/CodeChain-io/rlp.git", version = "0.2"
 snap = "0.2"
 table = { path = "../util/table" }
 util-error = { path = "../util/error" }
+num-traits = "0.2"
```
sgkim126: I couldn't agree more changing constants to enum is good, but I have a question whether using num-traits and num-derive or not.
I think implementing `RlpEncodable` and `RlpDecodable` instead of using them seems good either.
How about you @majecty?

majecty: It seems good to use `RlpEncodable` and `RlpDecodable`. @somniumism Could you implement Rlp encoding/decoding functions for the enum type?

majecty:
For the first commit, could you remove the issue number from the commit message? If you write the issue number in the commit message, GitHub will add your commit link to the issue whenever you fix your commit.

majecty:
Could you write why you created this commit in the commit message? You can write the effect of your commit. For example, "Compiler will warn programmers if they miss some enum values in a match statement."

types/src/errors/runtime_error.rs
majecty: You mean you met the Clippy error ["enum_variant_names"](https://rust-lang.github.io/rust-clippy/master/index.html#enum_variant_names)?

How about changing the name of the enum from `FailureReasonID` to `ScriptFailureReasonID`?

somniumism:
```rust
enum ScriptFailureReasonID {
    ShouldBeBurnt = 1u8,
    ShouldNotBeBurnt = 2u8,
    Error = 3u8,
}
```
Is it right?

majecty: Good!

core/Cargo.toml
majecty: What the `num-traits` is doing is that creating a conversion function from a number type to an enum type. The conversion function is required in the serialize/deserialize process.

If we implement `Encodable` trait and `Decodable` trait to the enum types, we don't need to cast. The Rlp code will receive and return enum type values directly.

somniumism:
@sgkim126 As @majecty said, I implement Encodable trait and Decodable trait to the enum type, but occur stack overflow error in test function `double_vote_verify_desirable_report()`. Before talking about errors, is it right I implemented? Could you give me some advice?

core/src/consensus/stake/actions.rs
```diff
@@ -26,13 +26,39 @@ use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
...
+    Revoke = 3,
+    SelfNominate = 4,
+    ReportDoubleVote = 5,
+    Redelegate = 6,
+    ChangeParams = 0xFF,
+}
+
+impl Encodable for ActionTag {
+    fn rlp_append(&self, s: &mut RlpStream) {
+        s.append(self);
```
HoOngEe: You should convert `enum ActionTag` type to rlp-appendable value like `s.append(&(*self as u8))` with derive `Clone` and `Copy` trait for the `ActionTag`

core/src/consensus/stake/actions.rs
```diff
@@ -26,13 +26,39 @@ use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
...
-const ACTION_TAG_DELEGATE_CCS: u8 = 2;
-const ACTION_TAG_REVOKE: u8 = 3;
-const ACTION_TAG_SELF_NOMINATE: u8 = 4;
-const ACTION_TAG_REPORT_DOUBLE_VOTE: u8 = 5;
-const ACTION_TAG_REDELEGATE: u8 = 6;
-const ACTION_TAG_CHANGE_PARAMS: u8 = 0xFF;
+
+#[repr(u8)]
+enum ActionTag {
+    TrasnferCCS = 1,
```
HoOngEe: Typo: `TrasnferCCS` -> `TransferCCS`

core/src/consensus/stake/actions.rs
sgkim126: As @HoOngEe said, you should call `append` with casted value. This code is the reason for stack overflow since it calls rlp_append infinitely.

sgkim126:
@somniumism as you said, there is no reason to use rlp for it. I think it should be enumeration without primitive conversion and we should modify the decode function at https://github.com/CodeChain-io/codechain/blob/7c9ef51002e2c020f9ab37e61d4964709c57095d/vm/src/decoder.rs#L86

core/src/miner/mod.rs
```diff
@@ -197,3 +198,14 @@ pub enum TransactionImportResult {
 
 #[cfg(all(feature = "nightly", test))]
 mod mem_pool_benches;
+
+fn fetch_account_creator<'c>(client: &'c dyn AccountData) -> impl Fn(&Public) -> AccountDetails + 'c {
```
sgkim126: Why does this commit contain this change?

somniumism: @sgkim126 Sorry for my mistake. I made a mistake because I wasn't used to using git. I solved it with @majecty 's help. Please check and review again.

Cargo.lock
```diff
@@ -1767,9 +1767,12 @@ dependencies = [
 
 [[package]]
 name = "num-traits"
-version = "0.2.5"
+version = "0.2.10"
```
sgkim126: We don't need to update `num-traits` in this commit.

vm/tests/executor.rs
```diff
@@ -655,32 +654,13 @@ fn dummy_input() -> AssetTransferInput {
     }
 }
 
-#[test]
-fn timelock_invalid_type() {
```
sgkim126: Why did you remove this test?

vm/tests/executor.rs
```diff
@@ -815,34 +795,14 @@ fn timelock_block_age_fail() {
     )
 }
 
-#[test]
-fn timelock_block_age_success() {
```
sgkim126: and it?

vm/tests/executor.rs
somniumism: Let's see this: `&[Instruction::Push(0), Instruction::ChkTimelock(5)]`
The ChkTimelock function has a argument of the TimelockType, so constant value(like u8 or int) is invalid.
In addition, TimelockType can only be expressed from 1 to 4(Block, BlockAge, Time, and TimeAge). The test code was removed because other types and values were not accepted as an argument.

somniumism: In byte code, however, the constant can be the argument. This code is to test that case.

vm/src/decoder.rs
```rust
    #[test]
    #[allow(non_snake_case)]
    fn ChkTimelock() {
        assert_eq!(decode(&[opcode::CHKTIMELOCK]), Err(DecoderError::ScriptTooShort));
        assert_eq!(decode(&[opcode::CHKTIMELOCK, 0]), Err(DecoderError::InvalidImmediateValue(0)));
        assert_eq!(decode(&[opcode::CHKTIMELOCK, 5]), Err(DecoderError::InvalidImmediateValue(5)));
        assert_eq!(decode(&[opcode::CHKTIMELOCK, 1]), Ok(vec![Instruction::ChkTimelock(TimelockType::Block)]));
        assert_eq!(decode(&[opcode::CHKTIMELOCK, 2]), Ok(vec![Instruction::ChkTimelock(TimelockType::BlockAge)]));
        assert_eq!(decode(&[opcode::CHKTIMELOCK, 3]), Ok(vec![Instruction::ChkTimelock(TimelockType::Time)]));
        assert_eq!(decode(&[opcode::CHKTIMELOCK, 4]), Ok(vec![Instruction::ChkTimelock(TimelockType::TimeAge)]));

        assert_eq!(decode(&[opcode::CHKTIMELOCK, 1, opcode::CHKTIMELOCK]), Err(DecoderError::ScriptTooShort));
        assert_eq!(
            decode(&[opcode::CHKTIMELOCK, 1, opcode::CHKTIMELOCK, 2]),
            Ok(vec![Instruction::ChkTimelock(TimelockType::Block), Instruction::ChkTimelock(TimelockType::BlockAge)])
        );
    }
```

sgkim126: Okay. I got it.

Cargo.lock
somniumism: @sgkim126 I don't know why `num-traits` was updated. It is probably because of the Cargo version; It's just my guess. But I think it's not a serious problem. Do you think this must be modified?

majecty: A commit should contain changes only needed. If the changes are not relevant to your work, please remove it.

core/src/consensus/tendermint/message.rs
```diff
@@ -13,7 +13,6 @@
 //
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
-
```
majecty: Please re add new line.

types/src/errors/history_error.rs
```diff
@@ -14,7 +14,6 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-
```
majecty: Please add a new line here.

somniumism: fixed! : )

vm/tests/executor.rs
sgkim126: It seems that it's a valid timelock argument.

somniumism: Fixed it! : )

somniumism:
@majecty @sgkim126 I modified some things you said. I'm waiting for the last review.

majecty:
@somniumism Please fix the conflict.

PR #1941, opened 2019-12-30
Add an RPC deleteAllPendingTransactions()

### LINKED ISSUE #1482: Add an RPC to remove pending transactions

junha1:
This new RPC takes no argument and deletes all pending transactions(both current and future), returning nothing.

Fixes #1482

### DIFF
--- a/core/src/miner/miner.rs
+++ b/core/src/miner/miner.rs
@@ -374,6 +374,11 @@ impl Miner {
         results
     }
 
+    pub fn delete_all_pending_transactions(&self) {
+        let mut mem_pool = self.mem_pool.write();
+        mem_pool.remove_all();
+    }
+
     fn calculate_timelock<C: BlockChainTrait>(&self, tx: &SignedTransaction, client: &C) -> Result<TxTimelock, Error> {
         let mut max_block = None;
         let mut max_timestamp = None;
--- a/rpc/src/v1/traits/mempool.rs
+++ b/rpc/src/v1/traits/mempool.rs
@@ -35,6 +35,10 @@ pub trait Mempool {
     #[rpc(name = "mempool_getErrorHint")]
     fn get_error_hint(&self, transaction_hash: TxHash) -> Result<Option<String>>;
 
+    /// Deletes all pending transactions in the mem pool, including future queue.
+    #[rpc(name = "mempool_deleteAllPendingTransactions")]
+    fn delete_all_pending_transactions(&self) -> Result<()>;
+
     /// Gets transactions in the current mem pool. future_included is set to check whether append future queue or not.
     #[rpc(name = "mempool_getPendingTransactions")]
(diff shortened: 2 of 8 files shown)


node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
```
sgkim126: It seems that you ran yarn in the root.
You should not include this file.

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
sgkim126: ditto

core/src/miner/miner.rs
```diff
@@ -374,6 +374,11 @@ impl Miner {
         results
     }
 
+    pub fn delete_all_transactions(&self) {
```
majecty: What do you think about using `delete_all_pending_transactions` as name instead?

rpc/src/v1/traits/mempool.rs
```diff
@@ -35,6 +35,10 @@ pub trait Mempool {
     #[rpc(name = "mempool_getErrorHint")]
     fn get_error_hint(&self, transaction_hash: TxHash) -> Result<Option<String>>;
 
+    /// Deletes all transactions in the current mem pool, including future queue.
```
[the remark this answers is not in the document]
junha1: No problem.

core/src/miner/miner.rs
junha1: Ok

majecty:
Could you change the PR's title as the Commit's title?
When we send a PR that contains only one commit, we set the PR's title as Commit's title.

We have the same policy on the PR's title and description as the commit's. You should start the PR's title with Verb.

core/src/miner/miner.rs
```diff
@@ -374,6 +374,11 @@ impl Miner {
         results
     }
 
+    pub fn delete_all_pending_transactions(&self) {
+        let mem_pool = &mut self.mem_pool.write();
```
```rust
let mut mem_pool = self.mem_pool.write();
```

junha1: The reason my (bad) code worked is that rust prolongs return value's lifetime when its reference is assigned directly with `let`, right? (I'll fix anyway)

sgkim126: Yes, the lifetime of a temporary object assigned to a let is extended to the enclosing block.
You can read https://doc.rust-lang.org/reference/expressions.html#temporary-lifetimes for more details.
And actually, I don't think your code is bad. I believe that the compiler is smart enough to optimize this code. I just want to make the code clear.

majecty:
Please use a full sentence in the PR description and commit message(not title).

PR #1945, opened 2020-01-02
Store the peer list on the DB of a node

### LINKED ISSUE #1569: Save the peer list on the disk
To reduce bootstrap nodes' burden, we need to save the received peer list in the disk.

MSNTCS:
Storing all stable nodes that have been connected to the node in the database. It fixes #1569".

### DIFF
--- a/core/src/db.rs
+++ b/core/src/db.rs
@@ -34,8 +34,10 @@ pub const COL_EXTRA: Option<u32> = Some(3);
 pub const COL_MEMPOOL: Option<u32> = Some(4);
 /// Column for Transaction error hints
 pub const COL_ERROR_HINT: Option<u32> = Some(5);
+/// Column for Storing Peer list
+pub const COL_PEER: Option<u32> = Some(6);
 /// Number of columns in DB
-pub const NUM_COLUMNS: Option<u32> = Some(6);
+pub const NUM_COLUMNS: Option<u32> = Some(7);
 
 /// Modes for updating caches.
 #[derive(Clone, Copy)]
--- a/core/src/db.rs
+++ b/core/src/db.rs
@@ -34,10 +34,8 @@ pub const COL_EXTRA: Option<u32> = Some(3);
 pub const COL_MEMPOOL: Option<u32> = Some(4);
 /// Column for Transaction error hints
 pub const COL_ERROR_HINT: Option<u32> = Some(5);
-/// Column for Storing Peer list
-pub const COL_PEER: Option<u32> = Some(6);
 /// Number of columns in DB
-pub const NUM_COLUMNS: Option<u32> = Some(7);
+pub const NUM_COLUMNS: Option<u32> = Some(6);
 
 /// Modes for updating caches.
 #[derive(Clone, Copy)]
--- a/core/src/peer_db.rs
+++ b/core/src/peer_db.rs
@@ -28,8 +28,8 @@ pub struct PeerDb {
 }
 
 impl PeerDb {
-    pub fn new(database: Arc<dyn KeyValueDB>) -> Arc<Self> {
-        Arc::new(Self {
+    pub fn new(database: Arc<dyn KeyValueDB>) -> Box<Self> {
+        Box::new(Self {
             db: database,
             peers_and_count: Default::default(),
         })
--- a/core/src/peer_db.rs
+++ b/core/src/peer_db.rs
@@ -1,0 +1,99 @@
+// Copyright 2020 Kodebox, Inc.
+// This file is part of CodeChain.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+use crate::db::COL_EXTRA;
+use cnetwork::{ManagingPeerdb, SocketAddr};
+use kvdb::{DBTransaction, KeyValueDB};
+use parking_lot::Mutex;
+use rlp::RlpStream;
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+pub struct PeerDb {
+    db: Arc<dyn KeyValueDB>,
+    peers_and_count: Mutex<(HashMap<SocketAddr, u64>, usize)>,
+}
+
+impl PeerDb {
+    pub fn new(database: Arc<dyn KeyValueDB>) -> Arc<Self> {
+        Arc::new(Self {
+            db: database,
+            peers_and_count: Default::default(),
+        })
+    }
+}
+
+impl ManagingPeerdb for PeerDb {
+    fn insert(&self, key: SocketAddr) {
+        let (peers, count) = &mut *self.peers_and_count.lock();
+        peers.entry(key).or_insert_with(|| {
+            *count += 1;
+            SystemTime::now().duration_since(UNIX_EPOCH).expect("There is no time machine.").as_secs()
+        });
+
+        if let Some(batch) = get_db_transaction_if_enough_hit(peers, count) {
+            self.db.write(batch).expect("The DB must alive");
+        }
+    }
+
+    fn delete(&self, key: &SocketAddr) {
+        let (peers, count) = &mut *self.peers_and_count.lock();
+        if peers.remove(key).is_some() {
+            *count += 1;
+        }
+
+        if let Some(batch) = get_db_transaction_if_enough_hit(peers, count) {
+            self.db.write(batch).expect("The DB must alive");
+        }
+    }
+}
+
+// XXX: It may not be needed. Generally, in the p2p networks, the old node lives longer.
+// Exaggeratedly, the new node does not affect the stability of the network. In other words, it
+// doesn't matter even we don't restore the fresh updated nodes.
+impl Drop for PeerDb {
+    fn drop(&mut self) {
+        let (peers, _) = &*self.peers_and_count.lock();
+        let batch = get_db_transaction(peers);
+        self.db.write(batch).expect("The DB must alive");
+    }
+}
+
+fn get_db_transaction_if_enough_hit(peers: &HashMap<SocketAddr, u64>, count: &mut usize) -> Option<DBTransaction> {
+    const UPDATE_AT: usize = 10;
+    if *count < UPDATE_AT {
+        return None
+    }
+
+    *count = 0;
+
+    Some(get_db_transaction(peers))
+}
+
+fn get_db_transaction(peers: &HashMap<SocketAddr, u64>) -> DBTransaction {
+    let mut s = RlpStream::new_list(peers.len());
+    for (address, time) in peers {
+        s.begin_list(2).append(address).append(time);
+    }
+    let encoded = s.drain();
+
+    let mut batch = DBTransaction::new();
+
+    const COLUMN_TO_WRITE: Option<u32> = COL_EXTRA;
+    const PEER_DB_KEY: &[u8] = b"peer-list";
+    batch.put(COLUMN_TO_WRITE, PEER_DB_KEY, &encoded);
+    batch
+}
--- a/network/src/lib.rs
+++ b/network/src/lib.rs
@@ -36,6 +36,7 @@ extern crate rand;
 extern crate rlp;
 #[macro_use]
 extern crate rlp_derive;
+extern crate kvdb;
 extern crate never_type;
 extern crate table as ctable;
 extern crate time;
@@ -64,6 +65,7 @@ pub use crate::extension::{
 pub use crate::node_id::{IntoSocketAddr, NodeId};
 pub use crate::service::{Error as NetworkServiceError, Service as NetworkService};
 
+pub use self::p2p::{Handler, ManagingPeerdb};
 pub use crate::filters::{FilterEntry, Filters, FiltersControl};
 pub use crate::routing_table::RoutingTable;
 
--- a/network/src/p2p/handler.rs
+++ b/network/src/p2p/handler.rs
@@ -91,7 +91,6 @@ pub struct Handler {
     socket_address: SocketAddr,
     listener: Listener,
 
-    peer_db: Arc<dyn (ManagingPeerdb)>,
     inbound_connections: RwLock<HashMap<StreamToken, EstablishedConnection>>,
     outbound_connections: RwLock<HashMap<StreamToken, EstablishedConnection>>,
     incoming_connections: RwLock<HashMap<StreamToken, IncomingConnection>>,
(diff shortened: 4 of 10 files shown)


rpc/src/v1/impls/mempool.rs
```diff
@@ -97,7 +97,6 @@ where
         }
     }
 
-
```
sgkim126: Please remove these kindes of changes not related to this commit.

network/src/p2p/mod.rs
```diff
@@ -15,7 +15,7 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 mod connection;
-mod handler;
+pub mod handler;
```
sgkim126: IMO, Reexporting `ManagingPeerdb` would be better than making the `handler` module public.

core/src/consensus/tendermint/engine.rs
```diff
@@ -314,11 +313,13 @@ impl ConsensusEngine for Tendermint {
         let client = self.client().ok_or(EngineError::CannotOpenBlock)?;
         let block_hash = match block_number {
             None => {
-                client.block_header(&BlockId::Latest).expect("latest block must exist").hash() // the latest block
+                client.block_header(&BlockId::Latest).expect("latest block must exist").hash()
+                // the latest block
```
sgkim126: I think this comment is not related to this PR.

core/src/consensus/tendermint/engine.rs
```diff
@@ -314,11 +313,13 @@ impl ConsensusEngine for Tendermint {
...
             }
             Some(block_number) => {
                 assert_ne!(0, block_number);
-                client.block_header(&(block_number - 1).into()).ok_or(EngineError::CannotOpenBlock)?.hash() // the parent of the given block number
+                client.block_header(&(block_number - 1).into()).ok_or(EngineError::CannotOpenBlock)?.hash()
+                // the parent of the given block number
```
sgkim126: ditto.

core/src/db.rs
```diff
@@ -37,6 +37,8 @@ pub const COL_MEMPOOL: Option<u32> = Some(4);
 pub const COL_ERROR_HINT: Option<u32> = Some(5);
 /// Number of columns in DB
 pub const NUM_COLUMNS: Option<u32> = Some(6);
+/// Column for Storing Peer list
+pub const COL_PEER: Option<u32> = Some(7);
```
sgkim126: It should be six and you should increase `NUM_COLUMNS` to seven.

core/src/peer_db.rs
```diff
@@ -0,0 +1,44 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+use crate::db::COL_PEER;
+use cnetwork::ManagingPeerdb;
+use kvdb::{DBTransaction, KeyValueDB};
+use std::sync::Arc;
+
+const COLUMN_TO_WRITE: Option<u32> = COL_PEER;
```
sgkim126: :+1:

core/src/peer_db.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2018-2019 Kodebox, Inc.
```
sgkim126:
```
// Copyright 2020 Kodebox, Inc.
```

network/src/p2p/handler.rs
```diff
@@ -1129,15 +1141,25 @@ impl IoHandler<Message> for Handler {
...
                     }
                 } else {
                     cdebug!(NETWORK, "Invalid outgoing token({}) on deregister", stream);
+                    None
                 }
             }
             _ => unreachable!("Unexpected stream on deregister: {}", stream),
+        };
+
+        if let Some(_address) = peer_socket_adress {
```
sgkim126: I don't think using an Option for this is a good indication of intent.
How about making this code a function instead of using `peer_socket_adress`?

MSNTCS: Can you provide more details? Do you think it is better to put a function in the first and second match?

sgkim126:
```rust
            let batch = DBTransaction::new();
            let remove_target = peer_socket_adress;
            let s = rlp::encode(&remove_target);
            self.peer_db.delete(batch, &s);
```
I meant extracting the code above into a method and calling the method at line 1065 and 1084.

majecty:
Please add an issue number in the description of PR like "Fixes #1569".
Add a description in the commit message and PR description about why we need to save the peer list in the DB and what will be changed after this PR or commit.

Please use a full sentence in the description of PR and commit message.

network/src/p2p/connection/established.rs
```diff
@@ -33,7 +33,7 @@ use crate::SocketAddr;
 
 pub struct EstablishedConnection {
     stream: SignedStream,
-    peer_addr: SocketAddr,
+    pub peer_addr: SocketAddr,
```
sgkim126: Why do you make it public?

network/src/p2p/handler.rs
```diff
@@ -493,6 +491,13 @@ impl IoHandler<Message> for Handler {
                 is_inbound: true,
             } => {
                 let mut inbound_connections = self.inbound_connections.write();
+                let batch = DBTransaction::new();
+                let target = connection.peer_addr();
+                let s = rlp::encode(target);
+                let time: u64 =
+                    SystemTime::now().duration_since(UNIX_EPOCH).expect("There is no time machine.").as_secs();
+                let key = rlp::encode(&time);
+                self.peer_db.insert(batch, &s, &key);
```
sgkim126: It seems that `key` is wrongly named.

network/src/p2p/handler.rs
```diff
@@ -1057,6 +1062,10 @@ impl IoHandler<Message> for Handler {
                     self.routing_table.remove(con.peer_addr());
                     self.inbound_tokens.lock().restore(stream);
                     ctrace!(NETWORK, "Inbound connect({}) removed", stream);
+                    let batch = DBTransaction::new();
+                    let remove_target = con.peer_addr();
+                    let s = rlp::encode(remove_target);
+                    self.peer_db.delete(batch, &s);
```
sgkim126: You missed removing an address of the outbound connection at line 1085.

network/src/lib.rs
```diff
@@ -66,6 +67,8 @@ pub use crate::service::{Error as NetworkServiceError, Service as NetworkService
 
 pub use crate::filters::{FilterEntry, Filters, FiltersControl};
 pub use crate::routing_table::RoutingTable;
+pub use p2p::Handler;
```
sgkim126:
```rust
pub use self::p2p::{Handler, ManagingPeerdb};
```

network/src/p2p/handler.rs
```diff
@@ -16,9 +16,10 @@
 
 use std::collections::{BTreeSet, HashMap, VecDeque};
 use std::iter::FromIterator;
+use std::marker::{Send, Sync};
```
sgkim126: You don't need these imports.
https://doc.rust-lang.org/std/prelude/index.html

network/src/p2p/handler.rs
```diff
@@ -80,14 +82,19 @@ const RETRY_SYNC_MAX: Duration = Duration::from_secs(10); // T1
 const RTT: Duration = Duration::from_secs(10); // T2
 const WAIT_SYNC: Duration = Duration::from_secs(30); // T3 >> T1 + RTT
 
+pub trait ManagingPeerdb: Send + Sync {
+    fn insert(&self, transaction: DBTransaction, key: &[u8], value: &[u8]);
```
sgkim126: Does `ManagingPeerdb` need `DBTransaction`?
The implementation might use a transaction but it seems implementation detail.
I think there is no need to pass the transaction.

sgkim126: The arguments are too generic.
Do you have a plan to insert other types except `SocketAddr`?
If not, I think the key type should be `SocketAddr` or `&SocketAddr`.

sgkim126: And what does the inserting value mean?
IMO, it is an implementation detail not for an interface.

MSNTCS: I cannot understand the second one. Can you explain more? 
I mean this one 
https://github.com/CodeChain-io/codechain/pull/1945/commits/916565582681895600fe754c43a0a078f3011e9c#r363035953

sgkim126: Currently, the `insert` method takes an rlp encoded socket address as `key` and an rlp encoded time as `value`. The key is necessary; the purpose of this trait is storing the address. I think the purpose of the time is to know how the connection is stable, but I wonder whether it's the role of `ManagingPeerdb`. How to store the address and what address to use are implementation details.

network/src/p2p/handler.rs
MSNTCS: Done

network/src/p2p/handler.rs
MSNTCS: Done

network/src/lib.rs
MSNTCS: Done

network/src/p2p/handler.rs
MSNTCS: Done

network/src/p2p/handler.rs
MSNTCS: Hopefully done.

MSNTCS: Got it and apply you desire changes.

MSNTCS:
I tested the unpassed test( `Change common Params that doesn't affect validator set`) on my local machine and it passed the test. My implementation does not have any problem with this test.

core/src/db.rs
```diff
@@ -35,8 +35,11 @@ pub const COL_EXTRA: Option<u32> = Some(3);
 pub const COL_MEMPOOL: Option<u32> = Some(4);
 /// Column for Transaction error hints
 pub const COL_ERROR_HINT: Option<u32> = Some(5);
+/// Column for Storing Peer list
+pub const COL_PEER: Option<u32> = Some(6);
 /// Number of columns in DB
-pub const NUM_COLUMNS: Option<u32> = Some(6);
+pub const NUM_COLUMNS: Option<u32> = Some(7);
```
sgkim126: @majecty I'm not sure whether it's safe to change the number of columns. Would you check whether it's compatible with the current db?

majecty: CodeChain failed to run if the NUM_COLUMNS changed. We might need a migration tool.

Here is the log message:
```
#550732000 2020-01-06 16:35:48.636848000  main WARN kvdb_rocksdb  DB corrupted: Invalid argument: You have to open all column families. Column families not opened: col3, col4, col1, col2, col5, col0, attempting repair
Error: "Low level database error. Some issue with disk?"
```

majecty: @sgkim126 What do you think about adding another issue about the migration? Since to handle it the programmer needs to know some RocksDB knowledge, it is not a good first issue.

sgkim126: Thank you for checking it @majecty.
@MSNTCS Would you upload this PR to https://github.com/CodeChain-io/foundry first?

MSNTCS: @sgkim126 I just did. You can check on the foundry.

sgkim126: Hi @MSNTCS,
I have a gut feeling that we can reuse an [extra column](https://github.com/CodeChain-io/codechain/blob/22fe8037f252b21b2db9e92ce747f1bcc39fba7f/core/src/db.rs#L33) for it because `KeyValueDB` has an [interface to iterate the column that has a specific prefix](https://docs.rs/kvdb/0.1.1/kvdb/trait.KeyValueDB.html#tymethod.iter_from_prefix).
This approach can make it difficult to manage later because it differs from the foundry implementation. But it seems to be easier to maintain than adding migration code.
Would you try it?

MSNTCS: Hello, @sgkim126, 
I will try to implement it, should it be as another commit or do I need to modify the previous one?

sgkim126: I prefer to modify it, but it shouldn't be matter anyway. Please take the way that is comfortable for you.

codechain/auto_self_nominate.rs
```diff
@@ -0,0 +1,156 @@
+// Copyright 2020 Kodebox, Inc.
```
sgkim126: It seems that this commit should not be included here.

sgkim126:
@MSNTCS Would you check the commits included in this PR again?

MSNTCS:
> @MSNTCS Would you check the commits included in this PR again?

Sorry, for the late response. I had changed when I saw your comment. Now is only related commit.

core/src/db.rs
majecty: This PR is still using COL_PEER.
@MSNTCS Did you try the extra column?

sgkim126:
https://github.com/CodeChain-io/foundry/pull/133 is a suggestion to implement PeerDb without implementing the migration code.

MSNTCS:
@majecty Should we update this based on the new implementation? or can we close this?

majecty:
Let's cherry-pick the Seulgi's commits to this PR
@sgkim126 I want to merge this PR with your commits!

MSNTCS:
> Let's cherry-pick the Seulgi's commits to this PR
> @sgkim126 I want to merge this PR with your commits!

@sgkim126 would you mind answering this?

sgkim126:
@MSNTCS I agree on @majecty. It would be better to merge the commits into this PR.

MSNTCS:
> @MSNTCS I agree on @majecty. It would be better to merge the commits into this PR.

Thanks

majecty:
Please run `rustfmt` whenever you create a commit.

majecty:
Your commit messages "Update handler.rs" and "Update service.rs" does not give any meaningful information to a reader.
It seems that the reason why you created the two commits is that some code is different from Foundry's and yours. Then could you give the information in the commit message?

majecty:
@MSNTCS Could you also cherry pick these commits?
IMO, you can squash them in your first commit.
https://github.com/CodeChain-io/foundry/commit/e22ea52a8a2500db5eda8989d64fddad47949359
https://github.com/CodeChain-io/foundry/commit/c6dcf5421d9e6f186a8eb1c3931bb7e54a859321
https://github.com/CodeChain-io/foundry/commit/82716dd2f6e99fecd8e770fadc530e1c1aa83336
https://github.com/CodeChain-io/foundry/commit/7471a905d139bf575c6735eb94c67fd36e1971e5
https://github.com/CodeChain-io/foundry/commit/d5306453aa25e00f37bbdc11544fc0166eaf7eeb
https://github.com/CodeChain-io/foundry/commit/97b8c62cdce33a292c4dc88cb79dc1e3d01f067f

core/src/db.rs
majecty: We will import Seulgi's commits from Foundry. The commits make the code to use COL_EXTRA. Let's finish this thread.

core/src/db.rs
```diff
@@ -35,8 +35,11 @@ pub const COL_EXTRA: Option<u32> = Some(3);
...
+
```
majecty: @MSNTCS Please remove unrelated new line.

network/src/service.rs
```diff
@@ -18,7 +18,8 @@ use crate::client::Client;
 use crate::control::{Control, Error as ControlError};
 use crate::filters::{FilterEntry, FiltersControl};
 use crate::routing_table::RoutingTable;
-use crate::{p2p, Api, NetworkExtension, SocketAddr};
+use crate::{p2p, Api, ManagingPeerdb, NetworkExtension, SocketAddr};
+
```
majecty: Please remove this empty line.

network/src/service.rs
```diff
@@ -28,6 +29,7 @@ use std::collections::HashMap;
 use std::net::IpAddr;
 use std::sync::Arc;
 
+
```
majecty: Please remove this empty line.

PR #1950, opened 2020-01-06
Upgrade fixed-hash crate to 0.2.2 to prevent the latest clippy fail

### LINKED ISSUE #1938: Upgrade fixed-hash
Currently, CodeChain uses fixed-hash@0.2.1 that has a duplicated macro definition.
The duplicated macro makes the latest clippy fail, so we should upgrade it to 0.2.5.
You can read https://doc.rust-lang.org/cargo/commands/cargo-update.html to find how to do it.

somniumism:
Currently, Codechain uses fixed-hash@0.2.1v that has a duplicated macro definition.
The duplicated macro makes the latest clippy(+nightly-2019-12-19) fail. So, we need to upgrade it.

When fixed-hash 0.2.5 was applied, an error(E0599) occurred due to incompatibility for some functions.
This problem can be solved as follows.
1. Upgrade Our rust-codechain-primitives(`primitives`) to use the upgraded fixed-hash.
2. Upgrade libraries using `primitives` used by CodeChain and modules in CodeChain.

However, too many libraries(rlp, rust-codechain-crypto, etc) are using `primitives` that are not updated. Also, this is not a problem that will be solved just by upgrading primitives. We need a large-scale update to use fixed-hash@≥0.2.5 and later versions of some packages.

Fortunately, we checked 0.2.2 had resolved the macro duplication problem. We thought it's best to upgrade to 0.2.2 under the current situation.

Fixed #1938


sgkim126:
Thank you for doing it @somniumism.
Would you apply this change to https://github.com/CodeChain-io/foundry?

PR #1953, opened 2020-01-08
Upgrade rust toolchains

### LINKED ISSUE #1937: Upgrade rust toolchains
Make CodeChain use the latest clippy and rustfmt.
You can reference f9729d5877115f35da9bbce962ed773c617e2b40.

somniumism:
Currently, Codechain uses the clippy(+nightly-2019-11-06) and the rustfmt(+nightly-2019-11-06).
We need to make codechain use the latest toolchains.
I upgraded the clippy to `+nightly-2019-12-19` and the rustfmt to `+nigtly-2019-12-19`.

I fixed clippy error occurred from new clippy version. The warning is this; `clippy waring:using tabs in doc comments is not recommended`

Fix #1937


somniumism:
@sgkim126 If it is fine, I will apply it to foundry too.

util/io/src/lib.rs
```diff
@@ -27,33 +27,33 @@
 //! struct MyHandler;
 //!
 //! struct MyMessage {
-//! 	data: u32
+//!    data: u32
```
majecty: @somniumism Does the rust-fmt change this code?

majecty:
Could you rebase the branch on the master?

util/io/src/lib.rs
somniumism: @majecty I changed this code because of clippy warning: `using tabs in doc comments is not recommended` I'm sorry I couldn't explain it in advance.

majecty: You can split the commit into "Update clippy version" and "Fix clippy error". Since the change is small you can just add a comment on the PR page about why you changed it.

majecty:
@somniumism Could you apply this in Foundry?

PR #1958, opened 2020-01-18
Do not mark the block having invalid transactions root as bad

sgkim126:


sync/src/block/extension.rs
```diff
@@ -717,6 +718,20 @@ impl Extension {
...
+                    .block_header(&(*header.parent_hash()).into())
+                    .expect("The parent header must exist")
+                    .view()
+                    .transactions_root();
+                let calculated_transactions_root =
+                    skewed_merkle_root(parent_transactions_root, transactions.iter().map(Encodable::rlp_bytes));
+                if *header.transactions_root() != calculated_transactions_root {
+                    cwarn!(SYNC, "Received corrupted body for ${}({}", header.number(), hash);
+                    self.body_downloader.remove_target(&[hash]);
+                    continue
```
majecty: I'm not sure whether calling `remove_target` and `continue` is safe or not. I'll investigate it.

majecty: This code makes a problem.

I'll explain how BodyDownloader works first.

If CodeChain receives headers, it adds them in `BodyDownloader::targets` with the same order(Order is urgent here. They should consist a chain)
If CodeChain receives bodies of targets, it marks the blockhashes in the `BodyDownloader::downloaded`.

When we call `BodyDownloader::drain()`, BodyDownloader returns a series of downloaded bodies order by targets. The blockhashes of returned bodies are always a prefix of the targets. If there is a body that is not downloaded yet, the body and following bodies are not returned by the drain function. The drain function also removes the hashes of returned bodies from `BodyDownloader::targets`.

So in this code, if the body was not valid, we should add the failed body's blockhash in the `BodyDownloader::targets` in the correct position.
Also, the bodies following the invalid body should be re-added in the `BodyDownloader::targets` and be marked as downloaded again.

sgkim126: Thank you for your explanation.
I think we should fix https://github.com/CodeChain-io/codechain/blob/d68f1eb50a767c8b2508d4bec9ad42291fcd4f77/sync/src/block/extension.rs#L735 too. I'll try to fix it.

sgkim126: I fixed what you had mentioned.
However, I cannot find how to handle the importing error that I mentioned above.
I'll upload another PR when I fix it.

sync/src/block/downloader/body.rs
```diff
@@ -118,4 +118,26 @@ impl BodyDownloader {
...
+        }];
+        new_targets.extend(remains.into_iter().map(|(hash, transactions)| {
+            let is_empty = transactions.is_empty();
+            self.downloaded.insert(hash, transactions);
+            Target {
+                hash,
+                is_empty,
+            }
+        }));
+        new_targets.append(&mut self.targets);
```
majecty: This is the correct order!

PR #1966, opened 2020-04-23
Remove the comment related to an old policy

HoOngEe:
We do not include failed transactions in blocks. Therefore, fee
imposition and sequence increase do not occur for a failed
transaction.


HoOngEe:
How can I resolve the
`clahub` *Expected - Waiting for status to be reported*

sgkim126:
@HoOngEe I don't know. It sometimes stops and not restored. I'll merge it.

PR #1968, opened 2020-05-21
Fix a deadlock in miner.rs

majecty:
There is a deadlock between immune_users and mem_pool.

A client worker thread locks immune_users first and mem_pool second in
prepare_block function.

An HTTP thread locks mem_pool first in import_own_transactions function
and locks immune_users in add_transactions_to_pool function.

It seems that this deadlock is added when we add immune_users lock. We
need to review the lock of immune_users.

### DIFF
--- a/core/src/miner/miner.rs
+++ b/core/src/miner/miner.rs
@@ -16,7 +16,6 @@
 
 use std::collections::HashSet;
 use std::iter::once;
-use std::iter::FromIterator;
 use std::ops::Range;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


HoOngEe:
How about just moving the line `let immune_users = self.immune_users.read();` to the 565th line ?

majecty:
@HoOngEe @sgkim126 I agree with Seulgi. Locking mem_pool first will be better.

core/src/miner/miner.rs
```diff
@@ -132,6 +132,26 @@ pub struct Miner {
     immune_users: RwLock<HashSet<Address>>,
 }
 
+struct NextAllowedReseal {
+    instant: Mutex<Instant>,
```
sgkim126, line 157: If' it's okay to change Mutex into RwLock, I think we can reuse NextMandatoryReseal for it.

majecty: It seems that we could introduce a generic structure.
However, I don't have any good name for it.

For the NextAllowedReseal and NextMandatoryReseal, I'll make `type NextMandatoryReseal = NextAllowedReseal`.

core/src/miner/miner.rs
```diff
@@ -529,11 +637,10 @@ impl Miner {
         let tx_total = transactions.len();
         let mut invalid_tx_users = HashSet::new();
 
-        let immune_users = self.immune_users.read();
```
majecty, line 532: @sgkim126 
By removing this, the deadlock fixed.

sgkim126: :+1:

PR #1976, opened 2020-06-03
Fetch consistent state in mempool functions

majecty:
We found that mempool mis-calculates sequence in the TPS test. After fixing the mempool to read consistent state, mempool calculates sequence well.

### DIFF
--- a/core/src/miner/mem_pool.rs
+++ b/core/src/miner/mem_pool.rs
@@ -33,7 +33,7 @@ use super::mem_pool_types::{
 use super::TransactionImportResult;
 use crate::client::{AccountData, BlockChainTrait};
 use crate::transaction::{PendingSignedTransactions, SignedTransaction};
-use crate::Error as CoreError;
+use crate::{BlockId, Error as CoreError};
 
 const DEFAULT_POOLING_PERIOD: BlockNumber = 128;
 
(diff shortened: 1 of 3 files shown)


core/src/miner/mem_pool.rs
```diff
@@ -429,12 +429,14 @@ impl MemPool {
 
     // Recover MemPool state from db stored data
     pub fn recover_from_db<C: AccountData + BlockChainTrait>(&mut self, client: &C) {
+        let recover_block_hash = client.chain_info().best_block_hash;
+        let recover_block_id = BlockId::Hash(recover_block_hash);
         let fetch_account = |p: &Public| -> AccountDetails {
             let address = public_to_address(p);
-            let a = client.latest_regular_key_owner(&address).unwrap_or(address);
+            let a = client.regular_key_owner(&address, recover_block_id.into()).unwrap_or(address);
```
sgkim126, line 436: @majecty Would you explain the difference of the latest_regular_key_owner and the regular_key_owner with the best_block_hash?

majecty: @sgkim126 
There is a possibility that fetch_account returns different results in the mempool::add function if the best block is changed between two fetch_acccount calls.

When running the TPS test, mempool::add function is called more than 1000 per second. When running the TPS test, we found that mempool's sequence calculation logic missed sometimes. After changing fetch_account to query state with explicit best block hash, the problem is disappeared.

majecty:
@junha1 I added you as a review since you were curious about the mempool bug.

majecty:
The failed e2e test is "Shutdown test". Since it is not related to this PR, I'll merge this PR. Thanks.

PR #1979, opened 2020-07-30
Move PRs from Foundry

majecty:
@dynaxis @sgkim126 @Byeongjee @junbeomlee @MSNTCS Could you check the commits you made? I moved commits from Foundty to CodeChain. Please tell me if there are omitted commits that should be moved to CodeChain or if I moved commits that should not be moved.

I run tests on my local machine. There are some flaky tests. If I run the tests multiple times, they eventually succeed.

I've moved refactoring and bug fix PRs from Foundry to CodeChain

- Execute self-nomination transactions on a new thread https://github.com/CodeChain-io/foundry/pull/98 @MSNTCS 
- Refactor BodyDownloader https://github.com/CodeChain-io/foundry/pull/116 @sgkim126 
- Add RPCs for checking peer best blocks & target blocks https://github.com/CodeChain-io/foundry/pull/117 @Byeongjee 
- Fix auto-self-nonimante https://github.com/CodeChain-io/foundry/pull/118 @majecty 
- Add an RPC getting minimum fees confiugred by each machine https://github.com/CodeChain-io/foundry/pull/127 @Byeongjee 
- Remove duplicated TermInfo https://github.com/CodeChain-io/foundry/pull/135 @majecty 
- Change constants to enumerations https://github.com/CodeChain-io/foundry/pull/136 @sgkim126 
- Fix typos https://github.com/CodeChain-io/foundry/pull/138 @Byeongjee 
- Upgrade codechain-crypto to 0.2.0 https://github.com/CodeChain-io/foundry/pull/141 @sgkim126 
- Fix typo https://github.com/CodeChain-io/foundry/pull/152 @Byeongjee 
- Report double votes in commit messages. https://github.com/CodeChain-io/foundry/pull/181 @Byeongjee 
- Fix typo in test helper's constants https://github.com/CodeChain-io/foundry/pull/224 Seonpyo
- Assign unbound port numbers in test instances https://github.com/CodeChain-io/foundry/pull/255 @Byeongjee 
- Fix a wrong argument name https://github.com/CodeChain-io/foundry/pull/270 @Byeongjee 
- Remove ChainNotify::transactions_received() https://github.com/CodeChain-io/foundry/pull/282 @sgkim126 
- Fix wrong debug message https://github.com/CodeChain-io/foundry/pull/293 @Byeongjee 
- Change the error message that Foundry prints when account is not unlocked https://github.com/CodeChain-io/foundry/pull/294 @majecty 
[...]

PR #1981, opened 2020-08-06
Make sync tests not to propagate transactions

majecty:
The current CodeChain master does not fail the tests without this commit. 
However, https://github.com/CodeChain-io/codechain/pull/1979 fails the tests quite often without this PR.

These tests use solo consensus. To generate a block with the solo
consensus, there should be a transaction. These tests send a
transaction to a node to make a block. Then the node propagates the
generated block to other nodes. These tests then check the other nodes
whether the received the block or not.

However, if the transaction is propagated before the block, these
tests fail. If the transaction is propagated, other nodes will
generate their own blocks. Then they will reject the block that the
first node generated.

### DIFF
--- a/test/src/e2e.long/sync2.test.ts
+++ b/test/src/e2e.long/sync2.test.ts
@@ -25,8 +25,10 @@ describe("sync 2 nodes", function() {
 
     describe("2 nodes", function() {
         beforeEach(async function() {
-            nodeA = new CodeChain();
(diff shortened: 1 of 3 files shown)


junha1:
So if there is no transaction, `solo` consensus never generate a block?

majecty:
@junha1 Yes it is. As far as I know, solo only generates a block when a transaction is sent.

PR #1994, opened 2020-11-25
Do not send email too frequently

majecty:


util/logger/src/logger.rs
```diff
@@ -110,7 +114,15 @@ impl Log for Logger {
...
             if log_level == Level::Error {
                 if let Some(email_alarm) = &self.email_alarm {
-                    email_alarm.send(&format!("{} {} {}", thread_name, log_target, log_message))
+                    let mut last_email_sent = self.last_email_sent.lock();
+                    let sent_recently = match *last_email_sent {
+                        Some(last_sent) => last_sent.elapsed() < Duration::from_secs(300),
+                        None => false,
+                    };
+                    if !sent_recently {
+                        email_alarm.send(&format!("{} {} {}", thread_name, log_target, log_message));
```
junha1, line 123: Make sure that this doesn't cause another log. It will lead to a deadlock. If not, LGTM.

junha1 (approved):

util/logger/src/logger.rs
majecty: The `send` function does not use log.

majecty:
The test succeeds in my local machine.

